PR #1, opened 2014-11-19
Initialize repo

sbidoul:


pedrobaeza:
Can you please add a README.md with at least flags, a brief description, and OCA disclaimer?

You can take this one as example: https://github.com/OCA/bank-payment/blob/8.0/README.md

sbidoul:
@pedrobaeza done

README.md
```diff
@@ -1,6 +1,17 @@
+[![Build Status](https://travis-ci.org/OCA/bank-statement-import.svg?branch=8.0)](https://travis-ci.org/OCA/bank-statement-import)
+[![Coverage Status](https://coveralls.io/repos/OCA/bank-statement-import/badge.png?branch=8.0)](https://coveralls.io/r/OCA/bank-statement-import?branch=8.0)
+
 Odoo bank importation
```
pedrobaeza: I would put here `OCA modules for bank importation for Odoo` for trying to give brand to OCA.

sbidoul:
@pedrobaeza I improved the README

pedrobaeza:
Thanks for the change!

Another thing: @gurneyalex is reporting in other projects that `UNIT_TEST="1"` is very slow, taking 30 min or more, so we should considered its inclussion.

sbidoul:
@pedrobaeza right. I never really understood this UNIT_TEST thing anyway, and I prefer that we keep modules and tests fully compatible. Should be ok now.

pedrobaeza:
:+1:

pedrobaeza:
The goal of UNIT_TEST is to tests isolated each module.

guewen:
:+1:

PR #2, opened 2014-12-15
8.0 move framework

NL66278:
Move backported import modules from bank-payment repository to here.

Did a simple move, because there is no real history attached to these modules anyway.


pedrobaeza:
One question about the change in the year: Do we have to maintain each year the change on the date to get correct tests?

NL66278:
@pedrobaeza At the moment this is really the case. Unless ofcourse testing on a database where periods for earlier years already exist. Maybe we should add to the documentation for which dates a period should exist. I am open to any better idea.

pedrobaeza:
Can we fix the test to stick to a year (although outdated)?

pedrobaeza:
You should include `#noqa` sentences to avoid flake8 and pylint checks documenting that it's due to odoo upstream.

NL66278:
@pedrobaeza your wish is my command......

NL66278:
The Travis Build fails because the ofx import needs the ofxparse module. Does anyone know wether this can be solved by adding a pip install to the .travis.yml file? Or any other suggestion how to solve this?

sbidoul:
Hi Ronald,

Adding pip install ofxparse before travis_install_nightly in .travis.yml should do it.
Example here: https://github.com/OCA/bank-payment/blob/8.0/.travis.yml
(except the sudo should not be necessary).

sbidoul:
@NL66278 can you add the #noqa in the **init**.py files too so travis becomes green?

NL66278:
It seems pylint is very difficult to satisfy on the imports in **init**.py... How come this is not a problem in other repositories?

yvaucher:
@NL66278 Relative import error means you needs to do this:

Instead of 

```
import local_stuff
```

Do:

```
from . import local_stuff
```

yvaucher:
@NL66278 By the way you might need to reconfigure your local git username and/or email. Commits aren't linked to your account. (make sure you use same email address in your local and on github, you can put multiple email addresses on github)

https://help.github.com/articles/setting-your-email-in-git/

sbidoul:
:+1: Thanks!

NL66278:
@yvaucher Thanks Yannick for the tip on my email address. Changed that now.

pedrobaeza:
:+1: 

Thanks!

PR #4, opened 2014-12-24
8.0 import parsers

NL66278:
- Add framework to allow for easy conversion of 7.0 and previous style bank statement import parsers.
- As an example provide a (fully functional) convertion of the camt parser.

TODO: add more of the 7.0 functionality.

REMARK: settings have been removed here, as all the settings really apply to reconciliation, not to import.


bank_statement_parse/__openerp__.py
```diff
@@ -0,0 +1,71 @@
...
+        'security/ir.model.access.csv',
+        'view/menu.xml',
+    ],
+    'js': [
+    ],
+    'description': '''
+    Base module to write parsers for bank statement import files.
+
+    The module account_bank_statement_import, backported from odoo 9.0, is
+    exentended, so parsers developed might hope to be compatible with future
```
pedrobaeza: s/exentended/extended

bank_statement_parse/__openerp__.py
```diff
@@ -0,0 +1,71 @@
...
+        'account_bank_statement_import',
+        ],
+    'data': [
+        'security/ir.model.access.csv',
+        'view/menu.xml',
+    ],
+    'js': [
+    ],
+    'description': '''
+    Base module to write parsers for bank statement import files.
```
pedrobaeza: Please extract this to README.rst file.

bank_statement_parse/__openerp__.py
```diff
@@ -0,0 +1,71 @@
...
+
+    The module account_bank_statement_import, backported from odoo 9.0, is
+    exentended, so parsers developed might hope to be compatible with future
+    development of Odoo.
+
+# TODO Not al of the functionality described below has been implemented yet!
```
pedrobaeza: s/al/all

bank_statement_parse/model/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import account_bank_statement_import
```
pedrobaeza: from .

bank_statement_parse/parserlib/bank_statement.py
```diff
@@ -0,0 +1,65 @@
...
+#  GNU Affero General Public License for more details.
+#
+#  You should have received a copy of the GNU Affero General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+class BankStatement(object):
+    '''
```
pedrobaeza: Use double quotas for docstrings.

bank_statement_parse/__openerp__.py
pedrobaeza: Now you don't need this if you put README.rst, and thus you don't need to maintain two texts.

NL66278:
Please do not merge for the moment.

Rethinkink my solution to the problem of lots of Undefined Banks in the statement - now solved by supressing the creation - I think the automatic creation should rather be enhanced.

This will make the import improvements be more compatible with the standard Odoo functionality of linking autocreated bank accounts to manually created partners on reconciliation.

NL66278:
Bank accounts are created again. I created the module bank_account_search to make this more reliable. This module could also be used separate from the import functions.

alexis-via:
I tried to used the module bank_statement_parse_camt. After figuring out that my file was camt.052 and not camt.053 and doing a few hacks to make it work with camt.052, I was very surprised to see the changes to the data model of the bank_statement_line and to the view of the bank_statement_line. This is probably fine when you ONLY use camt format, but when you use CAMT together with other formats (for other bank accounts in other banks), then it's a problem to add some fields and hide others. It also seems that bank_statement_parse has a lot of generic method that should belong to account_bank_statement_import : are they all complementary ?

It seems to me that "bank_statement_parse" is a sort of "framework" above the standard "account_bank_statement_import" framework... isn't it ?

NL66278:
Hello Alexis,

The bank_statement_parse is a framework above a framework in the sense
that it provides for easy conversion of the parsers written for 7.0 and
before. These parsers returned arrays of Bank Statement objects,
containing Bank Transaction objects. The convert_statements method added
to account_bank_statement_import does the heavy lifting of pressing the
information returned from these parsers into the format expected by the
framework backported from Odoo master.

Taking my conversion of the camt parser as a lead, one of our customers
provided the conversion of the mt940 parser in the space of two days.

I tested importing ofx files, camt files (from several banks) and mt940
files all in one database. This did not pose a problem.

I found the method used to automatically create bank-accounts lacking in
the originally backported code (before Odoo changed everything). It
still does not do a proper recognition of IBAN bank accounts. The
original code did not do a proper recognition of existing bank accounts
with a different formatting for the number. They did solve this at the
expense of needing a full table scan for the bank account number search.
That is why I have the bank_account_search module as a requirement for
the bank_statement_parse. It provides an indexed (and unique!) bank
number key, without all the formatting, allowing the user to format the
bank account in the user interface, just like they wish, while still
providing for fast searches.

As for the extra fields in bank_statement_line, their rationale is as
follows:
- data is for debugging purposes (to have the unformatted data that was
  used to create the line)
- transfer_type might be crucial information to identify where an amount
  came from (cash payment? cash withdrawal? direct debit?). Unfortunately
  the contents of these fields are not standardized between banks, but I
  think all, or nearly all banks have them.
- remote_owner is needed to make a transaction recognizable for the
  users when we do have a name, but the transaction has not been linked to
  an existing or new partner yet.
- remote_account will show the bank account information just as it was
  received in the transaction import file - unless a regular bank account
  could be linked or created. This one might not be strictly necesarry.

Another aspect of my code is that when a bank account is automatically
created, it stores all the partner data available. Later on, this can be
used to find the right existing partner, of provide the data to create a
new partner. See the bank_statement_link_partner module proposed for the
bank-statement-reconcile project.

I realise there is a lot to be said about all of this and I really
apreciate all remarks and questions.

Thanks for looking at the code!

Kind regards, Ronald

Op 19-02-15 om 23:22 schreef Alexis de Lattre:

NL66278:
@kvdb Hello I do not see quite to what function / line your remarks apply.

kvdb:
@NL66278 You should be able to click the link in the comment header and see the comment in code. I copied the links for you:
https://github.com/NL66278/bank-statement-import/commit/3d7802261c9e8abf0bd1cb701e24a96fe4cdf848#commitcomment-10222342 
https://github.com/NL66278/bank-statement-import/commit/3d7802261c9e8abf0bd1cb701e24a96fe4cdf848#commitcomment-10222327

NL66278:
@kvdb I updated the PR according to your remarks.

kvdb:
Thanks.

StefanRijnhart:
Can you resolve the conflict by rebasing against oca/8.0?

NL66278:
@StefanRijnhart  Conflict was only caused by pot file. So I took that from the main branch.

alexis-via:
Following my previous comment in this PR, I have developed an alternative module for CAMT import that supports both .052 and .053 and that only depends on account_bank_statement_import and NOT on bank_statement_parse (because I think that bank_statement_parse is a "framework above the framework", and I don't like that). It only has very few lines of code, so it is easy to modify and debug.

As I don't want to start a competition inside OCA for CAMT import module, I decided to publish this module outside OCA for the moment:
https://github.com/akretion/bank-statement-import-camt/tree/8.0/bank_statement_import_camt

If some OCA members support my initiative and want my CAMT module instead of the one in the PR, then I may propose it to OCA, but I won't take this initiative myself alone.

lmignon:
@alexis-via IMO, your module make sense and do what is expected by the new import framework backported from master. The code is simple, well tested and easy to maintain. Moreover, as it's closer to the new framework without requiring a 'meta framework' I'm favor to include yours.

NL66278:
Hello,

I will look into points made by Alexis and add camt.052 to my pull request.

Be aware that Alexis branch loses a lot of information from the camt
file that can be used to link partners to transactions or transactions
to open invoices. Therefore the two branches are not at all equivalent.

Also getting camt files in the form of zip files - standard with the
biggest bank in the Netherlands - is not supported by Alexis branch.

Kind regards,

Ronald

Op 25-03-15 om 09:42 schreef Laurent Mignon (ACSONE):

alexis-via:
@NL66278 
True, my camt module is not complete yet ; it's just a few weeks old !
I was not aware that you could get camt files in the form of zip file ; it doesn't seem to be the case in Germany (I'll ask my customer to confirm). Is there a particular "spec" for this ?

In the German camt file of my customer, I just have unstructured labels ; I don't have the bank account of the partner. So I use the module account_statement_completion_label_simple (https://github.com/akretion/bank-statement-reconcile-simple/tree/8.0/account_statement_completion_label_simple) to identify the partner from the label.

sbidoul:
Hi there,

I've not looked into details in this PR yet. So I have only a few high level comments at this stage:
- there is a lot of valuable effort here and many features in @NL66278's work that are potentially useful for all formats 
- @alexis-via's branch is appealing because it is lightweight, non-intrusive and orthogonal
- import of zip containing bank statements can and should be done independently of the format. @lmignon is working on it in this branch: https://github.com/acsone/bank-statement-import/tree/8.0-add-bulk-import
- I still believe we must carefully separate parsing functions from format-independent completion/reconciliation/matching functions; such features should go in OCA/bank-statement-reconcile (the 8.0 roadmap is still to write in that respect, perhaps a topic for a sprint during the Odoo conference?)
- in the 7.0 bank-statement-reconcile branch, there is a mechanism using serializable fields to store special fields from statement files in a way that does not change the bank statement line data model for each different format: it would be great if this approach could work in 8.0 too

@NL66278 do you think you could split your work in smaller PR's that isolate
- CAMT-specific parsing features
- any other "framework level" features that are useful for another formats
- completion/reconciliation features (in OCA/bank-statement-reconcile)

NL66278:
@sbidoul I will look into restructuring / streamlining my PR to split it up more.

There is ofcourse a direct dependency of reconciliation features (like the wizard to create a partner from the information in a bank transaction), on the data stored during the import. I tried to keep this distinction clear.

Later this day, or tomorrow I will look further into this. And give a more elaborate answer. Especially having two different solutions for handling zip files (which is completely working in my PR) seems to be a pity to me.

sbidoul:
@NL66278 regarding the zip files, I totally agree we need a single module for that. We did not know you had this feature in this PR - it was not visible without reading the code and github did not help because the diff was too large...

Let's try to see the merits of your approach and @lmignon 's and converge on the best.

NL66278:
Please update the CLA database. I have sent in a signed CLA at least a year ago.

NL66278:
@lmignon I looked into your code for bulk statement import and in itself it is very nice and clean. The problem is that is involves a rewrite of the base code (backported from master), splitting import_file and _import_file, that is very nice in itself too, but imcpmpatible with the idea of a backport.

gurneyalex:
@NL66278 the issue was a missing link between you github login and the CLA. I've now fixed this.

lmignon:
@NL66278 Thank you for the review. A PR has been submitted on master for a refactoring of the import framework and is under review. https://github.com/odoo/odoo/pull/5863. The same changes are ready to be proposed on OCA once the one on master will be merged.  (https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api)
The refactoring is required to avoid copy/paste of the code of the base class. I don't try to provide a replacement of the base import, just to enhance what is already done in the base import. An alternative would be to also redefine in my PR the import_file method but IMO it's very dangerous since the base implementation would not have been called and breaks modularity. (In your PR, what happens if an other addons has made changes to the import_file method?)
In my PR I try to provide a solution for one need: how to import a lot of bank statement files at once?
My answer is:
- accept zipfile as input for statement import,
- process each included file one by one by calling the normal import process in it's own transaction,
- display a report if some files parts of the archive are not properly processed.
  With the given answer it's also possible to mix different formats of bank statement files in the same archive.

Regards,

lmi

NL66278:
I proposed an alternative PR building on the work done by @lmignon.
See: https://github.com/OCA/bank-statement-import/pull/15

I will propose to use the zipfile functions and search functionality for bank accounts from Laurent, which were anyway very similiar to what I had in this PR, except that thanx to the refactoring that Laurent did it is no longer needed to overrule the whole base import.

kvdb:
I'd like to develop a new bank-statement importer for the Dutch ING creditcard statements (https://www.ingbanknetservice.nl). However, I'm not sure on which base to build it. Is there already an agreement on it? Or when can that be expected?

NL66278:
@kvdb No agreement has been reached. But I think the mainstream is to go with the new api version of the import framework. Unfortunately that has yet to be accepted by Odoo SA for master.

sbidoul:
@NL66278 same point of view here.

We have been trying to contact @qdp-odoo to see if the branch can be merged in odoo master, but no luck so far.

nbessi:
Should we close in this PR ?

gfcapalbo:
Found conflicts, needs rebasing.

NL66278:
As the approach based on the new api has been accepted by the community, I close this PR.  Even though it still contains some extra features, but I will provide these based on the new api as well.

PR #6, opened 2015-01-23
New backport from odoo/master

alexis-via:
This backport adds some new features:
- auto-detect account.journal for OFX files (I had to make a light change in the code to make it work in real life)
- block when trying to import lines that were already imported in another statement

Note that I disabled the auto-creation of account.journal, because I find it very hazardous.

This new backport also fixes bug #5 : no more manual creation of temporary file.

PR #7, opened 2015-02-05
8.0 Add module account_bank_statement_import_paypal

alexis-via:
This module is designed for the new bank statement system of the PR #6 

This module is certainly not perfect, in particular on the problems of multi-language. The problem is that the Paypal CSV files use strings in the language of the user... and I use a few of those strings in the parsing. But I don't see any easy solution for this...


pedrobaeza:
For the parsing question, I would use _() to also translate the string you are using. What do you think?

account_bank_statement_import_paypal/account_bank_statement_import_paypal.py
```diff
@@ -0,0 +1,201 @@
...
+        f = StringIO()
+        f.write(data_file)
+        f.seek(0)
+        transactions = []
+        i = 0
+        start_balance = end_balance = start_date_str = end_date_str = False
+        vals_line = False
+        user = self.pool['res.users'].browse(cr, uid, uid, context=context)
+        company_currency_name = user.company_id.currency_id.name
+        comission_total = 0.0
```
max3903: commission

alexis-via:
@pedrobaeza I think it's not as easy as that. For example: I use OpenERP in English because I prefer to use it in english to avoid all the bad French translation and to avoid the mix of translated and untranslated strings. But my Paypal bank statements are in French because Paypal knows I'm from France.

I think that the best solution is to have some _prepare method that can be inherited in a country-specific module. It is needed for:
- date parsing
- file encoding
- amount parsing (decimal separator)
- status of the line

i just implemented it. Tell me what you think.

pedrobaeza:
You can add then a parameter to say in which language is the file, and use .with_context(lang='...') in all parse calls.

alexis-via:
@pedrobaeza I'm really not sure that using translations is the good answer for this. The problem is that I don't have enough information on Paypal CSV bank statements to decide what is the best solution, that's why I propose to use the _prepare_*() solution for the moment.

pedrobaeza:
But this is also not a good option, because this forces to make subsequent account_bank_statement_import_paypal_xx for each language, and with no possibility of mixing paypal files from one language and another. We should provide a configurable mechanism.

sbidoul:
@alexis-via there is a pep8 issue:

`./account_bank_statement_import_paypal/account_bank_statement_import_paypal.py:164:25: W503 line break before binary operator`

And unicodecsv must be added to the travis config.

kvdb:
I've been using this PR with a couple of minor adjustments for localization purposes and a rebase on PR #15. It usually works fine, although one of the CSV exports couldn't be parsed.

I encountered this error:

```
Value '...' for the field 'commission' on line 14, cannot be converted to float 
```

Line 14 (I replaced some text, but only the sensitive data. Nothing related to those triple dot values):

```
"06-07-2014","19:16:35","GMT+02:00","ThinkGeek, Inc.","Order","Completed","USD","-132,80","...","...","info@foo.bar","paypal@thinkgeek.com","O-28941779A1111111111","","Non-Confirmed","","B19A6,promotion","","","","","","","","","","","","","","","5b711e1ae","","","...","Torenallee 20","","Eindhoven","NOORD-BRABANT","5617 BC","Netherlands","",
```

account_bank_statement_import_paypal/README.rst
```diff
@@ -0,0 +1,41 @@
+Import Paypal Bank Statements
+=============================
+
+This module allows you to import the Paypal CSV files in Odoo as bank statements.
```
kvdb, line 4: You might want to describe which kind of PayPal CSV file exports are accepted. I can download 3 types (all activity, balance affecting payments or completed payments) from PayPal of which sometimes only the balance affecting payments imports properly.

kvdb:
Another issue I noted: one transaction is showing up on the reconciliation screen as 'To British Pound : 5AL64185U4882084C" instead of a reference to "thetrainline". These are the CSV lines that lead up to it:

```
"21-09-2014","13:07:35","GMT+02:00","thetrainline","Express Checkout Payment Sent","Completed","GBP","-20,15","0,00","-20,15","info@foo.bar","ttl_paypal@thetrainline.com","2QM72241XA3211636","Non-U.S. - Verified","","","","0,00","","0,00","","","","","","","","","","","02D153823V1084457","2114079950","","","0,00","","","","","","","",
"21-09-2014","13:07:35","GMT+02:00","From Euro","Currency Conversion","Completed","GBP","20,15","0,00","20,15","","","0MD85557EX105261C","","","","","","","","","","","","","","","","","","2QM72241XA3211636","","","","20,15","","","","","","","",
"21-09-2014","13:07:35","GMT+02:00","To British Pound","Currency Conversion","Completed","EUR","-26,58","0,00","-26,58","","","5AL64185U4882084C","","","","","","","","","","","","","","","","","","2QM72241XA3211636","","","","1.027,69","","","","","","","",
"21-09-2014","13:07:35","GMT+02:00","PayPal","Temporary Hold","Removed","EUR","26,58","0,00","26,58","","","96K33589W03715300","","","","","","","","","","","","","","","","","","7TV55533PV057114D","","","","1.054,27","","","","","","","",
"21-09-2014","13:07:32","GMT+02:00","PayPal","Temporary Hold","Placed","EUR","-26,58","0,00","-26,58","","","7TV55533PV057114D","","","","","","","","","","","","","","","","","","02D153823V1084457","","","","1.027,69","","","","","","","",
```

train

zecke:
Changes to import German PayPal CSV:
- Match for 'Abgeschlossen' for complete statements.
- Check for 'Datum' instead of Date
- Numbers are 3.100,55.. and need to change
- balance is column 22 (not 34), transac_ref is column 18 (not 30).

adrienpeiffer:
Thank you for your work!

I tested this with several files. The format seems to be the same as you.

But I encounter some problems:
- Refund lines are ignored, or I think we need to import them.
- To work properly with base_transaction_ref module the id of the transaction should be in the name field of the bank statement line.
- Paypal_email_account is returned in _parse_file is treated as a bank account if that raises an exception.
- The thousand separator is not considered in the transformation in a float number.

I'll do a PR about it on your branch.

Thanks,
Adrien

alexis-via:
Just to make things clear:
1) I don't think OCA will ever merge this PR ; it there were some motivation of the reviewers to merge it, it would have been merged long ago.
2) I don't plan to work on this PR myself anymore.
3) these Paypal bank statement are a piece of !@*%$£§&#, without strict formatting, with a very strange behavior for transactions in other currencies, with stupid handling of languages and localisations, etc... really something that Paypal cannot be proud of. I really hope that Paypal will fix it one day and offer a better format for their bank statement... but I don't know if it will happen one day.

So feel free to take over this PR. If you think that the reviewers could maybe merge it one day, you can make some PR in this PR and I'll merge them immediately.

pedrobaeza:
As there's no plans to work on this from the author, I close. Feel free to reopen in case there's interest.

elicoidal:
@pedrobaeza @jbeficent Actually this coould be interesting to fix for the particular case of OCA Paypal Bank Statement.

Would it be worth the effort to fix it only for our particular case?

pedrobaeza:
Well, it's very specific to the downloaded format, but if you pass me a sample, I can check the coincidences. I have set up a variant of this one for an Spanish Paypal statement.

elicoidal:
Thanks I will private email you

PR #8, opened 2015-02-20
[IMP] account_bank_statement_import: POT and ES translation files

pedrobaeza:


yvaucher:
No idea about español, nevertheless :+1: technically speaking

antespi:
:+1:

PR #10, opened 2015-03-08
Add module account_bank_statement_import_be_bpost

alexis-via:
This module adds support for Bpost (Belgian Post) CSV bank statement files.

### DIFF
--- a/account_bank_statement_import_be_bpost/README.rst
+++ b/account_bank_statement_import_be_bpost/README.rst
@@ -1,0 +1,34 @@
+Import Bpost CSV bank statements
+================================
+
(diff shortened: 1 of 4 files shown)


pedrobaeza:
A better name would be _l10n_be_account_bank_statement_import_bpost_, following naming rule.

Don't you think this should be targeted on belgian localization, as I have made with spanish one: https://github.com/OCA/l10n-spain/pull/131.

alexis-via:
@pedrobaeza yes, you are certainly right. I'll move it there !

alexis-via:
Here is the new PR on l10n-belgium, with the new name : https://github.com/OCA/l10n-belgium/pull/18

PR #11, opened 2015-03-17
[FIX]Take a copy of the context before update

lmignon:
If the method `import_file` is called with the new API, the context is a frozendict and the update will raise `NotImplementedError: 'update' not supported on frozendict`
PR on master: https://github.com/odoo/odoo/pull/5798


pedrobaeza:
:+1:

laetitia-gangloff:
:+1:

sbidoul:
:+1:

PR #12, opened 2015-03-18
[FIX] bank_account created by the import must belong to the partner link...

lmignon:
...ed to the company of the provided journal

PR on master: odoo/odoo#5821

### DIFF
--- a/account_bank_statement_import/account_bank_statement_import.py
+++ b/account_bank_statement_import/account_bank_statement_import.py
@@ -198,9 +198,13 @@ class account_bank_statement_import(osv.TransientModel):
         # while 'counterpart' bank accounts (from which statement transactions originate) don't.
         # Warning : if company_id is set, the method post_write of class bank will create a journal
         if journal_id:
-            vals_acc['partner_id'] = uid
+            company_id = self.pool['account.journal'].browse(
+                cr, uid, journal_id, context=context).company_id.id
+            vals = self.pool['res.partner.bank'].onchange_company_id(
+                cr, uid, None, company_id, context=None)
+            vals_acc.update(vals.get('value', {}))
             vals_acc['journal_id'] = journal_id
-            vals_acc['company_id'] = self.pool.get('res.users').browse(cr, uid, uid, context=context).company_id.id
+            vals_acc['company_id'] = company_id
 
         return self.pool.get('res.partner.bank').create(cr, uid, vals_acc, context=context)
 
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Thanks :+1:

sbidoul:
I see the bug and I think your fix is correct.

But I'm not sure in which circumstances this bug is being triggered, because we don't want to autocreate company bank accounts (only counterpart bank accounts). I probably overlook something.

hbrunn:
@lmignon is this still WIP? And could you rebase please?

legalsylvain:
ping @lmignon

account_bank_statement_import/tests/test_import_bank_statement.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
NL66278, line 2: Please use shortened headers.

account_bank_statement_import/account_bank_statement_import.py
```diff
@@ -198,9 +198,13 @@ def _create_bank_account(self, cr, uid, account_number, journal_id=False, partne
         # while 'counterpart' bank accounts (from which statement transactions originate) don't.
         # Warning : if company_id is set, the method post_write of class bank will create a journal
         if journal_id:
-            vals_acc['partner_id'] = uid
+            company_id = self.pool['account.journal'].browse(
```
NL66278, line 201: Why not self.env and new API?

account_bank_statement_import/account_bank_statement_import.py
```diff
@@ -198,9 +198,13 @@ def _create_bank_account(self, cr, uid, account_number, journal_id=False, partne
...
+                cr, uid, journal_id, context=context).company_id.id
+            vals = self.pool['res.partner.bank'].onchange_company_id(
+                cr, uid, None, company_id, context=None)
+            vals_acc.update(vals.get('value', {}))
             vals_acc['journal_id'] = journal_id
-            vals_acc['company_id'] = self.pool.get('res.users').browse(cr, uid, uid, context=context).company_id.id
+            vals_acc['company_id'] = company_id
 
         return self.pool.get('res.partner.bank').create(cr, uid, vals_acc, context=context)
```
NL66278, line 209: Also here why not self.env and new API?

NL66278:
:+1:  With minor nitpicks and agreeing with the remark that this bug will only occur in those cases where autocreate of company bank accounts is done, which normally is not the case.

pedrobaeza:
This is conflicting, so I suppose this has been already improved/fixed meanwhile, and thus I close this. If not, please feel free to reopen fixing conflicts.

PR #13, opened 2015-03-21
FIX identify partner by res.partner.bank

alexis-via:
In account_bank_statement_import/account_bank_statement_import.py line 192, it creates a res.partner.bank with acc_number that has been "cleaned" with "replace(' ', '').replace('-', '')"

But, in the same file line 224, it searches res.partner.bank with identifying_string without doing the same "cleanup", so it doesn't find the res.partner.bank and therefore doesn't set the partner on the bank statement line.

This fix does the same "cleanup" on identifying_string, so that odoo can identify the partner by res.partner.bank.

### DIFF
--- a/account_bank_statement_import/account_bank_statement_import.py
+++ b/account_bank_statement_import/account_bank_statement_import.py
@@ -220,6 +220,7 @@ class account_bank_statement_import(osv.TransientModel):
                     bank_account_id = False
                     identifying_string = line_vals.get('account_number', False)
                     if identifying_string:
+                        identifying_string = identifying_string.replace(' ', '').replace('-', '')
                         ids = self.pool.get('res.partner.bank').search(cr, uid, [('acc_number', '=', identifying_string)], context=context)
                         if ids:
                             bank_account_id = ids[0]


pedrobaeza:
Why don't you refactorize the cleaning in a method and call it in both places for consistency?

Did you propose it upstream?

alexis-via:
Because the time I spend on Odoo/OCA bug fixing during my week-ends is limited. It's probably not the perfect fix, but it works.

pedrobaeza:
Yeah, but this is a future time-saving, and it's not too much time spending: make a method with the replace calls, and call this method ibn two places.

About the PR against Odoo upstream, it's mandatory on backports.

lmignon:
@alexis-via @pedrobaeza I've proposed a refactoring on upstream.  (https://github.com/odoo/odoo/pull/5863) The same refactoring is already back ported for OCA (https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api). I wait the review on upstream before proposing the changes on OCA. I'll fix the issue and add tests in the two repositories. (add a new method doing the cleanup)

pedrobaeza:
Cool, thanks for your work, @lmignon. This improvement works as is on 8.0?

lmignon:
@pedrobaeza yes in 8.0 the branch https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api is the backport of my PR on upstream and all tests succeed on travis. I use this branch for a future new addons to use for bulk import. https://github.com/acsone/bank-statement-import/tree/8.0-add-bulk-import

lmignon:
@pedrobaeza @alexis-via After digging the code, I'm not confident with what's actually done and what's proposed as change.  It's a recurring problem to search a partner by account number. In the completion framework for Odoo 7.0, the same issue has been fixed by defining a new method on the 'res.partner.bank' model. IMO a better solution would be to define a new computed field on 'res.partner.bank' to store the account_number unformatted and to extend the search method to use this field when searching on account_number. In this way we would have the same behavior everywhere.

pedrobaeza:
Yeah, having a computed field + modifying search method seems a good solution for me too. This overriding should replace current account number field in queries by the computed one, and the term queried by the cleaned version.

lmignon:
@pedrobaeza @alexis-via my proposal on master is here https://github.com/acsone/odoo/commit/82bef3772614d33dcd8ee7dd4ed6a07bbb26fc82 and the one for  OCA. https://github.com/acsone/bank-statement-import/commit/cc1f2f354d13fbac80cc8eb98f044d4301701d90.  A PR already exists on master. I wait for review on this one before proposing my branch of bank-statement-import to OCA. Meanwhile, it 'may be interesting to merge the current PR.

pedrobaeza:
@lmignon, I have made some comments.

Meanwhile, :+1: for this PR.

lmignon:
:+1:

brother-bernard:
:+1: It works perfectly !

lmignon:
Hi, be careful, I just noticed a potential issue in the code. The account number is not unique. IOW, more than one partner can be defined with the same account number. The import never check if more than one partner has been found for the given account number. This may result to link and reconcile the line with the wrong partner.

sbidoul:
OTOH, when you create an IBAN account manually, Odoo adds spaces.

sbidoul:
https://github.com/OCA/bank-statement-import/pull/13#issuecomment-84943793 is true but unrelated to this PR.

https://github.com/OCA/bank-statement-import/pull/13#issuecomment-86417885 is true too so this fix works only for account numbers that are automatically added by the statement import mechanism.

A more complete solution is in https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api (which is the backport of odoo/odoo#5863).

In the meantime there are three :+1: here. I add mine and merge.

PR #14, opened 2015-03-26
8.0 add bulk import

lmignon:
This module allows you to bulk import files containing bank statements. (files must be packed as a singled zip archive) 
This PR also includes the refactoring of the bank-statement-import addons proposed for merge on upstream https://github.com/odoo/odoo/pull/5863. The same changes are also backported in an isolated branch for 8.0 https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api. (this PR is based on #19 )


lmignon:
@sbidoul IMO the label 'work in progress' can be removed. The branch has been rebased on #19

sbidoul:
@anthony-muschang @lmignon it would be great to have the number of successfully imported and errored files in the resulting wizard.

nbessi:
How to manage the use case of statement files that are tar/zip but not bulk. 
Eg. postfinance provide a tar.gz by statement containing scans, signature etc

lmignon:
@nbessi good catch

NL66278:
I think the code in this PR is very important to provide a complete import framework. I have three issues with it:
- I really think the ability to import zip files should be part of the base code, rather then being part of a separate module;
- Mainly because in the present solution the _import_file method is overriden and then called recursively. This means other modules overriding _import_file will never be certain wether they have the whole file selected by the user, or only one of the files included in the zip file. It all depends on the actual order the modules are loaded;
- Then this module includes code to direct the user to an error overview when errors occur. This is very good generally, not depended on wether the user needs batch import or not. So this module is including sets of functionality that can stand on their own. But even here, I think it should just be part of the base code. The present redirection from importing a statement to immediate reconciliation wizard - taken from odoo master - IMHO is very bad. A user importing a file will expect to see the result of the import. And in many organisations, the people doing the import (likely to be more technically educated) will not be the people responsible for reconciliation.

In summary: please make the enhancements in this module part of the base code and don' t use a recursive call in _import_file.

hbrunn:
@lmignon could you rebase your work?

lmignon:
@hbrunn This module must be fixed / rewritten according to the comments expressed during the review process.

lmignon:
@sbidoul @hbrunn I close this one since this PR must be rewritten according to https://github.com/OCA/bank-statement-import/pull/14#issuecomment-107644312

PR #15, opened 2015-04-02
8.0 import parsers new api therp

NL66278:
Add camt and mt940 import, based on the refactored master proposed by Laurent Mignon, and including the changes I proposed to support multiple accounts and currencies per file.


NL66278:
This PR is an alternative approach to https://github.com/OCA/bank-statement-import/pull/4

However note that it is based on changes to master that have yet to be accepted by Odoo SA

nbessi:
Hello,

Tanks for the work.

Can you please remove the unported directory. You should focus on one point at a time in a PR.
Migrate core to new API should be one PR migrate exisiting importer to new API should be one PR. 
Etc. No body want to review such a huge PR that is hard to get into. 

 Also if you are not ready you should mark the PR as WIP. 

If you do not release fast code chunk especially on a modular parser there will quickly be divergence. People how want to integrate there own parser will start their work it is going to be merged and it will not be compatible with your PR.

My two cents

Nicolas

gurneyalex:
I have not followed closely the discussions about bank-statement-import. Is there a summary document of the plan available somewhere? 

Where are all these **unported** files coming from?

@StefanRijnhart your feedback is welcome on this.

nbessi:
@NL66278 Can you give a pointer to the PR on the Odoo core?

NL66278:
@gurneyalex @nbessi I will remove the **unported** directory. These were originally part of banking/bank-payment, but had to do with importing statements. Therefore moved to here. 

I still have a few issues with finishing all the work. One is that there is uncertainty about what direction Odoo SA will take with the banking addons. The new api rewrite has not (yet?) been accepted. Neither my proposal to make it compatible with multiple bank accounts and currencies per import file. Maybe we should no longer bother with what Odoo SA is doing and go our own way?

lmignon:
@nbessi Everything is in the description of https://github.com/OCA/bank-statement-import/pull/14 

> This PR also includes the refactoring of the bank-statement-import addons proposed for merge on upstream odoo/odoo#5863. The same changes are also backported in an isolated branch for 8.0 https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api. (this PR is based on the last one)

NL66278:
Having said that, the import  in this proposal (based on the new api) is fully functional. But still lacks to features that were in my proposal based on the old api (PR4): keeping the history of imports, and bulk import from zipfiles.

nbessi:
@lmignon thanks for your answer.  I'm shomehow lost I want to add a new parser for Postfiance and I do not get which branch should be used as base. I'm probably not the only one, and  I guess it is a perfect situation for divergence. If I base my work on PR 15 I guess my work will compatible with the PR 14.

It will be nice to settle the situation

lmignon:
@nbessi IMO, you should base you work on 8.0. That's what I did for an other parser https://github.com/OCA/l10n-belgium/pull/20. The PR #14 require a change in the base API therefore I first proposed a PR on upstream.

nbessi:
@lmignon thanks I guess it is wiser :smile:

lmignon:
@nbessi I'll create an issue to summarize the current situation

NL66278:
@lmignon Did you ever look at my PR on your odoo 9.0/master branch to support multiple accounts and currencies in import? https://github.com/acsone/odoo/pull/2

lmignon:
@nbessi https://github.com/OCA/bank-statement-import/issues/18 is an attempt to summarize the current situation

lmignon:
@NL66278 Thank you for the PR. I read your proposal. For now I'm still trying to have the feedback from  @qdp-odoo before including others changes in the PR on upstream. I prefer to limit changes in the PR to ease the review/merge.

gfcapalbo:
Found conflicts, needs rebasing.

NL66278:
@gfcapalbo Conflicts with what? You can see on this page that the branch can be merged withouth conflict.

gfcapalbo:
I'm sorry, wrong page.

rruebner:
Hi, is there any hint that this PR will be merged in the near future because we want to use the parser modules for another mt940 parser.

pedrobaeza:
Please rebase and I'll review it ASAP.

lmignon:
@pedrobaeza A PR for the migration to the new API has been provided https://github.com/OCA/bank-statement-import/pull/19

pedrobaeza:
New API reviewed

hbrunn:
@NL66278 now that #19 is merged, can you rebase your work on that?

NL66278:
@hbrunn Yes I will do that.

StefanRijnhart:
./bank_statement_parse_mt940/mt940.py:115:80: E501 line too long (80 > 79 characters)

account_bank_statement_import/account_bank_statement_import.py
```diff
@@ -89,90 +112,92 @@ def _import_file(self, data_file):
...
+        journal_id = self._get_journal(currency_id, bank_account_id)
+        # By now journal and account_number must be known
+        if not journal_id:
+            raise Warning(_('Can not determine journal for import.'))
         # Prepare statement data to be used for bank statements creation
-        stmts_vals = self._complete_stmts_vals(
-            stmts_vals, journal_id, account_number)
-        # Create the bank statements
-        return self._create_bank_statements(stmts_vals)
+        statement = self._complete_statement(
```
lmignon: Why do you rename stmts_vals by statement. IMO it's confusing. With the new API with such a name, I expect to recieve a browse_record but in fact it's a dict.

bank_statement_parse/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import parserlib
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: PLZ remove vim line

bank_statement_parse/__init__.py
```diff
@@ -0,0 +1,30 @@
+# -*- encoding: utf-8 -*-
+"""Classes and models to parse bank statements and import them into Odoo."""
+##############################################################################
+#
+#    Copyright (C) 2014 Therp BV - http://therp.nl.
+#    All Rights Reserved
```
lmignon: All Rights Reserved is deprecated

bank_statement_parse/__openerp__.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Therp BV <http://therp.nl>.
+#
+#    All other contributions are (C) by their respective contributors
+#
+#    All Rights Reserved
```
lmignon: PLZ remove All Rights Reserved

bank_statement_parse_camt/account_bank_statement_import.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            _logger.debug("Try parsing with camt.")
+            return convert_statements(parser.parse(data_file))
+        except ValueError:
+            # Not a camt file, returning super will call next candidate:
+            _logger.debug("Statement file was not a camt file.")
+            return super(AccountBankStatementImport, self)._parse_file(
+                cr, uid, data_file, context=context)
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: PLZ remove vim line

bank_statement_parse_camt/account_bank_statement_import.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+"""Add process_camt method to account.bank.statement.import."""
+##############################################################################
+#
+#    Copyright (C) 2013 Therp BV (<http://therp.nl>)
+#    All Rights Reserved
```
lmignon: idem

bank_statement_parse_camt/__openerp__.py
```diff
@@ -0,0 +1,35 @@
+##############################################################################
+#
+#    Copyright (C) 2013 Therp BV (<http://therp.nl>)
+#    All Rights Reserved
```
lmignon: idem

bank_statement_parse/README.rst
```diff
@@ -0,0 +1,6 @@
+
```
lmignon: README.rst must be formatted as https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

bank_statement_parse_camt/README.rst
```diff
@@ -0,0 +1,3 @@
+Module to import SEPA CAMT.053 Format bank statement files.
```
lmignon: README.rst must be formatted as https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

bank_statement_parse/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Bank Statement Import Parse',
+    'version': '0.5',
+    'license': 'AGPL-3',
+    'author': 'Banking addons community',
```
lmignon: PLZ add OCA in the list of author

``` python
    'author': 'Odoo Community Association (OCA)',
```

bank_statement_parse/parserlib.py
```diff
@@ -0,0 +1,115 @@
...
+        self.remote_bank_bic = False  # bic of other party's bank
+        self.provision_costs = False  # costs charged by bank for transaction
+        self.provision_costs_currency = False
+        self.provision_costs_description = False
+        self.error_message = False  # error message for interaction with user
+        self.storno_retry = False
+        # If True, make cancelled debit eligible for a next direct debit run
+        self.data = ''  # Raw data from which the transaction has been parsed
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: plz remove vim line

bank_statement_parse_camt/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'CAMT Format Bank Statements Import',
+    'version': '0.3',
+    'license': 'AGPL-3',
+    'author': 'Therp BV',
```
lmignon: Add OCA in the list of author

``` python
    'author': 'Therp BV,'
              'Odoo Community Association (OCA)',
```

bank_statement_parse_camt/camt.py
```diff
@@ -0,0 +1,245 @@
...
+        ns = root.tag[1:root.tag.index("}")]
+        self.check_version(ns, root)
+        statements = []
+        for node in root[0][1:]:
+            statement = self.parse_statement(ns, node)
+            if len(statement.transactions):
+                statements.append(statement)
+        return statements
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: remove vim line

bank_statement_parse_camt/tests/test_import_bank_statement.py
```diff
@@ -0,0 +1,68 @@
...
+            'Start balance %f not equal to 15568.27' %
+            statement_obj.balance_start
+        )
+        self.assertTrue(
+            abs(statement_obj.balance_end_real - 15121.12) < 0.00001,
+            'Real end balance %f not equal to 15121.12' %
+            statement_obj.balance_end_real
+        )
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: remove vim line

bank_statement_parse_mt940/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'MT940 Bank Statements Import',
+    'version': '1.1',
+    'license': 'AGPL-3',
+    'author': 'Therp BV',
```
lmignon: Add OCA in the list of author

``` python
    'author': 'Therp BV,'
              'Odoo Community Association (OCA)',
```

bank_statement_parse_mt940/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+    'website': 'https://github.com/OCA/bank-statement-import',
+    'category': 'Banking addons',
+    'description': '''
```
lmignon: Put the description in a README.rst file.

bank_statement_parse_nl_ing_mt940/README.rst
```diff
@@ -0,0 +1,25 @@
+Import MT940 IBAN ING Bank Statements
```
lmignon: Some sections are missing see  https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

bank_statement_parse_nl_ing_mt940/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'MT940 IBAN ING Format Bank Statements Import',
+    'version': '0.3',
+    'license': 'AGPL-3',
+    'author': 'Therp BV',
```
lmignon: Add OCA in the list of author

``` python
    'author': 'Therp BV,'
              'Odoo Community Association (OCA)',
```

bank_statement_parse_nl_ing_mt940/mt940.py
```diff
@@ -0,0 +1,69 @@
...
+        transaction = self.current_transaction
+        # If we have no subfields, set message to whole of data passed:
+        if not subfields:
+            transaction.message = data
+        else:
+            handle_common_subfields(transaction, subfields)
+        # Prevent handling tag 86 later for non transaction details:
+        self.current_transaction = None
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: remove vim line

bank_statement_parse_nl_ing_mt940/tests/test_import_bank_statement.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+
+    def test_statement_import(self):
+        """Test correct creation of single statement."""
+        self._test_statement_import(
+            'test-ing.940', 'NL77INGB0574908765-2014-02-20',
+            662.23, 564.35
+        )
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: remove vim line

bank_statement_parse_nl_rabo_mt940/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "MT940 import for dutch Rabobank",
+    "version": "1.1",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "complexity": "normal",
+    "description": """
```
lmignon: Put the description in a README.rst file.

bank_statement_parse_nl_rabo_mt940/account_bank_statement_import.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            _logger.debug("Try parsing with MT940 RABO.")
+            return convert_statements(parser.parse(data_file))
+        except ValueError:
+            # Returning super will call next candidate:
+            _logger.debug("Statement file was not a MT940 RABO file.")
+            return super(AccountBankStatementImport, self)._parse_file(
+                cr, uid, data_file, context=context)
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: remove vim line

bank_statement_parse_nl_rabo_mt940/mt940.py
```diff
@@ -0,0 +1,89 @@
...
+            # Use subfields for transaction details:
+            if 'NAME' in subfields:
+                transaction.remote_owner = ' '.join(subfields['NAME'])
+            if 'ADDR' in subfields:
+                # Do NOT join address fields, array is expected on other code!
+                transaction.remote_owner_address = subfields['ADDR']
+        # Prevent handling tag 86 later for non transaction details:
+        self.current_transaction = None
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: remove vim line

bank_statement_parse_nl_rabo_mt940/tests/test_import_bank_statement.py
```diff
@@ -0,0 +1,81 @@
...
+            'Start balance %f not equal to 4433.52' %
+            statement_obj.balance_start
+        )
+        self.assertTrue(
+            abs(statement_obj.balance_end_real - 4798.91) < 0.00001,
+            'Real end balance %f not equal to 4798.91' %
+            statement_obj.balance_end_real
+        )
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
lmignon: remove vim line

NL66278:
@lmignon @hbrunn @StefanRijnhart  And others. I think I have all remarks taken care off.

bank_statement_parse_nl_rabo_mt940/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+{
+    "name": "MT940 import for dutch Rabobank",
+    "version": "1.1",
+    'author': 'Odoo Community Association (OCA), Therp BV',
+    'website': 'https://github.com/OCA/bank-statement-import',
+    "category": "Account Banking",
+    "depends": [
+        'bank_statement_parse_mt940'
+    ],
+    "data": [],
```
lmignon: remove empty entries

lmignon:
@NL66278 a vim line remains in bank_statement_parse/**init**.py

bank_statement_parse_nl_rabo_mt940/__openerp__.py
StefanRijnhart: Thank you for your vigilant review! With regards to this particular comment, I was wondering what the rationale is for removing empty entries from the manifest file. A lot of people that I know use some sort of scaffolding for the Odoo module and manifest file structure, which leads to these entries. I don't remember any review in which this was a problem. Can we be more lenient on this point?

lmignon: @StefanRijnhart yes we can but we should at least remove the entries no more supported in V8 (js, css)

StefanRijnhart: Ah yes, agreed with removing obsolete entries.

NL66278:
@lmignon @StefanRijnhart  New comments being taken care off. Also standardized copyright headers, and category.

lmignon:
@NL66278 Thank you for your HARD work on this PR. I like the changes made on the `account_bank_statement_import`. These improve the extensibility of the import framework! :+1: for these changes.

IMO bank_statement_parser is useless. The new level between the parser and the import logic introduced by this new addons introduces some limitations/issues.
- It's memory consuming: The same data is duplicated in memory at least 1 more time (1 time as parsed line, 1 time as BankStatement and BankStatementTransaction objects, 1 time as dict)
- It's time consuming: Data are iterated one more time (when converted from BankStatement or BankStatementTransaction to dict)
- It restrict the extensibility of the parser (If a parser provides one additional data, the BankStatement or BankStatementTransaction must be modified as well as the convertion methods.
  I don't see the added value of all the code in this addons. May be I'm wrong, comments from others are welcome.

Once again thank you for all this hard work!

kvdb:
@NL66278 first of all, a big thank you for taking care of the Dutch ING import functionality while porting the parsing framework. I really look forward to using it from the 'official' OCA branch!

Howver, I encountered a problem while testing it. When using this 8.0_import_parsers_new_api_therp branch instead of the OCA bank-statement-import 8.0 branch, I get the following error when running openerp with the `--update=all` parameter:

```
2015-06-19 09:38:10,976 1 INFO dc openerp.modules.loading: loading account/account_view.xml
2015-06-19 09:38:11,903 1 ERROR dc openerp.addons.base.ir.ir_ui_view: Field `transfer_type` does not exist
```

My guess is that a small migration script is needed because my database doesn't have the field yet? Or perhaps I'm the only one having this problem due to testing out earlier revisions of the modules?

NL66278:
@kvdb Normally running with --update=all will not give problems. Like you already suspected, it has to do with testing out earlier versions. You can use a new database, or manually (through sql) delete the views added by the earlier versions of the modules.

The actual field transfer_type was for the moment removed/deleted, because not used at present anywhere. However it might be added back later, because it will for instance show the user wether a transaction was a cash withdrawal, or a direct debit, or a regular transfer. All usefull information of course when reconciling with account moves (new or existing) or open invoices....

NL66278:
@lmignon I rewote the objects BankTransaction and BankStatement to at least do away with the extra iterations and memory consumption. The big advantage of having those objects - apart from not having to completely rewrite the existing parsers - is the ability to use dot access instead of keyword access. And being able to check with pylint etc. if the right attributes are being used, instead of having strange problems occuring, or not being noticed, when for instance making typing errors.

For instance I can use
transaction.ererf = 'some reference' 
instead of
transaction['ref'] = 'some reference'
.
And using
statement.stattement_id = 'some id'
will immediately throw an error (notice the double ' t') while
statement['stattement_id'] = 'some id'
will not find anything to complain about.

Rather then removing those objects, I think a case can be made for using them throughout. Also because having an array of dictionaries, including an attribute that again is an array of dictionaries - statements[0].transactions[2]['amount'] IMHO is stretching how we should use those datastructures.

Thanks for your thorough review and constructive criticisms.

lmignon:
@NL66278 I agree, dot access is more natural and improve the code readability. I wonder if it's not possible to replace your new layer by instantiating directly the model...

```
...
statements =self.env['account.bank.statement'].new() 
...
for line in lines:
    statement = self.env['account.bank.statement'].new()
    statement.xyz = line['field']
    statements += statement
return statements
...
statements.create()
...
```

I dream :smile:

:+1:

kvdb:
@NL66278, removing the outdated view solved the issue, thanks.

bank_statement_parse_nl_ing_mt940/mt940.py
```diff
@@ -0,0 +1,66 @@
...
+        self.mt940_type = 'ING'
+
+    def handle_tag_61(self, data):
+        """get transaction values"""
+        super(MT940Parser, self).handle_tag_61(data)
+        re_61 = self.tag_61_regex.match(data)
+        if not re_61:
+            raise ValueError("Cannot parse %s" % data)
+        parsed_data = re_61.groupdict()
+        self.current_transaction.transferred_amount = (
```
kvdb: Refactor `transferred_amount` to `amount` (and please validate the amount of a transaction in the test, because it didn't get caught). Without the change, imported statements contain no amounts.

NL66278: actually I accidentally renamed attribute transferred_amount to amount in BankTransaction object. I have corrected that now. I did add the test as you suggested.

kvdb:
@NL66278 thanks, that's even more testing than I hoped for!

account_bank_statement_import/account_bank_statement_import.py
```diff
@@ -69,15 +72,35 @@ def import_file(self):
...
-        """ Create bank statement(s) from file
-        """
+        """ Create bank statement(s) from file."""
         # The appropriate implementation module returns the required data
-        currency_code, account_number, stmts_vals = self._parse_file(data_file)
-        # Check raw data
-        self._check_parsed_data(stmts_vals)
+        statement_ids = []
+        notifications = []
+        statements = self._parse_file(data_file)
```
hbrunn: @NL66278  This function's return value is different in upstream: https://github.com/OCA/bank-statement-import/blob/8.0/account_bank_statement_import/account_bank_statement_import.py#L101, https://github.com/odoo/odoo/blob/master/addons/account_bank_statement_import/account_bank_statement_import.py#L81

Is this an artifact of rebasing? We surely don't want to change this if not absolutely necessary

account_bank_statement_import/account_bank_statement_import.py
```diff
@@ -89,90 +112,92 @@ def _import_file(self, data_file):
...
+                    (e.g: 'KBC-INVESTERINGSKREDIET 787-5562831-01')
+                - 'date': date
+                - 'amount': float
+                - 'unique_import_id': string
+                -o 'account_number': string
+                    Will be used to find/create the res.partner.bank
+                    in odoo
+                -o 'note': string
+                -o 'partner_name': string
+                -o 'ref': string
```
hbrunn, line 163: see above

account_bank_statement_import/account_bank_statement_import.py
NL66278: Hello Holger, this is a concious change. Parsing can return data for multiple statements, possibly for different currencies and bank accounts.

hbrunn: @NL66278 thanks for the explanation. Are @StefanRijnhart, @lmignon, @nbessi et al aware of this part?

lmignon: @NL66278  @hbrunn I did not pay enough attention to this point. This change breaks the current API and therefore the parsers developed in some localisations (https://github.com/acsone/l10n-belgium/blob/8.0-add-account_bank_statement_import_coda/account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py#L62).
To keep the backward compatibility you can may be define a new function implemented by some specifics parsers `_parse_multi_smts_file` and refactor the code as

``` python
    ....
    def __parse_file(self, data_file):
        return False

    def _parse_multi_stmt_file(data_file):
        return []
    ....
    @api.model
    def _import_file(self, data_file):
        ....
        statements = []
        stmt_vals = self._parse_file(data_file)
        if stmt_vals:
            statements = [stmt_vals]
        statement.extend(self._parse_multi_stmt_file(data_file))
        ...
```

NL66278: @lmignon I do not quite see how your refactoring would work. It is still having _parse_file with the same return value that I proposed. Or do I miss something?
@lmignon @hbrunn I am not quite agreed that the "existing API" argument is valid, as the whole banking import framework is quite new and everybody knows it was/is still in flux. Look at the whole refactoring for the 8.0 API. Then, adapting to the (IMHO better) return value in many cases will only need two or three lines of codes. Look at the changes I made to the ofx and qif import. They are absolutely minimal.

lmignon: @NL66278 In m y proposal, the _parse_file stays as in upstream. New parser able to return multiple statements, for different currencies and bank accounts should implement `_parse_multi_stmt_file`.

hbrunn: I fear those changes might come back to haunt us. While it's true everything might change in upstream tomorrow, the code is at it is today and I think we should stick to its API. Or change upstream, and then adapt to that.

How about: Leave signatures as they are, but check keys `%currency_code` and `%account_number` (or whatever other prefix makes it unlikely for this to be a column name) in the statement dict which override the according value from the tuple.

NL66278:
@lmignon @hbrunn I changed the code to support both old and new style results from the parse. Reverted changes on ofx and qif parse to return the old style, in order to test with both styles.
As a sidenote: there is not really anymore an upstream for this code, as we agreed no longer to follow the changes (or lack of them) in master.

hbrunn:
If there are localized modules making use of the current api, they might choke on your renamed functions and changed signatures, but your changes make sense to me, so :+1:

lmignon:
:+1:

bank_statement_parse_camt/account_bank_statement_import.py
```diff
@@ -0,0 +1,43 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import logging
+from openerp import models
+from .camt import CamtParser as Parser
+
+
+_LOGGER = logging.getLogger(__name__)
```
sbidoul: The tradition is to use _logger (lowercase). Not blocking for me.

bank_statement_parse_camt/account_bank_statement_import.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _parse_file(self, cr, uid, data_file, context=None):
+        """Parse a CAMT053 XML file."""
+        parser = Parser()
+        try:
+            _LOGGER.debug("Try parsing with camt.")
+            return parser.parse(data_file)
+        except ValueError:
+            # Not a camt file, returning super will call next candidate:
+            _LOGGER.debug("Statement file was not a camt file.")
```
sbidoul: Adding exc_info=True to this log statement would help diagnosing parse errors.

bank_statement_parse/parserlib.py
```diff
@@ -0,0 +1,228 @@
...
+        # Fill default id, but might be overruled
+        transaction['unique_import_id'] = (
+            self.statement_id + str(len(self['transactions'])).zfill(4))
+        return transaction
+
+    def __init__(self):
+        super(BankStatement, self).__init__()
+        self['transactions'] = []
+        self.statement_id = ''
+        self.local_account = ''
```
sbidoul: Initializing local_account here is not necessary since the property getter/setter map it to account_number.

bank_statement_parse_mt940/mt940.py
```diff
@@ -0,0 +1,254 @@
...
+        Depending on the bank, there might be multiple 62F tags in the import
+        file. The last one counts.
+        """
+        stmt = self.current_statement
+        stmt.end_balance = str2amount(data[0], data[10:])
+        stmt.date = datetime.strptime(data[1:7], '%y%m%d')
+        stmt.statement_id = '%s-%s' % (
+            stmt.local_account,
+            stmt.date.strftime('%Y-%m-%d'),
+        )
```
sbidoul: In the examples I have, I get a statement number in tag 28C. I thought preserving it in statement_id, but this part overwrites it. Perhaps this should be done only if statement_id has not been set during parsing?

bank_statement_parse/parserlib.py
```diff
@@ -0,0 +1,228 @@
...
+        self.remote_owner_postalcode = False  # other parties zip code
+        self.remote_owner_country_code = False  # other parties country code
+        self.remote_bank_bic = False  # bic of other party's bank
+        self.provision_costs = False  # costs charged by bank for transaction
+        self.provision_costs_currency = False
+        self.provision_costs_description = False
+        self.error_message = False  # error message for interaction with user
+        self.storno_retry = False
+        # If True, make cancelled debit eligible for a next direct debit run
+        self.data = ''  # Raw data from which the transaction has been parsed
```
sbidoul: Is this **init** method still necessary now that everything is handled in properties mapping to dict entries?

bank_statement_parse/parserlib.py
```diff
@@ -0,0 +1,228 @@
...
+    @property
+    def remote_account(self):
+        """property getter"""
+        return self['account_number']
+
+    @remote_account.setter
+    def remote_account(self, remote_account):
+        """property setter"""
+        self['account_number'] = remote_account
+
```
sbidoul: Can you add a `note` property? It's part of account.bank.statement.line and is displayed in the :information_source: icon in the reconciliation screen so it's useful to store all data found in the electronic statement that could not be mapped to other fields but may nevertheless be useful for humans to determine what the transaction is about.

sbidoul:
@NL66278 I made https://github.com/NL66278/bank-statement-import/pull/5 with a few minor improvements covering some of my comments above. Can you have a look?

sbidoul:
@NL66278 I've finally done some in-depth testing by implementing a prototype for the Luxembourg MT940 format. Overall, things go along quite well. Thanks for this huge work!

I have two minor annoyances and one things that bothers me more.
- ideally, the modules should be named account_bank_statement_import_... to be in line with the other formats 
- I'm not sure that bank_statement_parse has a real added value: I've the feeling parserlib makes the code slightly more difficult to read because it uses a different vocabulary than the native Odoo field names for statements and transaction
- the thing that really bothers me is the numbering of the statements coming out of the parsers (IBAN-STATEMENT DATE). Since such an identifier later becomes part of the journal entry numbers it has a strange look in the general ledger. I understand this identifier format is chosen to have unique statement numbers? Would you see another approach? I personally like JOURNAL CODE/YEAR/STATEMENT NUMBER, which matches the Odoo default number for journal entries. And we normally always have statement numbers in the statement files. What do you think?

bank_statement_parse/parserlib.py
NL66278: pylint will complain when using self with an attribute not set in init.

bank_statement_parse_camt/account_bank_statement_import.py
NL66278: It is a conflict between two conventions. The more general is to use all uppercase for constants (of course python has no real constants, but variables considered to be constant), the odoo one, I agree is to use _logger. I accepted your suggestion to use _logger.

bank_statement_parse_camt/account_bank_statement_import.py
NL66278: Accepted your suggestion on this point as well. A pity that odoo master - which was backported as the base for all imports - is using exceptions to go through the different formats, although it is quite normal for a customer to have different formats for different banks, or even for different downloads for the same bank.

bank_statement_parse_mt940/mt940.py
NL66278: I will add a test to only replace empty statement_id' s. In my examples 28C mostly contains statement_id's made up of one or more zeroes. I will consider those to be empty.

NL66278:
@sbidoul I looked into your comment. Just the following remarks:
1. I am not in principle opposed to adding account_ to the module names, but I do feel it does not add any extra information to the name, and does lead to incredibly long names like account_bank_statement_parse_nl_rabo_mt940. What are other people thinking?
2. The terminology used is the terminology from the 7.0 parsers. It might not be essential, but it made migration a lot easier.
3. I hope your concern is diminished by the possibility to use the statement_id from the import file - if there is a meaningfull statement_id present. I do not quite see how I could inject the journal code into this, as the journal to be used is only determined (or even created) after parsing has completed. But maybe I am just missing something.

sbidoul:
> 1. I am not in principle opposed to adding account_ to the module names,
>    but I do feel it does not add any extra information to the name, and does
>    lead to incredibly long names like
>    account_bank_statement_parse_nl_rabo_mt940. What are other people thinking?
> 
> It's not only adding account_, it's making sure that all bank statement
> import module are named according to the same prefix. I would find it more
> consistent.
> 1. The terminology used is the terminology from the 7.0 parsers. It might
>    not be essential, but it made migration a lot easier.
> 
> I understand the migration point. Do you have other modules that rely on
> parserlib.py? Otherwise, I would suggest to move parserlib.py to the mt940
> base module, and merge this _without the camt module_. So we get the
> refactoring merged and we have mt940, and we take our time to review camt
> and the respective merit of your implementation and @alexis-via's. Given
> that CAMT is the module that will be most used in the future it's worth
> taking more time to get it in the best possible shape.
> 1. I hope your concern is diminished by the possibility to use the
>    statement_id from the import file - if there is a meaningfull statement_id
>    present. I do not quite see how I could inject the journal code into this,
>    as the journal to be used is only determined (or even created) after
>    parsing has completed. But maybe I am just missing something.
> 
> Absolutely. I also don't see how to get the journal code at the parser
> level. And I've also seen that the NL ING MT940 does not give sequence
> number (sic). So I'm totally ok if you can leave statement_id as is if the
> parser gave it a value.

If you wish I can do another PR towards your branch (but maybe not before
mid august).

Thanks again!

NL66278:
@sbidoul I think we are agreed on point 3 by now. May I propose a compromise? Let us rename all modules to the account_ prefix, so satisfying your request to have all import modules starting with the same name. That is point 1. As for point 2, for a start the camt implementation proposed by @alexis-via was a lean and mean proof of context, but not feature complete. After the rightfull criticisms from @alexis-via of the original camt module, I streamlined that one a lot. Also the parserlib can no longer be considered a separate "framework"  which I agree it originally resembled. But as it stands now it definitely IMHO helps in streamlining the based on code, with the dot member access it enables for attributes. And the checking it gives in pylint for any mistyped attribute names, which is not possible with the pure dictionary approach. To me it would not seem logical to put the parserlib in the mt940 is it is also used in the camt module. Maybe @alexis-via can you give your opinion on where you would like to go with the camt modules by now?
By the way, I will only be able to work on the banking modules also from the second half of august. So maybe we can have point 2 implemented as it is now in the PR, and discuss possible changes later? Then at least we would have a working mt940 and camt import in the official oca modules.

sbidoul:
Okay. Would you then move parserlib to account_bank_statement_import instead? 

So that would leave 3 things to do:
- [x] move parserlib.py to account_bank_statement_import
- [X] renaming
  - account_bank_statement_import_mt940_base
  - account_bank_statement_import_mt940_nl_ing
  - account_bank_statement_import_mt940_nl_rabo
  - account_bank_statement_import_camt
- [x] handling of statement_id

lmignon:
@NL66278 IMHO the parserlib has some advantages but also some restrictions. I'm not again it's inclusion it the account_bank_statement_import addons but contributors must remain free to use it or not in there developments and I would avoid getting the message that new parser should use this layer. The pylint check and the dot notation are valuable advantages but must be balanced with the limitations/issues introduced by this new layer. https://github.com/OCA/bank-statement-import/pull/15#issuecomment-112777035
@nbessi, @alexis-via I'm curious to know if you think that this new layer would ease or will ease the dev of your parsers?

sbidoul:
@NL66278 I implemented the rename in https://github.com/NL66278/bank-statement-import/pull/6

NL66278:
@sbidoul I merged your PR. Thanks a lot. I am off to two weeks holiday now....

sbidoul:
Thanks Ronald. Enjoy the holidays.

:+1:

hbrunn:
@sbidoul @lmignon I count three thumbs up. Ready to merge, right?

sbidoul:
@hbrunn  absolutely.

:clap: Thanks to all involved.

Issue #18, opened 2015-05-19
Active PRs status


lmignon:
A try to summarize the current situation...
- #14 Add the possibility to bulk import files containing bank statements. This PR requires a change to the API. Following a discussion with @qdp-odoo, a PR on upstream has been done including the required changes and the migration to the new API odoo/odoo#5863. The PR on upstream is under review. We have been trying to contact @qdp-odoo to see if the branch can be merged in odoo master, but no luck so far.
-  https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api is the backport of  odoo/odoo#5863 (WIP)
- #13 fix identify partner by res.partner.bank. This PR will not be necessary if the backport is merged since a more generic fix has been provided (see comments https://github.com/OCA/bank-statement-import/pull/13#issuecomment-84653446, ...)
- #4 Add camt and mt940 import, including the changes to support multiple accounts and currencies per file.  ??? superseded by #15 ???
- #15 same as #4 but based on the work done in https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api .

sbidoul:
Ok, things gets a bit clearer thanks to @qdp-odoo's comments on https://github.com/odoo/odoo/pull/5863. I think we have now reached a stage where maintaining a pure backport of master in 8.0 is not feasible anymore so it will be more efficient to maintain a separate version of the import framework for 8.0.

So I suggest @lmignon does a PR with https://github.com/acsone/bank-statement-import/tree/8.0-refactor-new-api and we review that as well as @NL66278 's additions. 

All while keeping an eye on how master evolves to make sure the parsers remain as much as possible forward compatible with master.

lmignon:
@sbidoul PR done https://github.com/OCA/bank-statement-import/pull/19

PR #19, opened 2015-05-21
8.0 refactor new api

lmignon:
The PR is a backport of odoo/odoo#5863.
The PR includes:
- Migration to new API
- Tests
- An orthogonal way to search for account_number whatever its formatting.  (sanitized_acc_number on res.partner.bank)
- A little refactoring of the import_file method to improve the extensibility (required for https://github.com/OCA/bank-statement-import/pull/14)

To ease the review, this PR doesn't include an interesting PR from @NL66278 on  odoo/odoo#5863 https://github.com/acsone/odoo/pull/2. 'Support import for different accounts and currencies.'


yvaucher:
@lmignon According to @qdp-odoo in odoo/odoo#5863 not much was kept from this PR on odoo.

Is this PR still valid?

sbidoul:
@yvaucher I understood that the concepts were kept on master even if the code ends up different.

See also my other comment about this https://github.com/OCA/bank-statement-import/issues/18#issuecomment-104199209.

I was the one who advocated to keep in sync with upstream for this part. I now think it's not realistic anymore. So I believe we should move forward while keeping an eye on master to make sure that at least the parsers remain compatible so migrating to v9 will be as painless as possible.

lmignon:
@yvaucher Yes it's. The description provides the improvements implemented in this PR and @sbidoul  explains in his latest comment why the PR diverges from the upstream.

yvaucher:
Ok then :+1:

account_bank_statement_import/__openerp__.py
```diff
@@ -1,18 +1,16 @@
...
     'name': 'Account Bank Statement Import',
-    'category' : 'Accounting & Finance',
+    'category': 'Accounting & Finance',
     'version': '1.0',
     'author': 'OpenERP SA',
     'depends': ['account'],
     'demo': [],
-    'description' : """Generic Wizard to Import Bank Statements.
-    
+    'description': """Generic Wizard to Import Bank Statements.
```
pedrobaeza: As we are going to diverge from master, please put this on a README.rst with the proper OCA format.

account_bank_statement_import_ofx/__openerp__.py
```diff
@@ -25,9 +25,8 @@
 create periods for the year 2013.
```
pedrobaeza: Extract description

account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
```diff
@@ -14,41 +12,49 @@
...
     from ofxparse import OfxParser as ofxparser
 except ImportError:
-    _logger.error("OFX parser unavailable because the `ofxparse` Python library cannot be found."
-                    "It can be downloaded and installed from `https://pypi.python.org/pypi/ofxparse`.")
+    _logger.warn("ofxparse not found, OFX parsing disabled.")
     ofxparser = None
 
-class account_bank_statement_import(osv.TransientModel):
+
+class account_bank_statement_import(models.TransientModel):
```
pedrobaeza: CamelCase

account_bank_statement_import_qif/__openerp__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     'name': 'Import QIF Bank Statement',
-    'category' : 'Accounting & Finance',
+    'category': 'Accounting & Finance',
     'version': '1.0',
     'author': 'OpenERP SA',
     'description': '''
```
pedrobaeza: Extract description

account_bank_statement_import_qif/account_bank_statement_import_qif.py
```diff
@@ -6,41 +6,49 @@
...
-    def _get_hide_journal_field(self, cr, uid, context=None):
-        return context and 'journal_id' in context or False
+    @api.model
+    def _get_hide_journal_field(self):
+        return self.env.context.get('journal_id') and True
 
-    _defaults = {
-        'hide_journal_field': _get_hide_journal_field,
-    }
+    journal_id = fields.Many2one(
```
pedrobaeza: Please put this field in the main module, because there are other parsers that can use it (spanish one, for example), and it makes no sense to duplicate it when you have all installed.

pedrobaeza:
Why changing the SVG file? I put one with the text converted to curves to avoid to need the font.

lmignon:
@pedrobaeza I'll take into account all your remarks ASAP. Once again Thank you for the review...

lmignon:
@pedrobaeza I've modified the code according to your remarks

NL66278:
:+1:  Time to get this merged.

pedrobaeza:
There's still the field and view for putting journal_id in the main module. For us is totally needed to avoid the dependency on qif file.

lmignon:
@pedrobaeza IMO, this change can be done in an other PR once it'll be required to support a new parser.

pedrobaeza:
It is already required by our spanish bank statement file, that doesn't include bank information in it. It's already merged in https://github.com/OCA/l10n-spain/tree/8.0/l10n_es_account_bank_statement_import_n43.

Please let's make it in one step.

lmignon:
@pedrobaeza done

pedrobaeza:
Thank you very much, you're great! :+1: 

Merging finally!

pedrobaeza:
When goes green..

lmignon:
Thank you @pedrobaeza :smirk:

PR #21, opened 2015-05-28
[ADD] migration script

hbrunn:
and a module that takes care of saving imported files (+migration)

This does a minor migration from account_banking and adds a module that can be used to restore 7.0's behavior of saving the imported file.

I do the PR against current 8.0, but it is meant to be applicable to #14, #15 and #19, I just want to PR this cleanly in order not to have to look through all the noise already to be reviewed in the other PRs


pedrobaeza:
The icon can be the sheet with a paperclip, that it's the usual symbol of attachments

lmignon:
:+1:  (Code review) The top would be to have automated tests. (for the new module)

hbrunn:
@lmignon good idea, here you go

account_bank_statement_import_save_file/models/account_bank_statement_import.py
```diff
@@ -0,0 +1,60 @@
...
+    _inherit = 'account.bank.statement.import'
+
+    @api.model
+    def _import_statement(self, statement):
+        (statement_id, notifications) = \
+            super(AccountBankStatementImport, self)._import_statement(
+                statement)
+        if statement_id:
+            # get raw file data from the stack
+            def get_data_file(frame):
```
nbessi: why the frame lookup don't we recive the file in the signature ?

hbrunn: back then, no. Now, it makes more sense to overwrite https://github.com/OCA/bank-statement-import/blob/8.0/account_bank_statement_import/account_bank_statement_import.py#L71

I'll put this to work in progress until i can follow up

hbrunn:
@nbessi rebased on current upstream with the necessary changes

nbessi:
:+1:

hbrunn:
@pedrobaeza now with a new icon

pedrobaeza:
:+1:

PR #23, opened 2015-06-09
account_bank_statement_import_ofx is not automatically installed anymore

### LINKED ISSUE #22: import ofx should not have auto_install set to True
I do not think  we should automatically install this module by defaut when we install bank statement import addons.

It has an external Python dependency that may prevent other localisation parser to be installed simply because the installation of the ofx parser failed.

nbessi:
Fixes #22


pedrobaeza:
:+1:

lmignon:
:+1:

pedrobaeza:
This can be obviously fast-tracked.

PR #27, opened 2015-06-10
Avoid demo data using an exisiting account number

### LINKED ISSUE #26: impossible to have green test with bank statement import
Hello, 

We have duplicate account number in demo data that break travis run.
https://github.com/OCA/bank-statement-import/blob/8.0/account_bank_statement_import/demo/partner_bank.xml#L7
and
https://github.com/odoo/odoo/blob/8.0/addons/account_payment/account_payment_demo.xml#L6

Also I think the demo data should be moved in their respective importer addons

nbessi:
Fixes #26


pedrobaeza:
:+1:

lmignon:
:+1:

sbidoul:
:+1: thanks!

PR #30, opened 2015-06-23
[FIX] account_bank_statement_import: Handle correctly the selection of the journal

pedrobaeza:
This has happened when moving the journal_id selection from QIF module


antespi:
:+1: 
@pedrobaeza  It works

Endika:
:+1:

steveferry:
:+1:

Issue #24, opened 2015-06-09
[8.0] Constraint unique over 'sanitazed_acc_number'


antespi:
We've just update a customer with last changes, including PR : https://github.com/OCA/bank-statement-import/pull/19
And we've have this issue.

Our customer has two partners with the same acc_number. I understand that this use case is not common (or maybe a mistake by user) in B2B. But in B2C it could happend. For example, my wife and me have the same acc_number (we are co-titulars), and want to make different registration in the same Odoo e-commerce. Both partners will have the same acc_number.

I suggest to create a new contraint (partner_id, sanitazed_acc_number), but maybe this is not the point of this restriction.

What do you think?

lmignon:
@antespi  IMO it's not safe to share an account number between different partners. In your case, your wife and you should be part of the same organisation and use the address of this organisation.
Moreover, when importing a bank.statement.line we expect to find only one partner linked to the account number.

My 2 cents,

lmi

antespi:
@lmignon bank statement import process is a good reason for this constraint, we will tell this argument to our customer to justify this change.

Thanks for your answer

PR #31, opened 2015-07-03
8.0 no uniqueness in sanitized account number

hbrunn:
Enforcing uniqueness on bank accounts is a bad idea. Think of partners sharing a bank account or institutions using the same bank account for different entities


pedrobaeza:
This comes from this feature that doesn't work correctly if this uniqueness is not set: https://github.com/OCA/bank-statement-import/pull/13.

This concern was also raised previously in https://github.com/OCA/bank-statement-import/issues/24, and you can check the answers.

hbrunn:
@sbidoul @lmignon @pedrobaeza I read the issues, but I think the technical argument about linking the wrong partner is weak. If that happens, we have to give the user doing reconciliation the possibility to change the partner afterwards, not enforce uniqueness.

My best real-world example would be me and my partner registering on an odoo-run webshop, both of us filling in our shared account. Do we really want the second one getting an error message with the question to register as a company first and then add us as employees? How about me forgetting my password and simply creating a new account? Then it breaks when I fill in my account info again. This is really not practical.

We also don't enforce unique email addresses for very similar reasons, even though this would make matching incoming email a lot less error prone.

pedrobaeza:
I don't have a strong opinion on this, so let's see what others say. @antespi, you can also tell your opinion about this.

lmignon:
@hbrunn I'm in favor to keep unique account number by default. IMO it's safer. Account number is not only used to get the res.partner but also to get the account.journal.  Is-it consistent to be able to create 2 bank.account with the same acc_number but a different journal_id?  A compromise could be to create a new addons 'uniq_acc_number' defining the constrains with depends=['account_bank_statement_import'] and with autoinstall='true' (thus installed by default). If the uniqueness on bank accounts is a problem, you can consciously uninstall the module. 
My 2 cents,
lmi

lmignon:
@hbrunn  as @pedrobaeza  says, let's see what others say.

sbidoul:
@lmignon I see your point. I also see the practical issues raised by @hbrunn.

Perhaps the default constraint should prevent duplicates only for bank accounts linked to res.company?

lmignon:
@sbidoul that make sense. :+1:

antespi:
I'm agreed with @sbidoul, an extra addon could be a good option. But if this addon is auto-installable, admin uninstall it and after that he updates all, the addon will auto-install again?

lmignon:
@antespi You're probably right. Auto-install can be set to false. The new addon could let the choice to the end user to enforce the uniqueness of acc_number.

antespi:
@lmignon Via a wizard after install or/and via checkbox in Setting > Configuration > Invoicing, isn't it?

hbrunn:
I'll be happy to provide this module, give me some time to add it to this PR

hbrunn:
@lmignon @antespi @sbidoul now we have the constraint in its own addon with a config setting in the accounting wizard

lmignon:
@hbrunn Thank you for your hard work! :+1: (once CI is green)

pedrobaeza:
:+1: I proceed with the merge and we decide later the convenience of sanitized_acc_number field.

PR #32, opened 2015-07-13
[FIX] pass parameters we got to super

hbrunn:


pedrobaeza:
:+1: This should be fast-tracked because it's very severe. Do you know if this also happens on upstream v9 version?

hbrunn:
@pedrobaeza thanks! It was introduced by #19, so it's our bug

PR #35, opened 2015-08-15
Update Translations Templates

ploegvde:


pedrobaeza:
This is not needed anymore, as Transifex is activated. You can check for translations here: https://www.transifex.com/oca/OCA-bank-statement-import-8-0/dashboard/

ploegvde:
But things are missing, so i thought updating the pot, will update the base templates on transifex. When the are up to date, i can translate to e.g. Dutch

pedrobaeza:
No, I'm afraid not. Each new commit, translations are re-exported from Odoo itself and posted to Transifex, so if there's something missing, we should see if there's a problem on the export job or on the modules themselves.

ploegvde:
Ok, and see something? Because there are things missing as you can see in my PR.

pedrobaeza:
No, but I think you're not understanding the functionality of the Transifex script. I'll try to explain:
- Every commit, the script runs the server and export the pot file for the involved modules.
- Already present pot files are ignored and they're not updated.
- New exported pot files are uploaded to Transifex.

So you have to check if in Transifex all terms are present or not. That's the real test.

PR #36, opened 2015-08-27
8.0 import parsers complete therp

NL66278:
Add automatic handling of zip import files.

The code to realise this IMHO does not warrant a separate module. I think it is preferable to support zip files out of the box, without the need to install a separate module.


account_bank_statement_import/models/account_bank_statement_import.py
```diff
@@ -132,7 +160,7 @@ def _import_statement(self, stmt_vals):
         return self._create_bank_statement(stmt_vals)
 
     @api.model
-    def _parse_file(self, data_file):
+    def _parse_file(self, dummy_data_file):
```
hbrunn: don't rename parameters if not absolutely necessary

NL66278: @hbrunn I added the dummy_ prefix because the parameter is not used in this skeleton function. And then pylint complains about unused parameters. Can we leave it? Otherwise I will rename it - at the cost of pylint complaining.

hbrunn: @NL66278 please rename and add `# pylint: disable=W0613` (or W0612?) to silence pylint about the parameter

hbrunn:
@NL66278 please rebase

StefanRijnhart:
:+1:

account_bank_statement_import/__init__.py
```diff
@@ -1 +1,2 @@
+# -*- encoding: utf-8 -*-
```
pedrobaeza: s/encoding/coding

account_bank_statement_import/models/account_bank_statement_import.py
```diff
@@ -2,16 +2,19 @@
...
-from openerp.exceptions import Warning
+from openerp.exceptions import Warning as UserError
 
-_logger = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
 class AccountBankStatementLine(models.Model):
     """Extend model account.bank.statement.line."""
+    # pylint: disable=too-many-public-methods
```
pedrobaeza, line 17: This is only a beta message so it's not a real error.

account_bank_statement_import_camt/__openerp__.py
```diff
@@ -19,7 +19,7 @@
 ##############################################################################
 {
     'name': 'CAMT Format Bank Statements Import',
-    'version': '8.0.0.3.0',
+    'version': '8.0.1.0.4',
```
pedrobaeza: Version should be 8.0.0.4.0

account_bank_statement_import_mt940_nl_ing/__openerp__.py
```diff
@@ -19,7 +19,7 @@
 ##############################################################################
 {
     'name': 'MT940 IBAN ING Format Bank Statements Import',
-    'version': '8.0.0.3.0',
+    'version': '8.0.1.0.4',
```
pedrobaeza: 8.0.0.4.0

account_bank_statement_import/models/account_bank_statement_import.py
NL66278, line 17: I like my pylint output to be as clean as possible. On the other hand this would be needed in every class based on models.Model, as the root problem lies in the Odoo code itself, which puts "everything but the kitchen sink" into one class.

I will remove it.

hbrunn:
:+1:

PR #38, opened 2015-10-04
german translations for bank-statement-import

rschnapka:


StefanRijnhart:
:+1: but I better check with @pedrobaeza if translation files like this can still be added and not conflict with the Transifex integration. Next time you will probably want to add translations through Transifex.

pedrobaeza:
No, there's no problem. Translation will be imported as initial suggestions. You'll need to enter Transifex to validate them though.

:+1: technically

StefanRijnhart:
Thanks @pedrobaeza for clarifying!

PR #41, opened 2015-11-03
8.0 FIX 2 variables had the same name (account_number) !!!

alexis-via:
Probably one of the most stupid bug I even found in Odoo: someone used the same variable name (account_number) for 2 different things: in the method() _complete_statement() in account_bank_statement_import/account_bank_statement_import.py:
- the first 'account_number' variable is the last argument of the method and contains the bank account number of the imported file. In this method, it is used as a prefix in the 'unique_import_id' field, so as to have a stronger 'unique_import_id' that contain the bank account number.
- the second 'account_number' variable is used in the loop on the bank statement lines and gets the account number of the partner related to this bank statement line.

So the second variable 'account_number' over-write the first one after the first pass in the loop !!!

In the scenario of my customer that found this bug, I was importing an OFX file. The OFX files don't contain any partner account_number on bank statement lines (at least for French banks). So, when I had a look at the unique_import_id of the bank statement lines, I had:
00143525710-201501900006BD27
201501900005BD27
201501900004BD27
201501900003BD27
201501900002BD27
201501900001BD27

The first bank statement line had the bank account number in the unique_import_id, but not the next ones, because of the over-write of the variable account_number. The problem is that the unique_import_id of the other lines (201501900004BD27) were already used by other bank statement lines from other accounts in other companies (this is a multi-company setup)... so the resulting bank statement was almost empty, because most lines had been "dropped" by the import thinking they were duplicates.

This is an important regressing ; it seems that the bug was introduced by @NL66278 in this commit https://github.com/OCA/bank-statement-import/commit/182ea2893a303c2d306d3abb9588464aabd33d5e

### DIFF
--- a/account_bank_statement_import/models/account_bank_statement_import.py
+++ b/account_bank_statement_import/models/account_bank_statement_import.py
@@ -299,16 +299,16 @@ class AccountBankStatementImport(models.TransientModel):
                 # closed.
                 partner_id = False
                 bank_account_id = False
-                account_number = line_vals.get('account_number')
-                if account_number:
+                partner_account_number = line_vals.get('account_number')
+                if partner_account_number:
                     bank_model = self.env['res.partner.bank']
                     banks = bank_model.search(
-                        [('acc_number', '=', account_number)], limit=1)
+                        [('acc_number', '=', partner_account_number)], limit=1)
                     if banks:
                         bank_account_id = banks[0].id
                         partner_id = banks[0].partner_id.id
                     else:
-                        bank_obj = self._create_bank_account(account_number)
+                        bank_obj = self._create_bank_account(partner_account_number)
                         bank_account_id = bank_obj and bank_obj.id or False
                 line_vals['partner_id'] = partner_id
                 line_vals['bank_account_id'] = bank_account_id


pedrobaeza:
Please check lynt errors. Have you checked if this bug is also in Odoo v9?

alexis-via:
@pedrobaeza The bug is NOT present in odoo v9, see https://github.com/odoo/odoo/blob/9.0/addons/account_bank_statement_import/account_bank_statement_import.py#L179
They still use "identifying_string" as the 2nd variable name. This bug is a pure OCA bug, introduced by Ronald on June 22nd 2015.

flake8/travis errors are not related to my bug fix, so they should be fixed in another PR, so avoid polluting my bug fix PR with changes not related to the bug.

pedrobaeza:
You can fix PEP8 in a separate commit to not let the branch red, and let the fix isolated in the current one.

rvalyi:
:+1: nice shot

StefanRijnhart:
Thank you @alexis-via for the fix and the thorough analysis. I understand this bug causes terrible problems which you have been hit by, and I would be upset too. However, I do find the name of the branch offensive (and that will stay in git history forever), and I am hoping that you could make the bug report a little less judgmental too. Could you push under a factual branch name please?

lmignon:
Thank you for the fix. A test reproducing the scenario of this bug would be a plus and would avoid future regression. @StefanRijnhart says, a factual branch name is IMO required to merge this fix. :+1: once the branch is renamed.

pedrobaeza:
Oh, I didn't see the branch name. Alexis, you should be less enthusiastic with your emotions when discovering others' bugs. Everyone can make a mistake, but that's why we can do PRs to amend it! @StefanRijnhart, if I change the merge commit message, the name still remains?

alexis-via:
OK I'll change the branch name. I guess I have to make another PR...

NL66278:
Thanx @StefanRijnhart  and @pedrobaeza for taking up the branch name and the way the bug is described. Yes, it might be a stupid bug - but after IMHO giving a large contribution to the development of the 8.0 banking import, I think it is not fair and demotivating to use this kind of strong language. I have been bitten in the past also by stupid bugs, both my own and others, but let us not forget that we must forge a community to improve the work we create.

pedrobaeza:
Superseeded by https://github.com/OCA/bank-statement-import/pull/42

alexis-via:
Sorry for the bad branch name. I wanted to go to bed early yesterday night to compensate for some long bug-fixing in Odoo for the same customer during my previous nights... and then I received this new email yesterday with a new bug report and had to spend part of my evening fixing it... and I was very upset when I found the cause of the bug, and the fact that it was introduced several months ago and nobody found it before me. And I went to bed late again because of this bug. When I choose the branch name, I thought that Odoo S.A. caused the bug ; I just found out that it was introduced by one of us when writing the last line of my bug report.

Please accept my apoligies @NL66278

NL66278:
@alexis-via  Thanx for your message. And sorry for the lost hours that this bug caused for you. So apologies accepted. Have a nice day today and kind regards,

StefanRijnhart:
Thanks for understanding!

PR #44, opened 2015-11-18
Port account_bank_statement_import_ofx for 9.0

ilyasProgrammer:
Originally module does not work on odoo 9.0
In order to make it work on 9.0 i removed modified module "account bank_statement_import" because ofx import won`t work with it. It is supposes to use same original module. Besides field "owner_name" now is gone. Its calls now "bank_name", as i assume.


zuher83:
@ilyasProgrammer 
:thumbsup:

gurneyalex:
:+1:

pedrobaeza:
Superseeded by #70

PR #46, opened 2015-11-22
N1

ilyasProgrammer:


account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
```diff
@@ -9,11 +9,11 @@
 
 _logger = logging.getLogger(__name__)
 
-try:
-    from ofxparse import OfxParser as ofxparser
-except ImportError:
-    _logger.warn("ofxparse not found, OFX parsing disabled.")
-    ofxparser = None
+# try:
```
pedrobaeza, line 12: Why are you commenting this?

account_bank_statement_import/models/account_bank_statement_import.py
```diff
@@ -1,366 +0,0 @@
...
-                self._import_statement(stmt_vals))
-            if statement_id:
-                statement_ids.append(statement_id)
-            notifications.append(new_notifications)
-        if len(statement_ids) == 0:
-            raise Warning(_('You have already imported that file.'))
-        return statement_ids, notifications
-
-    @api.model
-    def _import_statement(self, stmt_vals):
```
lmignon, line 103: We must preserve an override of the base model since we have a more defensive way to do the import.
- The creation of bank_account_journal is disallowed
- The creation of bank accounts for partners is disallowed.

PR #47, opened 2015-11-23
[FIX] have one list of notifications

hbrunn:
With the current code, you end up with a list of lists, I don't think that's what we want


pedrobaeza:
Does this happen also in upstream v9?

For this, :+1:

hbrunn:
@pedrobaeza no, this happens in the indirection `_import_file` which is specific to our code

pedrobaeza:
OK, thanks for the info.

sbidoul:
:+1: (code review, no test)

lmignon:
:+1:  (code review, no test)

PR #53, opened 2016-01-07
rename account_bank_statement_import

gurneyalex:
that module clashes with an official addon added in v9


gurneyalex:
I was investigating some CI build failures, and traced them down to the
module account_bank_statement_import which used to be a module in
OCA/bank-statement-import is now a core module. I have not traced the
specifics for this (i.e. whether both modules have some common ancestry
or not), but the fact is the presence of the module twice in the addons
path (even if one is not installable) seems to be troubling runbot and
travis

https://runbot.odoo-community.org/runbot/build/3135288
https://travis-ci.org/OCA/bank-statement-import/jobs/91784531

Since there are some ongoing ports of modules depending on
account_bank_statement_import in the OCA/bank-statement-import, I expect
these ports are working with the upstream module.

To solve the situation, I think the best way is to remove
account_bank_statement_import altogether from the
OCA/bank-statement-import 9.0 branch (which is what this PR does). 

I can proceed with this if we all agree on that.

gurneyalex:
the fact that some tests have been green is probably related to the move from **unported** to the root of the project repository of the addon.

sbidoul:
Having conflicting module names is bad indeed.

There is some feature parity check to be done however. There are a few features that have been added in the OCA version for which I'm not sure they are in upstream, namely (non exhaustive):
- the import of zip file
- the import of statements corresponding to several bank accounts (?)
- preventing the automatic creation of bank journals

cc/ @NL66278 @lmignon

lmignon:
:+1: with @sbidoul. @gurneyalex We should rename `account_bank_statement_import` in `account_bank_statement_import_base` and inherit from upstream to be sure to preserve all these improvements and our more `defensive approach`.

gurneyalex:
Thanks for the rapid feedback. 

@sbidoul OK I updated the PR do just do the rename. I don't really have time to do the proper refactoring / port to v9 of that addon (my main goal here is getting green status for repositories / PRs having dependencies on account_bank_statement_base).

sbidoul:
@gurneyalex could you add a note in the README saying that this module must be compared with the upstream version before being dropped?

lepistone:
The module is unported, so I think we can merge this to fix the CI, and whoever will port it will take care of figuring out its fate.

So :+1: thanks!

gurneyalex:
@sbidoul done

petrus-v:
@gurneyalex,

Thanks for background tasks! and the warning!

:+1:

sbidoul:
@gurneyalex thanks! 

:+1: 

Fast tracking.

NL66278:
:+1:  Agree with renaming now and doing a proper port later.

gurneyalex:
thanks!

PR #54, opened 2016-01-07
8.0 Add support for OFX files without <NAME> inside all <STMTTRN>

alexis-via:
Memo (<NAME>) and payee (<PAYEE>) are not required fields in OFX statement, cf section 11.4.3 Statement Transaction <STMTTRN> of the OFX specs: the required elements are in bold, cf 1.5 Conventions and these 2 fields are not in bold. But the 'name' field of account.bank.statement.line is
required=True (cf https://github.com/odoo/odoo/blob/8.0/addons/account/account_bank_statement.py#L930), so I write '-' if it's empty.

### DIFF
--- a/account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
+++ b/account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
@@ -45,12 +45,13 @@ class AccountBankStatementImport(models.TransientModel):
                 # (normal behavious is to provide 'account_number', which the
                 # generic module uses to find partner/bank)
                 bank_account_id = partner_id = False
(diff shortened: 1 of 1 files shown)


SMLGP:
:+1: I confirm it fixes the issue with OFX file from LCL

alexis-via:
It appears that this issue only happens for checks debited from the bank account... and there is another tag in the OFX file that has the check number. So I just improved the PR to get the transaction type (required in OFX) and the check number (if we have it), instead of "-".

MOSAMAD:
:+1:

hbrunn:
:+1:

PR #58, opened 2016-02-08
[IMP] if there's a date but no period, ger the date's period

hbrunn:
First, I though that was a deficiency in the mt940 parsers, but now I figure we can do this centrally. Of should I move it to the parser where I need it?

If we don't set a `period_id`, we'll get the current date's period, which is most probably wrong.

### DIFF
--- a/account_bank_statement_import/models/account_bank_statement_import.py
+++ b/account_bank_statement_import/models/account_bank_statement_import.py
@@ -354,7 +354,9 @@ class AccountBankStatementImport(models.TransientModel):
             # do this now
             try:
                 stmt_vals['period_id'] =\
-                    self.env['account.period'].find(dt=stmt_vals['date']).id
+                    self.env['account.period']\
+                        .with_context(account_period_prefer_normal=True)\
+                        .find(dt=stmt_vals['date']).id
             except RedirectWarning:
                 # if there's no period for the date, ignore resulting exception
                 pass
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

account_bank_statement_import/models/account_bank_statement_import.py
```diff
@@ -349,6 +349,15 @@ def _complete_statement(self, stmt_vals, journal_id, account_number):
                         bank_account_id = bank_obj and bank_obj.id or False
                 line_vals['partner_id'] = partner_id
                 line_vals['bank_account_id'] = bank_account_id
+        if 'date' in stmt_vals and 'period_id' not in stmt_vals:
+            # if the parser found a date but didn't set a period for this date,
+            # do this now
+            try:
+                stmt_vals['period_id'] =\
+                    self.env['account.period'].find(dt=stmt_vals['date']).id
```
sbidoul: According to https://github.com/odoo/odoo/blob/8.0/addons/account/account.py#L1042, you should put account_period_prefer_normal=True in the context to avoid obtaining an opening period.

hbrunn: @sbidoul thanks, done

sbidoul:
:+1:

elicoidal:
:+1:

PR #62, opened 2016-03-09
OFX: stronger unique_import_id

alexis-via:
Stronger unique_import_id (to lower the risk of having 2 identical unique_import_id in the same file.. which is possible with the current code, according to my experience)

Add bank account number in bank statement name, instead of only the bank number (important when you have several bank accounts at the same bank)

### DIFF
--- a/account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
+++ b/account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
@@ -58,7 +58,8 @@ class AccountBankStatementImport(models.TransientModel):
                         transaction.memo and ': ' + transaction.memo or ''),
                     'ref': transaction.id,
                     'amount': transaction.amount,
-                    'unique_import_id': transaction.id,
+                    'unique_import_id': '%s-%s-%s' % (
+                        transaction.id, transaction.payee, transaction.memo),
                     'bank_account_id': bank_account_id,
                     'partner_id': partner_id,
                 }
@@ -83,7 +84,7 @@ class AccountBankStatementImport(models.TransientModel):
             ))
 
         vals_bank_statement = {
-            'name': ofx.account.routing_number,
+            'name': ofx.account.number,
             'transactions': transactions,
             'balance_start': ofx.account.statement.balance,
             'balance_end_real':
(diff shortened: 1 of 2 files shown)


SMLGP:
+1: using it in production and I confirm it solves the problem".

pedrobaeza:
I see the first change correct, but the second one makes the move/statement number very long, so I would remove that part.

StefanRijnhart:
@alexis-via would it work if you followed @pedrobaeza's suggestion and leave out the memo?

alexis-via:
@StefanRijnhart @pedrobaeza 
The statement name is now shorter: it only contains the account number and not the bank number. This change is important for companies that have several accounts at the same bank (I have several such customers ; for example, they have an account in EUR and an account in USD at the same bank).

StefanRijnhart:
Sorry I misinterpreted Pedro's comment.

Is it now the case that the bank statement names of all statements of the same account are identical?

alexis-via:
Currently, the bank statement names of all statements from the same bank are identical.

pedrobaeza:
Thanks for the changes :+1:

StefanRijnhart:
If it is not a problem that all statements for the same bank have the same name, :+1:

pedrobaeza:
Ehm, I read yesterday very fast last comments, so I regret my approval for now. Why don't we use the bank statement sequence for the statement number and forget about other combinations (routing number, account number...)?

alexis-via:
@pedrobaeza I don't understand your point ; there is no sequence for account.bank.statement object.

I am very surprised that you give so much importance to the change I propose for the name of the bank statement. I only proposed to add the account number to the name because it helps make a distinction when you have several bank accounts at the same bank (the journal_id field should make the distinction, but it helps to have this distinction in the name).

In the module that import the France-specific format for bank statements, I use a more elaborate name, cf https://github.com/OCA/l10n-france/blob/8.0/account_bank_statement_import_fr_cfonb/account_bank_statement_import_fr_cfonb.py#L166 but it requires a bit more code. If you prefer this more elaborate approch the include start/end dates, I will update my PR.

StefanRijnhart:
@alexis-via looks good. @pedrobaeza do you agree to have statement names like 

```
        'name': _('Account %s  %s > %s') % (
            account_number, start_date_str, end_date_str),
```

alexis-via:
I propose to merge this PR... if the bank statement names are really a problem, we will still be able to change it later.

pedrobaeza:
OK, sorry for the delay. It hanged deep in my inbox.

Merging.

PR #63, opened 2016-03-10
[9.0] add german translations, if present then from *.pot

rschnapka:
havn't seen a Transifex-repo, so I chose to add in source.


melroy89 (changes requested):
NOPE! Do NOT merge. There is a Transifex repo:
https://www.transifex.com/oca/OCA-bank-statement-import-9-0/dashboard/

For version 8,9,10 and 11.

melroy89:
Reject PR please.

PR #64, opened 2016-04-07
[FIX] partner_id set to null if not bank_account on statement importation

omar7r:
Hi,

If not found bank account on statement vals, partner_id always was being null. I don't know any other statements importation file, that "n43" the typical at Spain, "l10n_es_account_bank_statement_import_n43" addon in Spain's repository. This format never brings the bank account but the vat or the name of the customer, usually yes, without this fix, we are unable to import statements, proposing partners, because always set null in partner_id field if not bank account.

Regards.

### DIFF
--- a/account_bank_statement_import/models/account_bank_statement_import.py
+++ b/account_bank_statement_import/models/account_bank_statement_import.py
@@ -351,8 +351,10 @@ class AccountBankStatementImport(models.TransientModel):
                         bank_obj = self._create_bank_account(
                             partner_account_number)
                         bank_account_id = bank_obj and bank_obj.id or False
-                line_vals['partner_id'] = partner_id
-                line_vals['bank_account_id'] = bank_account_id
+                if partner_id:
+                    line_vals['partner_id'] = partner_id
+                if bank_account_id:
+                    line_vals['bank_account_id'] = bank_account_id
         if 'date' in stmt_vals and 'period_id' not in stmt_vals:
             # if the parser found a date but didn't set a period for this date,
             # do this now


pedrobaeza:
On our module, we try to deduce later the partner_id, so I don't get the purpose of this module.

omar7r:
Sure? In  our implementation  is evaluated later. Spanish module load the
initial vals this is checked later, when vals are created.

pedrobaeza:
Maybe we have to think then on modifying the Spanish one, because in v9 this module is in Odoo core, and we cannot modify it.

omar7r:
Hi @pedrobaeza 

In general this is a code error, in any case where bank account is not set, partner_id is null. I think this is not correct.

Regards.

pedrobaeza:
As said, we can modify this, but in v9 we don't have this possibility as this is part of the core, so let's refactor our localization this way:
- Remove the call to `_get_partner` in the parse method.
- Overwrite `_complete_statement`.
- Call super.
- Then iterate over the lines and call `_get_partner` here.
- We can check if there's already a partner assigned to inhibit the N43 heuristic search.

So for this :-1:

omar7r:
Hi @pedrobaeza 

My PR is for 8.0, it is needed for 8.0. When the module will be migrated to 9.0 there will be to refactor it, because the tools in 9.0 are not the tools in 8.0.
I don't undestand your point...

pedrobaeza:
If you do it for 8.0, then you don't have to change it for 9.0, and as the research is needed to be done for 9.0 in any case, please do it directly on 8.0. As said, it's to change in our module the point where the partner is assigned, and put it after this assignation.

StefanRijnhart:
:+1:

pedrobaeza:
@omar7r, please check the right solution for handling with the standard module for v8 and v9: https://github.com/OCA/l10n-spain/pull/365.

I close this PR as the purpose is the same.

PR #65, opened 2016-05-09
[FIX] CAMT Statement id may not be unique over time

StefanRijnhart:
After happily importing a year's worth of Rabobank (NL) statements, we found that some statements would mysteriously be skipped when importing more statements afterwards. As it turns out, Rabobank statement identifiers are reused every year. This would trigger Odoo's rigorous duplicate dropping. This change adds the statement's date to the identifier, if it is not in there already.

### DIFF
--- a/account_bank_statement_import_camt/camt.py
+++ b/account_bank_statement_import_camt/camt.py
@@ -202,6 +202,11 @@ class CamtParser(object):
         if statement['transactions']:
             statement.date = datetime.strptime(
                 statement['transactions'][0].execution_date, "%Y-%m-%d")
+            # Prepend first statement's date to improve id uniqueness factor
+            date = statement.date.strftime('%Y%m%d')
+            if date not in statement.statement_id:
+                statement.statement_id = "%s-%s" % (
+                    date, statement.statement_id)
         return statement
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
- Bump version number.
- Maybe a migration script for being consistent in the identifier?

StefanRijnhart:
Not sure migration is possible. Is it easy to detect which bank statements have been imported with this parser?

StefanRijnhart:
Version bumped.

pedrobaeza:
There's no hard link, but maybe by journal.

StefanRijnhart:
Better not then.

pedrobaeza:
OK :+1:

StefanRijnhart:
Need to fix tests

luc-demeyer:
@StefanRijnhart 

PR https://github.com/OCA/bank-statement-import/pull/67 is another way to address this issue (and other issues since I made this PR to cope with 'no-clean' camt statements of a german bank.

NL66278:
This PR might be superseded by this one https://github.com/OCA/bank-statement-import/pull/83, where I took Stephan' s changes, but also adapted the tests.

PR #66, opened 2016-05-12
We added the bank_account_id as a condition for the sql_constraints b…

yk2kus:
…ecause we can have duplication of transaction ID in different bank accounts.


pedrobaeza:
You shouldn't resolve the duplication at this level, but it's the parser that must generate an unique-ID

PR #69, opened 2016-06-16
Add module for imports of Romanian BRD bank in MT940.

feketemihai:


feketemihai:
@OCA/banking-maintainers Can i have a review on this, please...

pedrobaeza:
I think this should be in l10n-romain, like https://github.com/OCA/l10n-spain/tree/8.0/l10n_es_account_bank_statement_import_n43

feketemihai:
@pedrobaeza Ok...i will create a pr in OCA/l10n-romania, i saw the import about Dutch banks here and that's why i did it here...Thanks.

PR #71, opened 2016-07-26
[UPD] parse more fields as of

hbrunn:
https://www.ing.nl/media/ING_ming_mt940s_24_juli_tcm162-46356.pdf
p 8ff

### DIFF
--- a/account_bank_statement_import_mt940_base/mt940.py
+++ b/account_bank_statement_import_mt940_base/mt940.py
@@ -83,14 +83,28 @@ def handle_common_subfields(transaction, subfields):
     for counterpart_field in ['CNTP', 'BENM', 'ORDP']:
         if counterpart_field in subfields:
             get_counterpart(transaction, subfields[counterpart_field])
+    if not transaction.message:
+        transaction.message = ''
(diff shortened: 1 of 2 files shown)


NL66278:
:+1: LGTM In general. But I have seen while testing that putting EREF also in the message field will lead to duplicate info.

account_bank_statement_import_mt940_base/mt940.py
```diff
@@ -83,14 +83,28 @@ def handle_common_subfields(transaction, subfields):
...
+            # /REMI/USTD//<remittance info>/
+            # or
+            # /REMI/STRD/CUR/<betalingskenmerk>/
+            if len(subfields['REMI']) >= 3 and subfields['REMI'][0] in [
+                'STRD', 'USTD'
+            ]
+            else
+            '/'.join(x for x in subfields['REMI'] if x)
+        )
+    # EREF: End-to-end reference
```
NL66278, line 102: I think EREF should not be put into message (which will get into name), because in many cases it will lead to duplicate information. Eref will already end up in the ref field of the bank statement line.

StefanRijnhart (approved):

PR #72, opened 2016-08-02
[IMP] use all ustrd nodes

hbrunn:
`Ustrd` can occur unlimited times, so we should put all of them into the message: https://www.iso20022.org/documents/messages/camt/schemas/camt.053.001.06.zip (line 1538 of the XSD)

### DIFF
--- a/account_bank_statement_import_camt/camt.py
+++ b/account_bank_statement_import_camt/camt.py
@@ -69,7 +69,7 @@ class CamtParser(object):
                 './ns:RmtInf/ns:Ustrd',
                 './ns:AddtlTxInf',
                 './ns:AddtlNtryInf',
-            ], transaction, 'message')
+            ], transaction, 'message', join_str='\n')
         # eref
         self.add_value_from_node(
             ns, node, [


NL66278:
:+1:  LGTM

pedrobaeza:
Check Travis error

NL66278:
@pedrobaeza I did check the Travis error. It is caused by tests on ofx module, and occurs also on other PR's on this repo. I think this will need a separate issue.

pedrobaeza:
OK, let's merge then this one, although I think this module should be in a localization repository.

NL66278:
@pedrobaeza Thanks for merging. But just wondered why you think this module should be in a localisation repo?

pedrobaeza:
CAMT format is only used in Netherlands AFAIK. Or is it used in another country?

NL66278:
CAMT is an EU standard and is used in multiple countries. Apart from that I think it was agreed that all bank statement import modules would be in this repo, because a company in one country might have a bank account in another. So even our specific modules for the Dutch ING (not this one) are in this repo, because a French company might haven an account with de ING.

pedrobaeza:
OK for the EU part. 

About the agreement that all goes here, that's not true. Each localization put their specific stuff in their repo. See https://github.com/OCA/l10n-spain/tree/8.0/l10n_es_account_bank_statement_import_n43. If you want that for a multi-country/multi-bank company, you download the corresponding localization repository.

Issue #52, opened 2015-12-15
8.0 crash when migrating


alexis-via:
Here is a crash when migrating to 8.0:

```
2015-12-15 17:33:28,437 11044 INFO o7_uni1 openerp.modules.migration: module account_bank_statement_import: Running migration [8.0.1.0>] post-migrate
2015-12-15 17:33:28,438 11044 INFO o7_uni1 openerp.modules.migration: module account_bank_statement_import: Running migration [8.0.1.0>] post-migrate
2015-12-15 17:33:28,439 11044 ERROR o7_uni1 openerp.sql_db: Programming error: column "trans" does not exist
LINE 4: ...         first_value(id) over (partition by trans) id, trans
                                                                  ^
, in query update account_bank_statement_line l set unique_import_id=l1.trans
        from (
            select distinct
            first_value(id) over (partition by trans) id, trans
            from account_bank_statement_line
        ) l1
        where l.id=l1.id
2015-12-15 17:33:28,439 11044 ERROR o7_uni1 openerp.modules.loading: Error executing post migration script for module account_bank_statement_import
`-> account_statement_completion_label_simple
: column "trans" does not exist
LINE 4: ...         first_value(id) over (partition by trans) id, trans
                                                                  ^
```

It is caused by this code, by @hbrunn : https://github.com/OCA/bank-statement-import/blob/8.0/account_bank_statement_import/migrations/8.0.1.0/post-migrate.py#L27

The comment above says "if we end up here, we migrate from 7.0's account_banking set transaction ids, taking care to enforce uniqueness". In my case, I didn't have account_banking on v7. This code should be "protected" to make sure that it is executed only when "trans" exists on account_bank_statement_line.

hbrunn:
did you rename some other module to account_bank_statement_import? This code indeed assumes that upgrades take place via OpenUpgrade, and there the only way to run into this migration script is having account_banking installed in 7, which is why it's safe in this context. How do you do you upgrade?

alexis-via:
I am doing the migration from 7 to 8 via openupgrade. On my v7 database, account_banking was not installed ; only account_banking_payment_export + account_banking_sepa_credit_transfer + account_banking_sepa_direct_debit (and their dependencies) are installed. I must say that I still don't understand why account_bank_statement_import is installed automatically during the upgrade (I was planning to install it later on)... I have to investigate this. But, for the moment, account_bank_statement_import is installed automatically and I get this crash.

hbrunn:
[...]

PR #73, opened 2016-09-21
migrate being called when version is None

pankk:
Tested the following issue:
https://github.com/OCA/bank-statement-import/issues/52
I made sure account_banking was not installed, and did not appear in ir_module_module of the source database. I ran it through a debugger and found that version was None. Compared this script with other scripts and recalled that it is indeed common to return if version is None.

### DIFF
--- a/account_bank_statement_import/migrations/8.0.1.0/post-migrate.py
+++ b/account_bank_statement_import/migrations/8.0.1.0/post-migrate.py
@@ -21,6 +21,8 @@
 
 
 def migrate(cr, version):
+    if not version:
+        return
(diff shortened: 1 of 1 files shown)


pankk:
Don't forget to also change this in 9.0 if this is accepted! :)

pedrobaeza:
:+1: This shouldn't be in 9.0 version (the entire folder).

moylop260:
👍

alexis-via:
Thanks :+1:

pedrobaeza:
Travis error unrelated, so I merge.

PR #77, opened 2016-10-21
[8.0][ENH] Provide alternative way to check for unique import lines.

NL66278:
Currently the uniqueness of import lines is checked by using the physical entry number in the import file (at least for camt files) combined with an identifier for the statement. This does not work in all cases, especially not for customers whre the same bank might be imported several times a day, with overlapping import lines.

This merge provides an optional alternative way to generate a unique id, based on the data in the imported entry. The default will be to not change anything. But for each bank the user will be able to specify the alternative way.

### DIFF
--- a/account_bank_statement_import/models/account_bank_statement_import.py
+++ b/account_bank_statement_import/models/account_bank_statement_import.py
@@ -4,10 +4,12 @@ import logging
 import base64
 from StringIO import StringIO
 from zipfile import ZipFile, BadZipfile  # BadZipFile in Python >= 3.2
+import hashlib
 
 from openerp import api, models, fields
 from openerp.tools.translate import _
 from openerp.exceptions import Warning as UserError, RedirectWarning
+
 
 _logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -234,6 +236,16 @@ class AccountBankStatementImport(models.TransientModel):
         return self.env.user.company_id.currency_id.id
 
     @api.model
+    def _get_bank(self, account_number):
+        """Get res.partner.bank."""
+        bank_model = self.env['res.partner.bank']
+        if account_number and len(account_number) > 4:
+            return bank_model.search(
+                [('acc_number', '=', account_number)], limit=1
+            )
+        return bank_model.browse([])  # Empty recordset
+
+    @api.model
     def _find_bank_account_id(self, account_number):
         """ Get res.partner.bank ID """
         bank_account_id = None
@@ -337,12 +349,19 @@ class AccountBankStatementImport(models.TransientModel):
     def _complete_statement(self, stmt_vals, journal_id, account_number):
         """Complete statement from information passed."""
         stmt_vals['journal_id'] = journal_id
+        statement_bank = self._get_bank(account_number)
         for line_vals in stmt_vals['transactions']:
-            unique_import_id = line_vals.get('unique_import_id', False)
+            unique_import_id = (
+                statement_bank.enforce_unique_import_lines and
+                'data' in line_vals and line_vals['data'] and
+                hashlib.md5(line_vals['data']).hexdigest() or
+                'unique_import_id' in line_vals and
+                line_vals['unique_import_id'] or
+                False
+            )
             if unique_import_id:
                 line_vals['unique_import_id'] = (
-                    (account_number and account_number + '-' or '') +
-                    unique_import_id
+                    statement_bank.acc_number + '-' + unique_import_id
                 )
             if not line_vals.get('bank_account_id'):
                 # Find the partner and his bank account or create the bank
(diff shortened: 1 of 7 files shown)


NL66278:
The test failure is due to the apparently unsolvable ofxparse issue that has nothing to do with this PR.

account_bank_statement_import/models/account_bank_statement_import.py
```diff
@@ -325,12 +337,19 @@ def _create_bank_account(
     def _complete_statement(self, stmt_vals, journal_id, account_number):
         """Complete statement from information passed."""
         stmt_vals['journal_id'] = journal_id
+        statement_bank = self._get_bank(account_number)
         for line_vals in stmt_vals['transactions']:
-            unique_import_id = line_vals.get('unique_import_id', False)
+            unique_import_id = (
+                statement_bank.enforce_unique_import_lines and
+                'data' in line_vals and line_vals['data'] and
+                hashlib.md5(line_vals['data']).hexdigest() or
```
pedrobaeza, line 357: I think not all the values should be taken into account for the hash (I don't know if right now for example `'statement_id'` is a key here, but with that, lines will never be the same). Only operation date, amount and concept should be taken into account.

NL66278:
statement_id is not a part of the line, therefore not taken into account when computing the hash. I am not quit clear what you mean with concept. Anyway there can be many lines with the same date and amount, for instance when selling the same product to different customers. So we also take into account other data in the line, like remote account, remote owner, reference etc. The code in this PR is in production use at the customer for which it was developed (actually for a part by a programmer from the customer), and runs without problems.

account_bank_statement_import_camt/camt.py
```diff
@@ -198,6 +182,7 @@ def parse_statement(self, ns, node):
         transaction_nodes = node.xpath('./ns:Ntry', namespaces={'ns': ns})
         for entry_node in transaction_nodes:
             transaction = statement.create_transaction()
+            transaction.data = etree.tostring(entry_node)
```
StefanRijnhart: Is this not redundant given that this attribute is also added in parse_transaction below?

StefanRijnhart:
As this is an optional feature, :+1:

account_bank_statement_import/models/res_partner_bank.py
```diff
@@ -33,6 +33,18 @@ class ResPartnerBank(models.Model):
     sanitized_acc_number = fields.Char(
         'Sanitized Account Number', size=64, readonly=True,
         compute='_get_sanitized_account_number', store=True, index=True)
+    enforce_unique_import_lines = fields.Boolean(
```
pedrobaeza, line 36: You should specify that this is only for CAMT files.

NL66278: Actually it is only (as yet) implemented for camt files, but there is no reason why it could not be extended to other formats.

account_bank_statement_import_camt/camt.py
NL66278: Yes it is. I decided to remove it from the parsing, as the transaction.data is filled with the element string contents as is, therefore not parsed.

pedrobaeza:
This is conflicting. Is this still relevant? Does Odoo change the way it performs this in newer versions?

NL66278:
@pedrobaeza Still relevant for our customer, and possibly others.

account_bank_statement_import/models/account_bank_statement_import.py
pedrobaeza: What about this comment?

NL66278: I answered that in the comment below this one. So statement_id is not part of the line data, and we need to take the maximum amount of information in the line into account, in order to prevent false positives on the test for duplication.

pedrobaeza: Sorry, I didn't see it, but I insist as in upper versions there's no need of this mechanism. Why here it is needed?

StefanRijnhart: @pedrobaeza what do you mean that it is needed in later versions? Isn't this just a feature that can be forward ported?

pedrobaeza: I'm saying the contrary: you don't need to make this distinction on upper versions. I haven't got any conflict importing lines in upper versions. Did you?

StefanRijnhart: Just did a test with account_bank_statement_camt_oca in Odoo 11, and if I import the same statement twice the duplicate entries are not detected as such so I end up with a corruption of my bank ledger.

NL66278: Our customer does have conflicts, as any customer would have, if they download statements multiple times a day. This is done as their sales orders are only confirmed for delivery when proof of pay is received. So the same lines will be in the statements downloaded in the morning as for a few our later.

pedrobaeza: But then the patch should be done on *_camt module, not on account_bank_statement_import, and more because in >=v9, we don't have control anymore over that module as it's integrated in core.

PR #79, opened 2016-11-10
9.0 account bank statement import camt

hbrunn:


StefanRijnhart (approved):
Tested, works great. You may remove the uuid imort that travis signalled, but otherwise good.

hbrunn:
thanks for the reminder!

account_bank_statement_import_camt/__openerp__.py
```diff
@@ -1,25 +1,9 @@
...
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# © 2013-2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'CAMT Format Bank Statements Import',
-    'version': '8.0.0.3.0',
+    'version': '9.0.0.4.0',
```
pedrobaeza: This should be 9.0.1.0.0

pedrobaeza (approved):
Merging when CIs finish.

Maybe you can review https://github.com/OCA/bank-statement-import/pull/81 in exchange :wink:

pedrobaeza:
There's an strange error in Travis after merging this:

https://travis-ci.org/OCA/bank-statement-import/jobs/182305614

Do you know which module adds account.account records?

hbrunn:
I think I remember something going on with account_coa, but I don't remember exactly what this was, and right now I'm in the middle of something else. Will sort that out next week probably, and then the requested review will happen too

StefanRijnhart:
It's the ledger account created from the bank journal in this module's demo data. https://github.com/odoo/odoo/blob/9.0/addons/account/models/account.py#L435. We should somehow make it so that the coa module is installed before this module is.

pedrobaeza:
Do we need that journal in demo?

pedrobaeza:
It's curious also that this only happens on Transifex stage, but not tests ones. Maybe the install sequence is not the same. This hasn't been detected on the PR as Transifex stage is inhibited for PRs (for not publishing incorrect terms).

StefanRijnhart:
Looks like the problem could be solved by creating the journal in the tests instead. Let me have a look.

PR #80, opened 2016-11-14
[10.0] port OFX module

alexis-via:


alexis-via:
I removed the dependency on l10n_generic_coa, because I don't think it's the right thing to depend on this module just for the unit-tests. In other modules such as the bank-payment module, we don't depend on l10n_generic_coa and we manage to have automated tests with accounting...

pedrobaeza (changes requested):
The search of the bank account is a weak search, but it can match anything. We can provide partner_id key directly (we do it on Spanish statement format). Better to restore it.

alexis-via:
I'm not against re-introducing something, but the current code was non-sense:
1) it was reading transaction.payee (Agrolait for example), searching it on the "bank_name" field of res.partner.bank. This bank_name field on res.partner.bank is:

```
bank_name = fields.Char(related='bank_id.name')
```

So it would search Agrolait among the names of the banks !!! Crazy !
2) in account_bank_statement_import, the 'partner_id' field is not parsed. I understand you may have some additionnal modules that use the 'partner_id' key... could you point us to these modules ? Otherwise, I suggest to introduce a 'partner_name' key and have an OCA modules to match the 'partner_name' key to a partner_id.

pedrobaeza:
About 1), the search is over `owner_name`, not `bank_name`.
About 2), I'm making the partner_id matching in _complete_statement (https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_account_bank_statement_import_n43/models/account_bank_statement_import_n43.py#L335), so yes, we need another step.

alexis-via:
1) In the v9 version, the search is on bank_name:  https://github.com/legalsylvain/bank-statement-import/blob/9.0/account_bank_statement_import_ofx/account_bank_statement_import_ofx.py#L49

But it's a modification made by @legalsylvain in his PR... I understand it was owner_name before.

pedrobaeza:
OK, then I let you decide if you want that match.

alexis-via:
@pedrobaeza I have added a _prepare method for each transaction line; that way, in your localisation, you can now easily add the 'partner_id' key that you want. I think that everybody should be happy now.

pedrobaeza (approved):
If Travis goes well

alexis-via:
The only remaining Travis error is:

```
File "/home/travis/build/OCA/bank-statement-import/account_bank_statement_import_ofx/models/account_bank_statement_import.py", line 82, in _parse_file
'balance_start': ofx.account.statement.balance,
AttributeError: 'Statement' object has no attribute 'balance'
```

which is the same pb as for v9 and is still a big mistery. This problem error doesn't show up when we run the tests locally.

legalsylvain:
Hi @alexis-via ,

I think that [this commit](https://github.com/OCA/bank-statement-import/pull/80/commits/30929b004ca0a7a625bd8318828bf22dcdce7625) broke runbot [check log](https://runbot.odoo-community.org/runbot/build/3272495).

I don't know exactly what is reason to depend on l10n_generic_coa, but it seems necessary.

About travis error, did you try to set a log, in your test file printing ofxparse version ? maybe it's not the same library installed. (Just an idea).

pedrobaeza:
@alexis-via, I have migrated QIF import, and I want to note several things you can improve in current migration:

* You can match partners adding this method: https://github.com/OCA/bank-statement-import/pull/81/files#diff-cfa9dcc04cb5300b5f080cf0b48e78e2R82
* Add the supported file format in the import view: https://github.com/OCA/bank-statement-import/pull/81/files#diff-9d93468a22194f919ca9251c99011324R9
* Python file must be in wizards directory, not in models.

alexis-via:
@pedrobaeza About your points:
1) OK, why not. I usually support that via my module "account_statement_completion_label_simple" https://github.com/akretion/bank-statement-reconcile-simple (which also supports more scenarios). I still have to think if there are potential drawbacks or not.
2) Already done (it was already the case in v9)
3) OK :)

pedrobaeza:
Travis is still red

StefanRijnhart:
You need to move the demo data to the test file, like we did here for camt: https://github.com/OCA/bank-statement-import/pull/82

StefanRijnhart:
I checked the test, and we are being hit by https://github.com/jseutter/ofxparse/issues/107. It works when installing ofxparse with the master branch from github.

pedrobaeza:
FYI, I don't rely on demo data for tests since a lot, because of their inconsistency across versions, their dependency on demo DBs and so on.

alexis-via:
@StefanRijnhart Congratulations, you made Travis green !!! Here we are !

StefanRijnhart (approved):

StefanRijnhart:
@alexis-via thanks for the fruitful collaboration!

pedrobaeza:
This shouldn't be merged until upstream ofxlib merges the fix. Setting this to WIP.

StefanRijnhart:
@pedrobaeza I don't agree. It can be mentioned in the README to install the library from Github, but otherwise it should be fine to merge this.

alexis-via:
I agree with Stefan. It works well with ofxparse 0.14 or 0.15. The bug is only with the automated tests, not when you use this Odoo module as a regular user.

pedrobaeza:
No, because you have changed the origin of the lib. When the patch will be merged and the branch dissapears, then we will have again a red branch here, and another PR needed for changing the source. That's why I say to keep it WIP. You can use it merging this branch in your deployments meanwhile.

StefanRijnhart:
@pedrobaeza it is the master branch of the project, not a bugfixing branch. It is not going anywhere ;-)

pedrobaeza:
It's not the main one: pip install git+https://github.com/jseutter/ofxparse.git, so you can expect all type of broken builds while this guy changes that branch for whatever purpose he has (and he can do it as that branch seems to be his laboratory branch).

alexis-via:
When you look at https://pypi.python.org/pypi/ofxparse/0.15 you will read:
```
Source: https://github.com/jseutter/ofxparse
```

StefanRijnhart:
@pedrobaeza your comment about laboratory branch makes no sense to me at all. When I look at the commits, this branch looks really well managed: the changes are mostly reviewed PRs, and consists of contributions from a variety of people. https://github.com/jseutter/ofxparse/commits/master

NL66278 (approved):
Went through the code and LGTM

pedrobaeza:
OK, let me squash those PEP8 commits and I'll merge the branch.

alexis-via:
@pedrobaeza  Happy to see that you revised your opinion on this.

StefanRijnhart:
@pedrobaeza yeah! Thank you.

grauschnabel:
I tried it by copy account_bank_statement_import_ofx to my addons folder.  Install via Addons works and when I press the import button on the finnace dashboard it now tells me that this is supported:

Open Financial Exchange (.OFX Money)

this wasn't there before so it comes from this addon.

But when I select a file and press import it tells me that this file will not make sence and asks me if I have installed the correct module.  The ofx file I'm using works correct in odoo 8 (i switched to 10 this year) and was created by ofxstatement 0.6.0 from a csv which I got from the austrian raiffeisen bank.

whats wrong here?

grauschnabel:
ok, sorry I was to fast in reporting, upgrading ofxparse from 0.15 to 0.16 made it work.

hubearth:
In my case, upgrading to ofxparse 0.16 did not add the ofx support in the importation. The addon installs normally, but do not change anything. Any leads on this?

EDIT1:
Uninstallin account_bank_statement_import doesn't even remove the import csv options. I'm lost.
Do I need to install another OCA module? account_bank_statement_import comes from official addons.

EDIT2:
Nevermind. I was really lost. I found this module is accessible via the board, but I was going through the importation under settings->accounting->bank account->bank statement->import.
That been said, it would be nice that this option would also refer to ofx module. This would make things clearer.
By the way, thanks for your work.

StefanRijnhart:
@huearth the convention in version 10 set by upstream Odoo is to import through the dashboard.

If you have another module installed that adds a regular import menu and it does not process OFX files properly, then it does not use Odoo's standard bank statement import mechanism and there is nothing the OFX module could do to support it.

hubearth:
@StefanRijnhart 
I don't have another module installed. Doing importation from the dashboard works, but there is two other way to import bank statement. 
1. in the dashboard, if you go to the 'More' button of your bank account, the import statement option will bring you to a csv import form.
2. in accounting->settins->bank account->your bank account, in the 'action' button, there is the bank statement option. From there, you can select import and it will bring you to the same csv importation form.

PR #82, opened 2016-12-09
[RFR] Create demo data during the execution of tests

StefanRijnhart:
This ensures that the demo chart of accounts is already installed when the
bank journal account is created for the bank journal that is needed to
test the import. Otherwise, the demo chart from l10n_generic_coa will fail
to install because there is already a journal account in the system.

https://github.com/OCA/bank-statement-import/pull/79#issuecomment-265771506


pedrobaeza:
Thank you very much! As we can't see the result until this is merged, I fast-track it.

pedrobaeza:
Fixed!!

StefanRijnhart:
Good! Thanks for signalling the problem and taking swift action on this one.

pedrobaeza:
Thanks to you for fixing it!

PR #83, opened 2016-12-10
8.0 camt statement

NL66278:
Need date in statement id for camt, because some banks reuse id's in other years.

### DIFF
--- a/account_bank_statement_import_camt/camt.py
+++ b/account_bank_statement_import_camt/camt.py
@@ -200,13 +200,12 @@ class CamtParser(object):
             transaction = statement.create_transaction()
             self.parse_transaction(ns, entry_node, transaction)
         if statement['transactions']:
-            statement.date = datetime.strptime(
-                statement['transactions'][0].execution_date, "%Y-%m-%d")
-            # Prepend first statement's date to improve id uniqueness factor
-            date = statement.date.strftime('%Y%m%d')
-            if date not in statement.statement_id:
+            execution_date = statement['transactions'][0].execution_date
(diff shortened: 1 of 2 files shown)


NL66278:
Actually relevant tests run OK. Only error in travis is the permanent and seemingly unsolvable ofx module error.

pedrobaeza:
The problem with this approach is that you can query the same statement in several different days, having a different execution time, so it would be imported twice, isn't it?

NL66278:
@pedrobaeza If you query the same statement, all the lines will already have been imported, and so the import will be refused.

pedrobaeza:
OK, and there's no chance that the line IDs are also reused?

NL66278:
The line id's are by default composed of statement_id and line number in the statement. There is a problem when a customer imports several days for the same bank, where transactions in the statements overlap. Because of the different statement id, the duplication of transactions is not detected in that case. To solve that problem an alternative way to determine line uniqueness is provided by this PR: https://github.com/OCA/bank-statement-import/pull/77

StefanRijnhart:
Thank you! :+1:

pedrobaeza (approved):

StefanRijnhart:
I see you cherry-picked the fix for the tests. That has been merged now. Can you rebase https://github.com/OCA/bank-statement-import/pull/83/commits/7c17c497d2ce1340f67d0833e49439dac0ebb328 onto 8.0 to get rid of the redundant commit?

NL66278:
@StefanRijnhart Rebase done.

StefanRijnhart (approved):

PR #87, opened 2016-12-28
[FIX] filter out unused fields

hbrunn:
The parser returns more fields than statement lines can swallow. I'd like to keep it this way, such that whenever some module add new fields (for example, a type of transaction), it will be filled automagically. But to avoid warnings during tests, we need to filter out the extra fields before passing them super.

### DIFF
--- a/account_bank_statement_import_camt/models/account_bank_statement_import.py
+++ b/account_bank_statement_import_camt/models/account_bank_statement_import.py
@@ -22,11 +22,12 @@ class AccountBankStatementImport(models.TransientModel):
             parser = Parser()
             _logger.debug("Try parsing with camt.")
             line_fields = self.env['account.bank.statement.line']._fields
+            extra_fields = ['account_number']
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Does such an extension need to extend the extra_fields parameter (which looks impossible at the moment)? Isn't it easier to mute the logger during the test?

hbrunn:
probably muting the logger is better, thanks. Are you referring to some `extra_fields` outside of this PR I'm not aware of or have you presupposed a generalization already that doesn't exist?

StefanRijnhart:
The presuppose thing, yes.

StefanRijnhart:
@hbrunn do you still need this change?

hbrunn:
well, it makes my tests pass, so yes. But I agree this is quite hacky, so let's put this on hold until I've the time to do this right

pedrobaeza:
@hbrunn have you find another way?

PR #90, opened 2017-01-11
[10.0] OFX: fix start/end balance

alexis-via:
Remove start/end dates in name of statement, because it's better to have it via 2 computed fields

### DIFF
--- a/account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
+++ b/account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
@@ -3,7 +3,7 @@
 import logging
 import StringIO
 
-from odoo import api, models, fields, _
+from odoo import api, models, _
 from odoo.exceptions import UserError
 from odoo.tools import float_is_zero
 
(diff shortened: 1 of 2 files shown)


alexis-via:
The fix on start/end balance is not new ; it restores the code before this commit:
https://github.com/OCA/bank-statement-import/commit/c7c6954df7d1932afe539a17c855cb87acec70b3

account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
```diff
@@ -65,30 +65,23 @@ def _parse_file(self, data_file):
...
         except Exception, e:
             raise UserError(_(
                 "The following problem occurred during import. "
                 "The file might not be valid.\n\n %s") % e.message)
 
+        balance = float(ofx.account.statement.balance)
         vals_bank_statement = {
-            'name': _('Account %s %s > %s') % (
-                ofx.account.number, start_date, end_date),
+            'name': ofx.account.number,
```
StefanRijnhart, line 81: Over at the CAMT corner, we are really struggling with statement identifiers that are recycled every year. Are you sure that is not the case for OFX?

StefanRijnhart (approved):
Just a question

pedrobaeza (approved):
@alexis-via please answer later the question from Stefan, although I merge as it's not related.

PR #91, opened 2017-01-11
[8.0][FIX] Start/end balance on OFX statement

alexis-via:
This is just a "go back" to the code before this commit: https://github.com/OCA/bank-statement-import/commit/c7c6954df7d1932afe539a17c855cb87acec70b3

### DIFF
--- a/account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
+++ b/account_bank_statement_import_ofx/account_bank_statement_import_ofx.py
@@ -83,12 +83,12 @@ class AccountBankStatementImport(models.TransientModel):
                 "The file might not be valid.\n\n %s" % e.message
             ))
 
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Hi @alexis-via, thank you. This does raise some questions. Is this  bug present on the origin 10.0 branch as well? Why is the start and end balance in the test the same amount?
cc. @lmignon

StefanRijnhart:
Self answering my first question: https://github.com/OCA/bank-statement-import/pull/90

alexis-via:
@StefanRijnhart Answering your second question: in the test OFX file, start and end balance are NOT the same: 2516.56 vs 2156.56... look carefully at the second digit :-)

StefanRijnhart (approved):
I Need new glasses!

PR #94, opened 2017-01-24
[FIX] Make sure that search() called with count=True returns correct value

NL66278:
Without these changes a view that shows 80 bank-accounts or more will crash, because search with count=True will return a (invalid) recordset, instead of a count.

### DIFF
--- a/account_bank_statement_import/models/res_partner_bank.py
+++ b/account_bank_statement_import/models/res_partner_bank.py
@@ -45,9 +45,8 @@ class ResPartnerBank(models.Model):
         self.sanitized_acc_number = self._sanitize_account_number(
             self.acc_number)
 
-    @api.returns('self')
-    def search(self, cr, user, args, offset=0, limit=None, order=None,
-               context=None, count=False):
+    @api.model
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
I don't really see the problem with previous method... Maybe a wrong overwrite on your dependency chain?

NL66278:
@pedrobaeza the standard search has these decorators:
    @api.returns('self',
        upgrade=lambda self, value, args, offset=0, limit=None, order=None, count=False: value if count else self.browse(value),
        downgrade=lambda self, value, args, offset=0, limit=None, order=None, count=False: value if count else value.ids)
The type of stuff to be returned was overwritten by just returning self. This went wrong when a count was requested. Rather then adding also the very complicated @returns parameters to the overwritten method, I thought it was easier to use the same approach as in account_bank_statement_advanced/models/res_partner_bank.py of noviat-apps.

pedrobaeza (approved):
Thanks for the explanation

StefanRijnhart (approved):

PR #95, opened 2017-01-27
[ADD] account_bank_statement_import_mt940_uk_hsbc

gdrius:


pedrobaeza:
I think this should go in the UK localization

hbrunn:
I think it shouldn't - choice of banks has nothing to do with your geographical location. Imagine all people engaging in tax evasion had to install the Swiss, Irish and Dutch localization...

but it's an absolutely deadly sin to recycle code without attribution. technically, the code would be a lot better if it depended on https://github.com/OCA/bank-statement-import/tree/8.0/account_bank_statement_import_mt940_base and used it in a similar manner as the dutch mt940 modules.

pedrobaeza:
We have always differ in this point, Holger, but I'll try to explain my reasons:

* You don't need to install all localizations, but go there for searching the corresponding stuff, and install selected modules. It's the same as you go to sale-workflow or purchase-workflow for these areas and install only needed modules, but from both places.
* If you put here all the possible region variants, this repository will be very crowded and we will need to give maintainer permissions to all the possible people that maintains each module, which is unmanageable. Not only that, if I'm a repo maintainer, my job is very hard dealing with country specific modules. Moving this to the localization repository, the maintenance remains more focused.
* The nomenclature itself is not the correct according guidelines. It should be l10n_uk_account_bank_statement_import_mt940_hsbc (The same as we have in l10n-spain the module l10n_es_account_bank_statement_import_n43). This is not relevant for the repository thing, but it also indicates that the correct target is the localization repository.

astirpe:
I agree with @hbrunn. Dealing with customers who just want a working parser for a bank, this repository is the natural place where to look into.

The location of the branch of a bank doesn't make me willing to look into any Odoo localization repo. The Odoo localization concept is something else IMO.

pedrobaeza:
@astirpe have you read my reasoning about the maintenance permission issue?

astirpe:
@pedrobaeza since they are not country-specific modules, I don't see how hosting those modules sparsely in localisation repositories can be an advantage for the maintenance.

The maintainer of the module does not necessarily belong to the country of the module. For example, in my case my company is Dutch and I'm working for customers in the Netherlands, but all the mt940 modules I was working on are of banks in other countries. Is it just a coincidence?

pedrobaeza:
OK, no more objections by my part.

PR #97, opened 2017-03-01
[9.0][MIG] account_bank_statement_import_mt940_base: Migration to 9.0

pedrocs-exo:
I tried to change as little as possible because there are modules that depend on this one:
a) added the parserlib.py file without modifications from the 8.0 account_bank_statement_import because the v9 version will never include it. Also I found no alternative location for the file and couldn't drop it because it's used extensively on the MT940 addons.
b) The parser method was adapted to return 3 values instead of 1 as demanded by the v9 structure.

If you're migrating dependent modules from v8, pls note:
1. Override __check_journal_bank_account_ instead of __find_additional_data_
2. Copy account_bank_statement_import.xml from the CAMT module and adapt it.

Enjoy

### DIFF
--- a/account_bank_statement_import_camt/i18n/fi.po
+++ b/account_bank_statement_import_camt/i18n/fi.po
@@ -1,0 +1,34 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# * account_bank_statement_import_camt
+# 
+# Translators:
+# Jarmo Kortetjärvi <jarmo.kortetjarvi@gmail.com>, 2017
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 9.0c\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2016-12-10 05:00+0000\n"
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Ideally, you would refactor out the parserlib and integrate with Odoo's version of account_bank_statement_import directly. The parserlib is just syntactic sugar. Have a look at how @hbrunn migrated the Camt module to 9.0: https://github.com/OCA/bank-statement-import/commit/4ea2015640f4cbf2cc1ae60d8eb9d88f26839bd1

pedrocs-exo:
I see your point and agree. I was trying to make life easier for the 
authors of the ING and Rabo submodules. But you are one of the authors 
of those modules so I guess I'll try another PR...



----
Pedro Castro Silva
Sossia
911797276


Em qua, 1 de mar de 2017 às 18:18, Stefan Rijnhart (Opener) 
<notifications@github.com> escreveu:

StefanRijnhart:
Thank you for understanding! BTW. as far as I'm concerned, the camt.053 format makes the other Dutch banking import modules that you mention obsolete.

PR #99, opened 2017-03-22
[ADD] account_bank_statement_import_auto_reconcile

hbrunn:
Automatic reconciliation after import
=====================================

This addon allows you have Odoo reconcile transactions from a bank statement import automatically in cases where the matching transaction can be determined unambigously.

Configuration
=============

To configure this module, you need to:

#. go to the journal your bank account uses
#. in the field ``Automatic reconciliation rules``, add at least one rule

Usage
=====

After a journal is configured for automatic reconciliations, it simply happens during an import on this journal. If there were automatic reconciliations, you'll see a notification about that and the lines in question will also show up as reconciled.

Background
==========

Mainly, this module is a framework for conveniently (for programmers) adding new custom automatic reconciliation rules. To do this, study the provided AbstractModels.

Known issues / Roadmap
======================

* add more matching rules:
    * AmountDiffuse (let the user configure the threshold)
    * SameCompany (if A from company C bought it, but B from the same company/organization pays)
    * AmountTransposedDigits (reconcile if only two digits are swapped. Dangerous and a special case of AmountDiffuse)
    * whatever else we can think of
* add some helpers/examples for using the options field
* allow to fiddle with the parameters of configured rules during a specific import


hbrunn:
(tests are missing yet, that's why the work in progress label)

pedrobaeza:
Isn't this be better hosted on bank-statement-reconcile?

hbrunn:
I thought about that too, yes. Whatever the community prefers, but please don't make me rename the module, I just got used to the horribly long names of the models resulting from this...

pedrobaeza:
hahaha, no, this time we get to the same module name.

oca_dependencies.txt
```diff
@@ -0,0 +1,5 @@
+# list the OCA project dependencies, one per line
+# add a github url if you need a forked version
+web
+# TODO: change this back to server-tools once https://github.com/OCA/server-tools/pull/738 is merged
```
melroy89, line 4: Dependency with: https://github.com/OCA/server-tools/pull/738

My proposal is to first approve PR 738.

melroy89 (changes requested):
Change back to server-tool, first merge 738.

account_bank_statement_import_auto_reconcile/models/account_bank_statement_import.py
```diff
@@ -0,0 +1,47 @@
...
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import _, api, fields, models
+
+
+# pylint: disable=R7980
+class AccountBankStatementImport(models.TransientModel):
+    _inherit = 'account.bank.statement.import'
+
+    auto_reconcile = fields.Boolean('Auto reconcile', default=True)
```
thomaspaulb, line 11: Customer thinks this would be better off by default, and I agree. It's not a disaster if someone leaves it off by default and realizes he would be better off having it on, because he can just delete the bank statement and try again. But if it's on and he would have wanted it to be off, recovery is much more tedious.

Aside from that, this is a bit of a lone checkbox. Should we make it configurable on the Bank account journal instead, like we did now for the [batch payments](https://github.com/OCA/bank-statement-import/pull/175/files#diff-9c7d7b289511979198dd286f0ee48534R11)? Another configuration option "Split statements on import" is also there, so it seems like a logical place.

hbrunn: if there are no automatic reconciliation rules configured for the journal, this should do nothing, right?

this checkbox is meant to be able to temporarily turn off the automatic reconciliation, if you never want it, just don't configure rules on the journal

PR #101, opened 2017-03-29
[10.0] OFX: Remove the auto-delete lines with 0 amount 

alexis-via:
I remove the "auto-delete lines with 0 amount" feature because it is now provided by the module account_bank_statement_import (from the official addons), cf this line of code: https://github.com/odoo/odoo/blob/10.0/addons/account_bank_statement_import/account_bank_statement_import.py#L216


legalsylvain (approved):
Sounds good. Thanks.

pedrobaeza (approved):

PR #102, opened 2017-03-30
[9.0][IMP] handle transaction grouping in camt files

ghost:
This is primarily an addition to the camt parser to be able to handle these cases where multiple `NtryDtls/TxDtls` are used to represent different banking lines grouped in one `Ntry`.  Previously the module would drop all `TxDtls` nodes but the first, and use it to complement the information in Ntry.

I also added exhaustive unit tests on the parser using the data file from Therp BV, and also a new one Open Net Sarl is donating.  This serves to document that the changes have probably not broken anything for statements from banks that issue one `Ntry` per transaction, but adds support for those of some banks in Switzerland that do it this way.


ecino:
👍 I tested the changes and it works well.

davidwul:
:+1:  thanks for this contribution. Being able to read grouped transactions is really needed.

ghost:
@davidwul You're welcome!

davidwul:
I actually have two issues while importing non postfinance camt (Raiffeisen):
1. the VCOM entries (grouped ISR transaction) don't have any xxxDtls and are not imported at all. Here is one full entry as example.

         <Ntry>
            <Amt Ccy="CHF">23172</Amt>
            <CdtDbtInd>CRDT</CdtDbtInd>
            <RvslInd>false</RvslInd>
            <Sts>BOOK</Sts>
            <BookgDt>
               <Dt>2017-04-03+02:00</Dt>
            </BookgDt>
            <ValDt>
               <Dt>2017-04-03+02:00</Dt>
            </ValDt>
            <AcctSvcrRef>232185471010000000048749690000</AcctSvcrRef>
            <BkTxCd>
               <Domn>
                  <Cd>PMNT</Cd>
                  <Fmly>
                     <Cd>RCDT</Cd>
                     <SubFmlyCd>VCOM</SubFmlyCd>
                  </Fmly>
               </Domn>
            </BkTxCd>
            <AddtlNtryInf>Crèdit LSV</AddtlNtryInf>
         </Ntry>`

2. The other issue is with salary entries. It's a grouped transaction, but as the DTA must "ask to hide details" then all sub-transactions are at 0. Then when ungrouped is done, the amount for each sub-transaction is the amount of the full transaction.

ghost:
@davidwul 
1. I see, that's a plain bug in my code since it actually ignores a move. With this last commit, you should have 1 transaction for that entry. Would you mind obfuscating sensitive information from the camt you have at hand and making it available for a regression test?

2. What are you doing exactly? Is that a problem with my PR specifically?

davidwul:
Thanks @lbeopennet 
With the last commit, now the entry is importing fine, 
Attached is a file with a salary entry, you'll see how it is construct... 
[test_file_raiffeisen.txt](https://github.com/OCA/bank-statement-import/files/976215/test_file_raiffeisen.txt)

account_bank_statement_import_camt/camt.py
```diff
@@ -118,27 +124,54 @@ def parse_transaction(self, ns, node):
...
+            data = copy(node)
+            for j, dnode in enumerate(data.xpath(
+                    './ns:NtryDtls/ns:TxDtls', namespaces={'ns': ns})):
+                if j != i:
+                    dnode.getparent().remove(dnode)
+            transaction['data'] = etree.tostring(data)
+            yield transaction
+
+    @staticmethod
+    def _deduplicate_names(transactions):
```
StefanRijnhart: I fail to see the general benefit of this change. It also might mess up automatic matching systems that rely on a specific reference format.

Maybe you can keep this in your own customizations? If you really want to share this, make it more general by implementing it at the Odoo bank statement level instead of the camt level.

Do you see my point, or am I missing yours?

ghost: True. It's not super important to us, I just did it out of perfectionnism. I'd be ok with retracting that commit.

account_bank_statement_import_camt/camt.py
```diff
@@ -118,27 +124,54 @@ def parse_transaction(self, ns, node):
...
-        return transaction
+        if len(details_nodes) == 0:
+            yield transaction
+            return
+        transaction_base = transaction
+        for i, dnode in enumerate(details_nodes):
+            transaction = transaction_base.copy()
+            self.parse_transaction_details(ns, dnode, transaction)
+            # transactions['data'] should be a synthetic xml snippet which
+            # contains only the TxDtls that's relevant.
```
ghost: Wasted effort maybe: that field isn't super important afaik and #103 is trying to remove the field anyway. I'd be ok with retracting that commit.

eLBati:
I'm porting this to 8.0 with https://github.com/OCA/bank-statement-import/pull/111

account_bank_statement_import_camt/camt.py
StefanRijnhart: In case you were waiting for me: yes please!

ghost:
nuked 3 "improvements", squashed a bit, and rebased onto recent 9.0.

ghost:
had to adjust golden files using travis log.

ecino (approved):

ghost:
rebase to take #110 into account.

ecino (approved):
I think this is ready to be merged. We have bank statements that are not correctly parsed without those changes, to this is a good improvement.

ghost:
Thanks a lot @ecino. I happened to miss an import when I last rebased so here's yet another rebase, sorry about that.

lbeopennet:
Still failing I know. will fix tomorrow morning .

ecino:
@StefanRijnhart Are you ok with the PR now?

davidwul:
Hi, 
This PR is really needed in the l10n_ch, is it also relevant for other localization, I'm asking non CH persons?
If not we should close this and include this in the swiss localization, if yes merging would be welcome.

PR #103, opened 2017-04-03
account_bank_statement_import_camt: parse unique_import_id; properly …

andrius-preimantas:
…select statement date; remove unexpected values like execution_date, value_date, etc. - standard Odoo does not expect those

PR #105, opened 2017-04-18
[9.0][ADD] camt_details bank statement import

ecino:
New module to read more details in camt statements and put them in bank.statement.line

This is an extension to the standard camt parser.

I chose to put the features in a new module since I am adding some fields in bank.statement.line that not everybody may need. But in some statements, some useful information can be added to better track who is the payer.

### DIFF
--- a/account_bank_statement_import_camt_details/README.rst
+++ b/account_bank_statement_import_camt_details/README.rst
@@ -1,0 +1,48 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Bank Statement Parse Camt Debitor Details
+=========================================
+
+Module to add debitor details in bank statement lines from SEPA CAMT.053 Format bank statement files.
+
+Based on the Banking addons framework.
+
+Known issues / Roadmap
+======================
+
+* None
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/bank-statement-import/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/bank-statement-import/issues/new?body=module:%20account_bank_statement_import%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Emanuel Cino <ecino@compassion.ch>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
+This module should make it easy to migrate bank statement import
+modules written for earlies versions of Odoo/OpenERP.
--- a/account_bank_statement_import_camt_details/models/account_bank_statement_line.py
+++ b/account_bank_statement_import_camt_details/models/account_bank_statement_line.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+"""Add process_camt method to account.bank.statement.import."""
+# © 2017 Compassion CH <http://www.compassion.ch>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import models, fields
+
+
+class AccountBankStatementLine(models.Model):
+    """Add process_camt method to account.bank.statement.import."""
+    _inherit = 'account.bank.statement.line'
+
+    partner_address = fields.Char()
+    partner_account = fields.Char()
+    partner_bic = fields.Char()
+
+    def get_statement_line_for_reconciliation_widget(self):
+        """ Returns the data required by the bank statement
+        reconciliation widget to display a statement line
+        """
+        data = super(AccountBankStatementLine,
+                     self).get_statement_line_for_reconciliation_widget()
+        if self.partner_address:
+            data['partner_address'] = self.partner_address
+        if self.partner_account:
+            data['partner_account'] = self.partner_account
+
+        return data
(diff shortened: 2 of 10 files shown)


ecino:
Not sure why runbot failed, it seems working when connecting to it.

ghost:
Could be useful, and the separate module with extension fields is tidy. Thank you.

I'll feed it camts next monday, shouldn't pose problems but we'll see.

I don't like the parser extension interface you've added because two modules can't independently extend the parser. When it's done with api.Model it just works though. I think this can be done using dynamic multiple inheritance [like so](https://github.com/lbeopennet/bank-statement-import/commits/9.0_105-super), would you be ok with including that? Could be handy if for some reason I have to make #107 into a separate module.

ecino:
I included your fix which looks good 👍 
However, testing again it looks like the reconcile view extension is not showing. I will try to look why.

ecino:
Corrected the reconcile view. The PR is ready.

pedrobaeza:
Why not include this in the main module as an option?

ecino:
I was thinking the added fields in bank.statement.line may not be needed for everyone. But at the end I'm not minding including it in the main module.

pedrobaeza:
OK, I'm not a user of these modules, so I can't judge. Let's see the rest what think.

ghost:
@ecino would you mind using 892ba48e66a0e0f62390125f22cced943cf8caa7 in this PR? I have a module in #107 that uses the same parser-extending hook, and I have factored out the changes to `account_bank_statement_import_camt` in said commit so future merges are easier.

On the branch for this PR, use:
```
git checkout 892ba48e66a0e0f62390125f22cced943cf8caa7
git checkout dd8ba494cf38700f1fa1073823d97edde96cd375 -- account_bank_statement_import_camt_details/
git commit -C 9a05f51e3911003c59fa94a4ba291da3dca5a0db
git cherry-pick 3bac71fa53c929045ddb048b9a650ce0f407f805
```
And you'll get something along the lines of [this.](https://github.com/lbeopennet/bank-statement-import/commits/9.0_camt-details) It makes both PRs effectively orthogonal.

Also here's [evidence that both modules are compatible.](https://travis-ci.org/lbeopennet/ci-staging/builds/229118717)

ecino:
@lbeopennet Done

ghost:
Excellent ! Thank you ! feels so good to just go `git merge refs/pull/{102,105,106,107}/head`.

ecino:
@pedrobaeza It seems we don't get any opinions about having this inside the base camt parser or not. What do you suggest? As I said, my the reason I made another module is because we are adding fields in bank.statement.line model that may not be needed for everyone. But maybe those fields don't bother anyway.

pedrobaeza:
OK, let's continue this way then. Odoo modularity allows it well and there's no limit in the maximum number of installed modules :wink:

account_bank_statement_import_camt_details/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+"""Add process_camt method to account.bank.statement.import."""
+# © 2017 Compassion CH <http://www.compassion.ch>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import models, fields
+
+
+class AccountBankStatementLine(models.TransientModel):
```
ghost: Oops, this should be `models.Model` instead of `models.TransientModel`.  It will break in 10.0. ([evidence](https://travis-ci.org/lbeopennet/ci-staging/jobs/236312784#L712))

ghost (changes requested):

account_bank_statement_import_camt_details/models/account_bank_statement_line.py
ghost: (maybe the runbot build failure is due to that ?)

ecino: Thanks for noticing that! Just fixed it, let's see if runbot will pass now.

ecino:
Any update on this? Is it mergeable?

account_bank_statement_import_camt_details/README.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Credits
+=======
+
+Contributors
+------------
+
+* Emanuel Cino <ecino@compassion.ch>
+
+Maintainer
```
lbeopennet: ok

ghost (approved):

ghost:
In its current form, it wouldn't hurt the base branch, but it conflicts with #103 and #110 .
#103 nukes the 'partner_bic' field in the base parser. It would have to be moved into this modules' parser extension.
#110 is yet another way to extend the parser. Sorry about this, but it's probably the final one.

ecino:
Since #110 is merged, I adapted the code. Tests are passing. @pedrobaeza @hbrunn willing to review?

pedrobaeza:
I don't know too much about this format. I can follow @hbrunn's opinion.

account_bank_statement_import_camt_details/models/camt_parser.py
```diff
@@ -0,0 +1,47 @@
...
+                address_values.append(building_node[0].text)
+            zip_node = address_node[0].xpath(
+                './ns:PstCd', namespaces={'ns': ns})
+            if zip_node:
+                address_values.append(zip_node[0].text)
+            city_node = address_node[0].xpath(
+                './ns:TwnNm', namespaces={'ns': ns})
+            if city_node:
+                address_values.append(city_node[0].text)
+            transaction['partner_address'] = ', '.join(address_values)
```
hbrunn: I'd suggest to add a hook here so that localizations can change presentation here

account_bank_statement_import_camt_details/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,27 @@
...
+from openerp import models, fields
+
+
+class AccountBankStatementLine(models.Model):
+    """Add process_camt method to account.bank.statement.import."""
+    _inherit = 'account.bank.statement.line'
+
+    partner_address = fields.Char()
+    partner_account = fields.Char()
+    partner_bic = fields.Char()
```
hbrunn, line 14: I'm a bit doubtful about those two. Shouldn't they be encoded in https://github.com/OCA/OCB/blob/9.0/addons/account/models/account_bank_statement.py#L363

hbrunn (changes requested):
small nits. Seems you also should work on the coverage of your tests

account_bank_statement_import_camt_details/models/account_bank_statement_line.py
ecino: The IBAN is already encoded in `bank_account_id` thanks to [https://github.com/OCA/bank-statement-import/blob/9.0/account_bank_statement_import_camt/models/parser.py#L98](url)

This is in addition so that if the bank account is not matched to a known partner, we can still see the account information in the reconcile view. By default they are not showing in view.

ecino:
@hbrunn Corrections made, coverage improved. See my other comment for account fields. Thanks for the review.

hbrunn (approved):

ecino:
@pedrobaeza can we merge?

PR #107, opened 2017-04-20
camt: parse the type of reference

ghost:
There's a node that contains a categorisation of what kind of reference the `ref` contains. This module extends the parser to read its value and the statement line model to store it.

### DIFF
--- a/account_bank_statement_import_camt_reftype/README.rst
+++ b/account_bank_statement_import_camt_reftype/README.rst
@@ -1,0 +1,48 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Bank Statement Parse Camt Reference Type
+========================================
+
+Module to add reference information in bank statement lines from SEPA CAMT.053 Format bank statement files.
+
+Based on the Banking addons framework.
+
+Known issues / Roadmap
+======================
+
+* None
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/bank-statement-import/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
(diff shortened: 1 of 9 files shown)


lbeopennet:
This is WIP, unit tests are not ready, , and I've not yet used it in production.

davidwul:
out of curiosity what's the purpose of extracting the type of reference?

ghost:
@davidwul If it says "ISR Reference", I arrange for it to be used for reconciliation. (That's Swiss specific and I'll PR l10n-switzerland soon) Point is to mimic the behaviour of the good old payment slip V11 importer without potentially crashing on non-ISR references.

davidwul:
@lbeopennet Ok, while testing camt054 import (with ISR reference) I just [updated base_transaction_id](https://github.com/davidwul/bank-statement-reconcile/commit/67361807f1846fed7db0d906d489414847e193f9) (as you pointed in other thread). I guess it could be more elegant but so far it give good results. 
What matching strategy are you looking at except matching on ref instead name? 
Are you planning to add completion of partner before having the reconciliation proposition?

StefanRijnhart:
Would it hurt to add this to the core parser, and if it does, would it make sense to encapsulate the core parser in an AbstractModel so that it's methods can be overridden in the usual way?

ghost:
As is, the functionnality is self-contained and optional, so it's a good candidate for being a separate module imho.  It adds a database field, which is a cost that maybe not everyone who uses the core parser wants to pay, so I'd rather not force it into the core parser. Same goes for #105.  As for the `AbstractModel`, I hadn't thought of it and it's clever.  We use Plain Old Python Instance Attributes in the parser, but they seem to work on `Model`, as I've [superficially checked.](https://github.com/lbeopennet/bank-statement-import/commits/9.0%2Bcamt-reftype%2Bcamt-details_2) And it's probably safe to use since they instanciate a Python object [each time.](https://github.com/odoo/odoo/blob/9.0/openerp/models.py#L5381) Should we switch to that?

StefanRijnhart:
I tend to think so. @hbrunn, any opinion on this matter? It's about making the camt parser class inheritable.

hbrunn:
absolutely `AbstractModel`. Should have done that while migrating to some version, so if anyone would do the honors, I'll be happy to review it. Don't forget to add a few hooks in the middle of loops to make customizations simple

ghost:
Excellent! I already did some work on this.  I can look into adding a few hooks later this week, then I'll ask for a review on #110.

ghost:
rebase to take into account the merging of #110

PR #108, opened 2017-04-26
CAMT import for ING bank Netherlands gives error on saving account lines

thomaspaulb:
CAMT import for ING bank Netherlands gives error on saving account entries where `Ntry/NtryDtls/TxDtls/RmtInf/Ustrd` element is missing. Selecting `Ntry/NtryDtls/TxDtls/RltdPties/CdtrAcct/Tp/Prtry` in this case, which contains 'Zakelijke spaarrekening' ("Business savings account") in this case. To me that seems clear enough for the transaction name.

### DIFF
--- a/account_bank_statement_import_camt/camt.py
+++ b/account_bank_statement_import_camt/camt.py
@@ -75,14 +75,19 @@ class CamtParser(object):
         """Parse TxDtls node."""
         # message
         self.add_value_from_node(
-            ns, node, [
+            ns,
+            node,
+            [
                 './ns:RmtInf/ns:Ustrd',
                 './ns:AddtlTxInf',
                 './ns:AddtlNtryInf',
                 './ns:RltdPties/ns:CdtrAcct/ns:Tp/ns:Prtry',
                 './ns:RltdPties/ns:DbtrAcct/ns:Tp/ns:Prtry',
-            ], transaction, 'message', join_str='\n',
-               default=_('No description')
+            ],
+            transaction,
+            'message',
+            join_str='\n',
+            default=_('No description')
         )
         # eref
         self.add_value_from_node(
--- a/account_bank_statement_import_camt/camt.py
+++ b/account_bank_statement_import_camt/camt.py
@@ -21,11 +21,14 @@
 ##############################################################################
 
 import re
+from copy import copy
 from datetime import datetime
 from lxml import etree
+
+from openerp import _
+
 from openerp.addons.account_bank_statement_import.parserlib import (
     BankStatement)
-from copy import copy
 
 
 class CamtParser(object):
@@ -46,7 +49,8 @@ class CamtParser(object):
         return amount
 
     def add_value_from_node(
-            self, ns, node, xpath_str, obj, attr_name, join_str=None):
+            self, ns, node, xpath_str, obj, attr_name, join_str=None,
+            default=None):
         """Add value to object from first or all nodes found with xpath.
 
         If xpath_str is a list (or iterable), it will be seen as a series
@@ -63,6 +67,9 @@ class CamtParser(object):
                     attr_value = join_str.join([x.text for x in found_node])
                 setattr(obj, attr_name, attr_value)
                 break
+        else:
+            if default:
+                setattr(obj, attr_name, default)
 
     def parse_transaction_details(self, ns, node, transaction):
         """Parse TxDtls node."""
@@ -72,7 +79,11 @@ class CamtParser(object):
                 './ns:RmtInf/ns:Ustrd',
                 './ns:AddtlTxInf',
                 './ns:AddtlNtryInf',
-            ], transaction, 'message', join_str='\n')
+                './ns:RltdPties/ns:CdtrAcct/ns:Tp/ns:Prtry',
+                './ns:RltdPties/ns:DbtrAcct/ns:Tp/ns:Prtry',
+            ], transaction, 'message', join_str='\n',
+               default=_('No description')
+        )
         # eref
         self.add_value_from_node(
             ns, node, [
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Have you got an (anonymized) example of such an Entry?

account_bank_statement_import_camt/camt.py
```diff
@@ -67,6 +67,7 @@ def parse_transaction_details(self, ns, node, transaction):
         self.add_value_from_node(
             ns, node, [
                 './ns:RmtInf/ns:Ustrd',
+                './ns:RltdPties/ns:CdtrAcct/ns:Tp/ns:Prtry',
```
StefanRijnhart: To be on the safe side, can you add this 'account type' field at the bottom of the list so that the actual message fields have a higher priority?

thomaspaulb:
Hi Stefan, I have the following anonimized snippet. I could insert it into the test file maybe,

```
<Ntry>
   <NtryRef>021511234567890123000000001</NtryRef>
   <Amt Ccy="EUR">1234.00</Amt>
   <CdtDbtInd>DBIT</CdtDbtInd>
   <Sts>BOOK</Sts>
   <BookgDt>
      <Dt>2017-01-01</Dt>
   </BookgDt>
   <ValDt>
      <Dt>2017-01-01</Dt>
   </ValDt>
   <AcctSvcrRef>abc0e6e273abcdef6d0abf5497b2b6c</AcctSvcrRef>
   <BkTxCd>
      <Domn>
         <Cd>PMNT</Cd>
         <Fmly>
            <Cd>ICDT</Cd>
            <SubFmlyCd>BOOK</SubFmlyCd>
         </Fmly>
      </Domn>
      <Prtry>
         <Cd>00370</Cd>
         <Issr>ING Group</Issr>
      </Prtry>
   </BkTxCd>
   <NtryDtls>
      <TxDtls>
         <RltdPties>
            <CdtrAcct>
               <Id>
                  <Othr>
                     <Id>0574908765</Id>
                     <SchmeNm>
                        <Prtry>PrtryAcct</Prtry>
                     </SchmeNm>
                  </Othr>
               </Id>
               <Tp>
                  <Prtry>Zakelijke Spaarrekening</Prtry>
               </Tp>
               <Ccy>EUR</Ccy>
            </CdtrAcct>
         </RltdPties>
      </TxDtls>
   </NtryDtls>
</Ntry>
```

StefanRijnhart (approved):
Thanks for the change and the clarification!

woepwoep:
I have the same problem. When transferring money from or to the savings account with ING Business account, there is no separate Description (RmtInf.Ustrd) and therefore 'Name' is not filled.

Solution I used is to insert the string 
\<RmtInf\>\<Ustrd\>Inserted by Ronald\</Ustrd\>\</RmtInf\>
right in between \</RltdPties\> and \</TxDtls\>.

Hope this helps.
Thx R

StefanRijnhart:
@woepwoep maybe you can apply this change and verify that you can then import that bank statement without modifications? You can then approve this pull request.

woepwoep:
@StefanRijnhart how do i apply this change? thx R

woepwoep:
This is causing the error:
2017-06-08 10:31:41,068 6279 ERROR odoo_sunflower openerp.sql_db: bad query: INSERT INTO "account_bank_statement_line" ("id", "unique_import_id", "name", "sequence", "partner_name", "partner_id", "amount", "date", "ref", "bank_account_id", "statement_id", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval('account_bank_statement_line_id_seq'), 'NL06INGB0007573381-2017-04-20-2017060812280401700000010004', NULL, 4, NULL, NULL, '199.50', '2017-04-20', NULL, 28, 31, 13, 13, (now() at time zone 'UTC'), (now() at time zone 'UTC')) RETURNING id

Ok i found it. 
this line is to be added when searching for a meaningfull Name:
                './ns:RltdPties/ns:DbtrAcct/ns:Tp/ns:Prtry',

IMHO the fallback should be that, in case Name is still NULL, the camt.py should fill her up with some dummy text like 'NOT PROVIDED'.

thx Ron

woepwoep:
tested, works with this line added below 71 in camt.py

                './ns:RltdPties/ns:DbtrAcct/ns:Tp/ns:Prtry',

StefanRijnhart:
@woepwoep thanks for testing! So this change does not fix the problem for you. @thomaspaulb what do you think of adding a fallback for this case as Ronald is suggesting?

woepwoep:
@StefanRijnhart this PR solves half the problem. the other half is not in this PR yet, but can be solved by adding yet another line in camt.py for the opposite direction of money transfer from and to savings account.

StefanRijnhart:
Ah, thanks for the clarification! Still, a good suggestion to pick up in this PR.

thomaspaulb:
Hey stefan, i think like ronald says that besides what i already added in
the pr, that other one the "dbtraccount" should also be added to the list
of possible names for the line, and then indeed if there is still no line,
its IMHO better to fill it up with something like "unknown" than giving the
exception like now - stefan agree?

Op 9 jun. 2017 06:58 schreef "Stefan Rijnhart (Opener)" <
notifications@github.com>:

StefanRijnhart:
I see, thanks. Sounds good to me!

thomaspaulb:
New entry and fallback added, tested with real ING CAMT export.

danger89:
also not merged?

melroy89 (approved):

thomaspaulb:
@pedrobaeza What will be the best way to get this merged?

pedrobaeza:
You need 2 approvals for being merged according rules. You can review other PRs and ask in exchange that they review this one. Any way, Travis is red, so having CIs correct is another requirement

thomaspaulb:
@pedrobaeza You are right about Travis, I fixed that, but it seems that there are two approved reviews?

image

pedrobaeza:
Yeah, the second one doesn't appear on the bottom. Please fix Travis and tell back.

PR #109, opened 2017-05-01
[IMP][8.0] Extract method from account_bank_statement_import_mt940_base

astirpe:
Proposal to extract a simple method from the parser of module `account_bank_statement_import_mt940_base`.

This could help in case we want to make some modifications to the actual parser.

I had to deal with files being imported that have a particular format. Having this method at disposal, it would be much easier to adapt the parser.

### DIFF
--- a/account_bank_statement_import_mt940_base/mt940.py
+++ b/account_bank_statement_import_mt940_base/mt940.py
@@ -153,7 +153,7 @@ class MT940(object):
                     self.handle_header(line, iterator)
                 line = iterator.next()
                 if not self.is_tag(line) and not self.is_footer(line):
-                    record_line += line
+                    record_line = self.add_record_line(line, record_line)
                     continue
                 if record_line:
                     self.handle_record(record_line)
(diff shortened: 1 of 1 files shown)


hbrunn (approved):
I'm a bit surprised that whatever customization you need can't be done in `handle_record`, but if this helps you, why not?

astirpe:
The method `handle_record()` is invoked later on, when the previously buffered lines are already appended to the `record_line`.

Such buffered lines are actually generated by an iterator that is made by invoking a `split('\n')`. In my import files, there is just a `\n` character to separate the information that I need to parse. So the result is that I cannot handle the information I need by just using the `handle_record()`.

If we look at the modules actually present in this repo (NL ING and NL RABO) we assume that there is always a slash `/` (or any other character other than `\n`) that separates the information to be parsed.
In my particular case, I need a way to handle the `\n` character as separator.

Thanks for the approval, indeed is an help!

hbrunn:
what is this file's record separator then? if crlf, your root problem is `data.replace('\r\n', '\n')` in https://github.com/OCA/bank-statement-import/blob/10.0/account_bank_statement_import_mt940_base/mt940.py#L133 - how about making this configurable?

astirpe:
`\n` is both record separator and line separator.
That's the issue.

StefanRijnhart (approved):

PR #110, opened 2017-05-22
camt: allow modules to enhance the parser

ghost:
see discussion in #107.

PR #119, opened 2017-07-05
10.0 camt improvements (port of 9.0)

ecino:
This is a combination of several improvements made for camt parser in version 9.0 and ported for version 10.0

- #105 Camt details
- #110 Camt parser inheritance
- #102 Handle Transaction groups in Camt


ecino:
I have a strange error in the tests, the statement import does not return any statement_id in the action as it is supposed in [https://github.com/odoo/odoo/blob/10.0/addons/account_bank_statement_import/account_bank_statement_import.py#L60](account_bank_statement_import). When running the tests locally, everything is fine. Maybe someone can help me on this?

ghost:
The failure is caused by [this mock bank statement parser](https://github.com/OCA/bank-statement-import/blob/10.0/account_bank_statement_import_save_file/tests/test_save_file.py#L14) combined with the fact that unit tests are less isolated in Odoo 10.0.  As you can see [here,](https://travis-ci.org/lbeopennet/ci-staging/jobs/252028709#L724) the unrelated mock ends up in the MRO of the wizard in your unit test and the code that's being tested is never reached. This doesn't occur in [9.0](https://travis-ci.org/lbeopennet/ci-staging/jobs/252035100#L722).

I don't know if Odoo SA had a good reason to change the way tests are ran.

ecino:
Thanks for pointing me to that, I did not notice this! Any idea to circumvent the mock for this particular test?

ghost:
@ecino No idea. I don't even know if the new behavior is a bug or a feature and I don't know where to look for an answer.

In the meantime, I suggest you drop #105 from this PR so it has better chances to be merged. That would still benefit the community because we'd have the forward port of #110 sooner. (good for https://github.com/OCA/l10n-switzerland/pull/342) To be completely honest, I'm not even sure #102 belongs in because it hasn't been merged in 9.0.

ecino:
Anyone willing to review?

ecino:
I will do separate pull requests to help the merging process.

PR #130, opened 2017-10-31
[10.0][MIG] Allow camt parser inheritance

ecino:
Modify camt parser to be an Odoo AbstractModel instead of a pure python object.
This allows inheritance in sub-modules to customize the parser and is needed in Switzerland to extend the functionnalities of the parser.

This is already present in version 9 and should be done as well in version 10.


ecino:
I think this should be merged before camt parser will be ported to v11.

ecino:
Can someone review this please?

hbrunn (approved):

astirpe (approved):

pedrobaeza:
Please change module version number

ecino:
Changed version of module

PR #132, opened 2017-11-21
[10.0] [FIX] Solve encoding issue and add AMEX support

sebastienbeau:
This PR solve 3 issues:

- importing AMEX qif bank statement (header is CCARD)
- solving encoding issue by guessing it with the lib "chardet" (optionnal lib)
- make inheritable the way to parse the date

### DIFF
--- a/account_bank_statement_import_qif/wizards/account_bank_statement_import_qif.py
+++ b/account_bank_statement_import_qif/wizards/account_bank_statement_import_qif.py
@@ -12,6 +12,15 @@ from odoo.tools.translate import _
 from odoo import api, models
 from odoo.exceptions import UserError
 
+import logging
+_logger = logging.getLogger(__name__)
+
+try:
+    import chardet
+except (ImportError, IOError) as err:
+    chardet = False
+    _logger.debug(err)
+
 
 class AccountBankStatementImport(models.TransientModel):
     _inherit = "account.bank.statement.import"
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Changing an external dependency in a stable version is not convenient, as it makes all existing installations to break.

sebastienbeau:
@pedrobaeza you are right, I can make it optionnal.

sebastienbeau:
I have changed my code, now I use utf-8 by default (this should work with the majority of case) if you had the chardet lib, it will try to guess the right encoding.
Also I have refactored the code to be able to inherit the way to parse the date in custom module

account_bank_statement_import_qif/wizards/account_bank_statement_import_qif.py
```diff
@@ -12,6 +12,14 @@
 from odoo import api, models
 from odoo.exceptions import UserError
 
+import logging
+_logger = logging.getLogger(__name__)
+
+try:
+    import chardet
+except (ImportError, IOError) as err:
+    _logger.debug(err)
```
pedrobaeza: Put here `chardet = False` instead for easy handling

account_bank_statement_import_qif/wizards/account_bank_statement_import_qif.py
```diff
@@ -20,10 +28,26 @@ class AccountBankStatementImport(models.TransientModel):
     def _check_qif(self, data_file):
         return data_file.strip().startswith('!Type:')
 
+    def _get_qif_encoding(self, data_file):
+        try:
```
pedrobaeza: Don't put the message each time. With the other change, you can do:
```
if chardet:
    chardet.detect(data_file)['encoding']
else:
    return 'utf-8'
```

pedrobaeza (changes requested):
And please change module version number

sebastienbeau:
Thanks for your feedback Pedro !
I have done the fixes

pedrobaeza (approved):

bealdav (approved):

PR #134, opened 2017-11-29
[11.0][MIG] account_bank_statement_import_camt

astirpe:


feketemihai (approved):

melroy89:
Why is this still not merged?

astirpe:
@danger89 because this PR got only one approval.
Did you already try this module? If this is the case, you could write down your feedback and perhaps approve this PR.

melroy89:
Sorry I can only test MT920. I wish I could help you. 😭

CasVissers-360ERP (approved):

BT-astauder:
Just my 2 cents: There exists the same modulename in enterprise. Because odoo.sh adds all addons it can find to the addons_path this name collision matters. What do you think about renaming it starting from 11.0?

StefanRijnhart:
Thank you for the work so far! There have been some updates and you lost the ownership of some of the more recent changes that you did take. Would you mind rerunning the git-format-patch method again? +1 for a rename too (account_bank_statement_import_camt_format?)

astirpe:
@StefanRijnhart yes, you are right, there were many changes applied in the 10.0 branch. Thanks!
I'm recreating the commits history and I'm realizing that the changes made in january and february must be ported to V11 as well. Unfortunately I will not have enough time for it during the next days, so in the meanwhile if someone else wants to do the job and open a new PR I will be happy to close this one.

StefanRijnhart:
No hurry ;-)

pedrobaeza:
I would go for `account_bank_statement_import_camt_oca`. I see no problem on put `oca` in module name and promote brand.

StefanRijnhart:
@pedrobaeza indeed!

PR #136, opened 2017-12-04
[10.0] Find correct bank account when account is IBAN

njeudy:
- When importing OFX file, and having banque account define with IBAN account number (ref), Odoo hangs !!

capture d ecran 2017-12-04 14 16 19

@alexis-via : can you check ?

Now Correctly extract account number from IBAN (if acc_type = 'iban').

TODO:

- remove __logger.debug ?


njeudy:
@alexis-via do you think that we have to update sanitized_acc_number with only account number part when user write FRXXXXXXXXX ?

I use this map data from odoo: https://github.com/odoo/odoo/blob/10.0/addons/base_iban/models/res_partner_bank.py#L81

So most of IBAN country are supported

njeudy:
@gurneyalex @alexis-via  can you review and check this if it's correct ? I think we can find a more generic way .. actualy it support all IBAN account number (all country)

njeudy:
@alexis-via  really need your advice on it :)

account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
```diff
@@ -18,6 +20,22 @@
...
     _inherit = 'account.bank.statement.import'
 
+    def _check_journal_bank_account(self, journal, account_number):
+        res = super(
+            AccountBankStatementImport, self
+        )._check_journal_bank_account(journal, account_number)
+        if not res:
+            e_acc_num = journal.bank_account_id.sanitized_acc_number
+            e_acc_num = e_acc_num.replace(" ", "")
+            validate_iban(e_acc_num)
```
ecino, line 30: I think you should encapsulate this in try clause because this may raise a ValidationError, in case you want to just return False

ecino (changes requested):

gurneyalex (approved):

gurneyalex:
@fclementic2c can you give this a look?

account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
pedrobaeza: As everything is already encapsulated with a general try, I think this can be considered OK.

PR #141, opened 2018-01-04
Add contributing section

melroy89:


pedrobaeza (approved):
Very good idea. Fast-tracking as this doesn't require more review process.

Issue #138, opened 2017-12-21
Field 'ref' is empty when debit card payments are reconciled by our bank


woepwoep:
Odoo v10:

In a restaurant, customers can pay with debit card.
The total amount paid is reconciled the next day by a payment from our bank.
This leads to the following CAMT record:

```
<Ntry>
    <Amt Ccy="EUR">3690.55</Amt>
    <CdtDbtInd>CRDT</CdtDbtInd>
    <RvslInd>false</RvslInd>
    <Sts>BOOK</Sts>
    <BookgDt>
        <Dt>2017-10-02</Dt>
    </BookgDt>
    <ValDt>
        <Dt>2017-09-30</Dt>
    </ValDt>
    <BkTxCd>
        <Prtry>
            <Cd>216</Cd>
        </Prtry>
    </BkTxCd>
    <NtryDtls>
        <TxDtls>
            <BkTxCd>
                <Prtry>
                    <Cd>216</Cd>
                </Prtry>
            </BkTxCd>
            <RmtInf>
                <Ustrd>AFREK. BETAALAUTOMAAT Maestro      REFNR. 897401                      DAT. 20170929 AANT. 41</Ustrd>
            </RmtInf>
        </TxDtls>
    </NtryDtls>
</Ntry>
```

As you can see, the field 'ref' will not be filled, because the indicators are missing
                './ns:RmtInf/ns:Strd/ns:CdtrRefInf/ns:Ref',
                './ns:Refs/ns:EndToEndId',
Whereas the field 'note' will be filled, because the indicators are there
                './ns:RmtInf/ns:Ustrd',
                './ns:AddtlNtryInf',
                './ns:Refs/ns:InstrId',

Suggestion is to have the field 'note' overwrite the field 'ref' whenever the latter is empty.

Thx Ron

pedrobaeza:
@nzaterrence @woepwoep  module implementation has changed a lot since this issue report. Can you check if the problem is still there? If so, please update the PR accordingly.

astirpe:
Our customers reported some issues regarding the `ref` and `note`.
It turned out that actually this module imports the `ref` description in a wrong way; in facts it seems that `ref` and `note` are swapped while importing.

So we adopted a solution, different that the one proposed here. What we do is to invert `ref` and `note` when we import, so `ref` is always present (like in the example of this issue) and `note` could be empty. In any case,  `ref` and `note` should be inverted.

thomaspaulb:
[...]

PR #143, opened 2018-02-01
Fill 'ref' with 'note' when 'ref' is empty

nzaterrence:


thomaspaulb:
Solves #138

thomaspaulb:
@astirpe A question, [here](https://github.com/OCA/bank-statement-import/issues/138) you claim that since your swap of name and note, 'ref' is always filled. But what is the basis for this? If 'name' is filled, couldn't 'ref' still be empty?

astirpe:
@thomaspaulb We didn't experience any case of 'ref' being empty, but if you encountered such issue, for me is a +1 to merge this PR

astirpe:
Superseded by https://github.com/OCA/bank-statement-import/pull/178

thomaspaulb:
@nzaterrence Can you close

PR #144, opened 2018-02-02
10.0 paypal bank statement

sebastienbeau:
This is a new version of paypal bank statement module.
@max3903 @pedrobaeza @adrienpeiffer @alexis-via @mourad-ehm

This module is based on the new version of paypal bank statement.

A previous version of this module was here https://github.com/OCA/bank-statement-import/pull/7


JordiBForgeFlow:
This would be a GREAT tool for OCA's odoo, since it collects the membership money very often from paypal.

sebastienbeau:
Yes It can help the board @jgrandguillaume  ;)

sbidoul:
@jbeficent I just installed this module on the OCA instance.

sebastienbeau:
@sbidoul great !

account_bank_statement_import_paypal/README.rst
```diff
@@ -0,0 +1,59 @@
...
+=============================
+
+This module allows you to import the Paypal CSV files in Odoo as bank statements.
+
+Installation
+============
+
+This module depends on the module *account_bank_statement_import* which
+is available:
+* for Odoo version 8: in the OCA project `bank-statement-import <https://github.com/OCA/bank-statement-import>`
```
mourad-ehm: Odoo version 8 ?

account_bank_statement_import_paypal/models/account_bank_statement_import_paypal.py
```diff
@@ -0,0 +1,265 @@
...
+        for header in HEADERS:
+            if data_file.strip().startswith(header):
+                return True
+        return False
+
+    def _convert_paypal_line_to_dict(self, idx, line):
+        date_dt = datetime.strptime(line[0], self._get_paypal_date_format())
+        rline = {
+            'date': fields.Date.to_string(date_dt),
+            'time': line[1],
```
mourad-ehm, line 69: I had a bad experience where the client could download csv files whose number and order of columns is not always the same. It was a bug source (ex balance was not in column 8 but 10). One solution can be the use of csv.DictReader if we can detect csv lang (fr, en, ...) or add lang in import wizard. Column names will match with HEADERS

sebastienbeau: We do not use the same csv file as before. Indeed the previous version of this module was using the configurable export (where you can specify if you want to use your separtor, you can filter the transaction you want). Now we are using the official statement report (not configuration is possible). So I do not think we will have such issue

mourad-ehm: OK, but the "HEADERS = [ '"Date","Heure","Fuseau horaire","Description","Devise"', .. # French" Contain only 5 columns but here we have at least 18 columns ('origin_transaction_id': line[17]). 

I propose either to put all the columns of the official statement report in the HEADER or to add a comment of the correspondence between the columns used and it index (ex: line [17] # col "transaction origin". Paypal can change the format, or a user may be mistaken and use a configurable export format. In which case it is extremely difficult to debug if we do not know the names of the columns used.

sebastienbeau: I have set the full header to avoid the issue you describe

mourad-ehm: :+1:  Thanks

mourad-ehm (approved):

account_bank_statement_import_paypal/models/account_bank_statement_import_paypal.py
```diff
@@ -0,0 +1,275 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014-2017 Akretion (http://www.akretion.com).
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+from datetime import datetime
+from openerp import models, fields, api, _
+from openerp.exceptions import UserError
```
alexis-via, line 10: We are in v10, so please use "from odoo" instead of "from openerp"

account_bank_statement_import_paypal/models/account_bank_statement_import_paypal.py
```diff
@@ -0,0 +1,275 @@
...
+            except:
+                raise UserError(
+                    _("Value '%s' for the field '%s' on line %d, "
+                        "cannot be converted to float")
+                    % (rline[field], field, idx))
+        return rline
+
+    def _parse_paypal_file(self, data_file):
+        f = StringIO()
+        f.write(data_file)
```
alexis-via, line 98: FYI, you can the 2 lines in 1 : f = StringIO(data_file)

account_bank_statement_import_paypal/README.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Configuration
+=============
+
+In the menu Accounting > Configuration > Accounts > Setup your Bank Accounts, make sure that you have your Paypal bank account with the following parameters:
+* Bank Account Type: Normal Bank Account
+* Account Number: the email address associated with your Paypal account
+* Account Journal: the journal associated to your Paypal account
+
+============
```
alexis-via, line 14: Seems there is something missing here... and the README uses an old template.

alexis-via (approved):
Approved with some remarks

pedrobaeza:
Closing in favor of v11 version #161

PR #146, opened 2018-02-07
[FIX] Fixes #145

### LINKED ISSUE #145: 8.0 [BUG] Bank statement import in Multi-company environment could assign inaccessible partner_id to statement lines
The search for a matching partner to a statement line occurs on [bank account number](https://github.com/OCA/bank-statement-import/blob/8.0/account_bank_statement_import/models/account_bank_statement_import.py#L349), but this model is not multi-company-aware (there is a `company_id` field, but this is only meant to be filled when the bank account belongs to one of your own companies and not for partner records). 

Therefore, it is possible that when person X belonging to company A imports a bank statement, that bank accounts belonging to partners of company B are found, which later on during reconcilation leads to a `res.partner` access error.

thomaspaulb:
Fixes #145

### DIFF
--- a/account_bank_statement_import/models/account_bank_statement_import.py
+++ b/account_bank_statement_import/models/account_bank_statement_import.py
@@ -345,16 +345,22 @@ class AccountBankStatementImport(models.TransientModel):
                 bank_account_id = False
                 partner_account_number = line_vals.get('account_number')
                 if partner_account_number:
-                    bank_model = self.env['res.partner.bank']
-                    banks = bank_model.search(
-                        [('acc_number', '=', partner_account_number)], limit=1)
-                    if banks:
-                        bank_account_id = banks[0].id
-                        partner_id = banks[0].partner_id.id
-                    else:
-                        bank_obj = self._create_bank_account(
-                            partner_account_number)
-                        bank_account_id = bank_obj and bank_obj.id or False
+                    partner_model = self.env['res.partner']
+                    partner_with_bank = partner_model.search([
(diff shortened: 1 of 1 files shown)


astirpe:
@thomaspaulb it seems that tests are failing after your changes, could you check?

thomaspaulb:
@astirpe Several unforeseen stuff addressed.

account_bank_statement_import/models/account_bank_statement_import.py
```diff
@@ -345,16 +345,25 @@ def _complete_statement(self, stmt_vals, journal_id, account_number):
                 bank_account_id = False
                 partner_account_number = line_vals.get('account_number')
                 if partner_account_number:
-                    bank_model = self.env['res.partner.bank']
-                    banks = bank_model.search(
-                        [('acc_number', '=', partner_account_number)], limit=1)
+                    partner_with_bank = self.env['res.partner'].search([
+                        ('bank_ids.acc_number', '=', partner_account_number)
+                    ], limit=1)
```
gfcapalbo, line 350: This is necessary , because it is theoretically possible , in old code to select a bank  with an inaccessible partner_id, therefore raising a AccessDenied error.
By searching  Partner_id first, we are certain banks[0].partner_id will be accessible to current user.

this is possible because there can be 2 banks with same number.

I think this change makes code more reliable.

gfcapalbo (commented):
Request to re-open and merge.
this small change has been useful to avoid an AccessDenied error certain users had on reconciliation, because the selected bank[0].partner_id wasnot accessible to current user.

PR #148, opened 2018-02-20
[FIX] don't cause a warning for the data key

hbrunn:


pedrobaeza (approved):
The patch is correct avoiding the warning, so I'm going to fast-track it as the PR is here for a long time and comes from a known contributor.

PR #149, opened 2018-02-21
[10.0][MIG] camt_details module

ecino:
Migrate camt_details module from version 9.0

See #105 for reference.

### DIFF
--- a/account_bank_statement_import_save_file/tests/test_save_file.py
+++ b/account_bank_statement_import_save_file/tests/test_save_file.py
@@ -9,13 +9,24 @@ from odoo.tests.common import TransactionCase
 
 
 acc_number = 'BE1234567890'
+module_name = 'account_bank_statement_import_save_file'
 
 
 class HelloWorldParser(models.TransientModel):
+    """ Fake parser that will return custom data if the file contains the
+    name of the module. """
     _inherit = 'account.bank.statement.import'
 
     @api.model
     def _parse_file(self, data_file):
+        if module_name in data_file:
+            return self._mock_parse(data_file)
+        else:
+            return super(HelloWorldParser, self)._parse_file(data_file)
+
+    def _mock_parse(self, data_file):
+        """ method that can be inherited in other tests to mock a statement
+        parser. """
         return (
             'EUR',
             acc_number,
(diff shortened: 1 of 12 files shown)


ecino:
I need help on the test results. When launching the test in local instance, everything is fine. The call to the import function [should return](https://github.com/odoo/odoo/blob/10.0/addons/account_bank_statement_import/account_bank_statement_import.py#L60) an action that has the statement_ids put in the context. For whatever reason on travis, it seems that the statement_ids are not put in the context?

ecino:
I fixed the tests that were failing because of overwriting the parse method in test_save_file tests...
@hbrunn willing to review?

account_bank_statement_import_save_file/tests/test_save_file.py
```diff
@@ -16,20 +16,23 @@ class HelloWorldParser(models.TransientModel):
...
     def _parse_file(self, data_file):
-        return (
-            'EUR',
-            acc_number,
-            [{
-                'name': '000000123',
-                'date': '2013-06-26',
-                'transactions': [{
-                    'name': 'KBC-INVESTERINGSKREDIET 787-5562831-01',
+        if self.env.context.get('test_save_file'):
```
hbrunn: the underlying problem here is that the parser model is not cleaned up after the test. That's also difficult to do, so let's not start with it. But instead of the context key (if you really need this, better namespace with your module name), why don't you analyze the content of the file? That actually could for other modules be extended that for tests, we use a file that contains the module's name. That's simple to detect then.

hbrunn (approved):
I'm willing to do a lot of things, usually time is the problem...

Code looks good, just a small proposal

account_bank_statement_import_save_file/tests/test_save_file.py
ecino: Thanks for the review! Time is always the problem. I will change it as you suggest as soon as possible.

ecino:
@hbrunn I just changed the test based on your suggestion.

hbrunn:
thanks!

ecino:
@pedrobaeza Is this one good to merge?

pedrobaeza:
Well, as other times with CAMT, I'm not a direct user of this, but seems to be correct and Holger has already approved it, so I continue with the merging.

PR #152, opened 2018-03-27
[ADD] module to restore placeholder in partner_id field

kaerdsar:
* [ADD] module to restore placeholder in partner_id field of reconciliation widget.
* [IMP] update account_bank_statement_import_qif to show partner name even if the partner is not found.


pedrobaeza:
Please don't use plurals in names. I suggest ` account_reconciliation_widget_partner` or ` account_reconciliation_partner_placeholder`. Any way, AFAIK, there's already another placeholder in the partner combo. Can't you use that one?

kaerdsar:
@pedrobaeza Thanks for the quick check. I like account_reconciliation_partner_placeholder. Don't know about the other placeholder, but I can try to find it. Is it in other OCA module?

pedrobaeza:
I'm talking about standard, but I haven't seen yet in v11. In v10, I find a placeholder in the partner combo box:

seleccion_006

kaerdsar:
@feketemihai could you please take a look?

kaerdsar:
@pedrobaeza I see the combo box empty in the image you attached. Where did you find the placeholder?

pedrobaeza:
I know, I have taken directly a runbot v11 screenshot, but with no possibility of changing data for showing it. As said, I have seen it in v10, but not v11, so maybe it's a removed feature.

kaerdsar:
Ok, I will also check in v10 in order to compare. Thanks

feketemihai:
@pedrobaeza The ideea of the module is to have another combo in case of not matching partner... Which is done by account number,  it's a feature from v8 where Odoo made a assumption of the partner based also on partner name from imported dict... And clicking on the combo will select the partner.

feketemihai:
And it will only show when no partner matching, otherwise will not see any change.

pedrobaeza:
OK, thanks for the clarification.

feketemihai:
@pedrobaeza This is how it looks with the placeholder...
placeholder
Please see difference between partners found and the one that have partner name only

feketemihai:
@kaerdsar Can you please rename the module, not much to review since it's just a small improvement and it's working ok.

kaerdsar:
@feketemihai I already renamed the module to 'account_reconciliation_partner_placeholder'. Do you have any other suggestion for the name?

feketemihai (approved):

account_reconciliation_partner_placeholder/static/src/js/reconciliation/reconciliation_renderer.js
```diff
@@ -0,0 +1,21 @@
+odoo.define('lr_account.ReconciliationRenderer', function (require) {
```
StefanRijnhart, line 1: Would it be good convention to use the full module name for the prefix?

pedrobaeza: Yeah, I think so. We were using that as a 'de facto' convention.

pedrobaeza:
There's an opened PR in Odoo for restoring it in core, not via a module: https://github.com/odoo/odoo/pull/26172

pedrobaeza:
Closing as now there's no need of this thanks to odoo/odoo#26172

PR #142, opened 2018-01-18
11.0 mig account bank statement import ofx

njeudy:

PR #154, opened 2018-04-11
[11.0] [MIG] account_bank_statement_import_ofx

keshrath:


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

keshrath:
Thanks for the hint, I actually thought that's what happens when the pull request is initiated via Github. But I just came to the conclusion that there is already an existing pull request for this migration, so it wouldn't have been necessary :) But now I'll know it for the next time.
Keep up the good work! :)

pedrobaeza:
Indeed this a duplicate of #142

PR #156, opened 2018-04-26
[ADD] [11.0] account_bank_statement_import_mt940_pl_raiffeisen

nedaszilinskas:


pedrobaeza:
This should be in the corresponding localization repository, not here.

nedaszilinskas:
I don't understand. Where is corresponding localization repository? How its different from account_bank_statement_import_mt940_nl_ing and account_bank_statement_import_mt940_nl_rabo modules in branch 10.0?

pedrobaeza:
That's because they weren't moved. I assume this is for poland, so OCA/l10n-poland should be requested.

nedaszilinskas:
Understood. How (or who) do I request OCA/l10n-poland to be created?

nedaszilinskas:
@pedrobaeza how (or who) do I request OCA/l10n-poland to be created?

pedrobaeza:
Please send an email to contribute@odoo-community.org

okuryan:
I have sent request to create this repo

astirpe:
@pedrobaeza we already discussed about which repository should be adopted: https://github.com/OCA/bank-statement-import/pull/95
This repository (bank-statement-import) is correct for this module. Moving the module to a localization repository is not required IMO.

pedrobaeza:
Again this is something specific of the country. It shouldn't be here. The other time you argue it can be for other countries, and I don't continue with the debate for avoiding flames, but my points are still totally valid.

nedaszilinskas:
@pedrobaeza @astirpe , kindly, so what is the final call on this?

pedrobaeza:
I said my preference, but I don't want to block.

feketemihai:
@astirpe @okuryan @nedaszilinskas @pedrobaeza  We already had this problem with romania and Croatia localisation, and the modules were moved into localisation, even it was banking, currency updates... So I think we should keep the same approach. So I vote for moving to localisation just like @pedrobaeza .

PR #111, opened 2017-05-23
IMP account_bank_statement_import_camt: handle transaction grouping in camt files

eLBati:
ported from https://github.com/OCA/bank-statement-import/pull/102

PR #165, opened 2018-07-17
[FIX] `ref` from 10.0 is called `eref` in 8.0

hbrunn:
when merging #111, we overlooked that `ref` is called `eref` in 8.0


pedrobaeza:
Fast-tracking this

PR #171, opened 2018-08-09
[10.0][FIX] fix disambiguation in account_bank_statement_import_camt …

ghost:
…module name

We have the same module name `account_bank_statement_import_camt` for enterprise
https://github.com/odoo/enterprise/tree/10.0/account_bank_statement_import_camt
as for OCA
https://github.com/OCA/bank-statement-import/tree/10.0/account_bank_statement_import_camt
this two modules can't be installed simultaneously.


pedrobaeza:
The problem here is that we are changing a module that is in a stable branch, and thus we will break all the people that are using it currently. The solution I see is to let a symbolic link to the old name, so we have both.

README.md
```diff
@@ -12,7 +12,7 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
-[account_bank_statement_import_camt](account_bank_statement_import_camt/) | 10.0.1.1.3 | CAMT Format Bank Statements Import
+[account_bank_statement_import_camt_oca](account_bank_statement_import_camt_oca/) | 10.0.1.1.3 | CAMT Format Bank Statements Import
```
pedrobaeza: No need to change this. It's done automatically.

account_bank_statement_import_camt_details/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     'website': 'https://github.com/OCA/bank-statement-import',
     'category': 'Banking addons',
     'depends': [
-        'account_bank_statement_import_camt',
+        'account_bank_statement_import_camt_oca',
```
pedrobaeza: Don't change this dependency for the reasons stated.

account_bank_statement_import_camt_oca/i18n/account_bank_statement_import_camt.pot
```diff
@@ -0,0 +1,50 @@
+# Translation of Odoo Server.
```
pedrobaeza: This should be renamed also to `_oca`

pedrobaeza (changes requested):

ghost:
@pedrobaeza now better?

pedrobaeza:
@StefanRijnhart @hbrunn what do you think about this?

leemannd:
Hello @simahawk @yvaucher @gurneyalex can we have your feedback?

account_bank_statement_import_camt/i18n/account_bank_statement_import_camt_oca.pot
```diff
@@ -0,0 +1,50 @@
+# Translation of Odoo Server.
```
simahawk, line 1: this file still have `_oca`

simahawk: plus, is it really needed?

simahawk (changes requested):

simahawk:
the symlink seems an hack but probably is the cheapest way to get to the result.
Not sure about the side effects tho...

account_bank_statement_import_camt/i18n/account_bank_statement_import_camt_oca.pot
ghost: because we in `account_bank_statement_import_camt_oca.pot`
Only things we can have is translation problem.

yvaucher:
@oleksandrpaziuk could you detail a bit more the description of this PR.
My guess is that 2 modules have the same names. Can you confirm? 

If so, are they compatible? To be known, if 2 modules have the same name the first to appear in the addons_path will be used. But it would make a conflict on app.odoo.com thus a renaming is indeed advised.

ghost:
@yvaucher description updated a little bit.

yvaucher:
"this two modules can't be installed simultaneously."

@oleksandrpaziuk to clarify your description do we need them to be installable both at the same time? or one will superseed the other?

To extend my thinking: in Odoo 9.0 we had a module named `web` on both enterprise and core. When installing enterprise you would superseed the core one. In this case I wonder if it can work.

It could, but we might want to ensure

1. no confusion about the enterprise module and the OCA module
2. That noone breaks by inadvertance the law in respect of dependencies with *GPL licenses and proprietary code.
3. the OCA module becomes available on apps.odoo.com.

I think this needs to be renamed anyway.

yvaucher (approved):

ghost:
@yvaucher No, we don't need to install them at the same time. But if you have them both somehow in the system it requires extra moves to remove one and install another with `.cfg` file manipulation. So that is why my proposition is to remove this extra moves by adding the symlink with the different name. 

P.S.
But as practice shows, if use this solution inside docker container it requires explicit recreation of symlink because docker doesn't support related symlink that added as the code.
```
RUN rm -f .../bank-statement-import/account_bank_statement_import_camt_oca && \
    ln -rs .../bank-statement-import/account_bank_statement_import_camt \
           .../bank-statement-import/account_bank_statement_import_camt_oca
```

Issue #157, opened 2018-05-22
[10.0] batch payments wrong amounts


yung-wang:
When I import CAMT statement 053.001.02 the system will split the lines in a batch entry. But the amount it takes is from the total of that entry, not from the txdtls lines.

You will see 2 lines with 429,56 when you import the file, but it should be 329,56 and 100
![selection_001](https://user-images.githubusercontent.com/30628967/40355364-76bbf886-5db6-11e8-8b2b-34fc116c09a9.png)

See anonymized statement

```
<?xml version="1.0" encoding="UTF-8"?>
<Document xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="urn:iso:std:iso:20022:tech:xsd:camt.053.001.02">
  <BkToCstmrStmt>
    <GrpHdr>
      <MsgId>0554789741.2018-05-16</MsgId>
      <CreDtTm>2018-05-22T10:20:21.41+02:00</CreDtTm>
      <MsgPgntn>
        <PgNb>1</PgNb>
        <LastPgInd>true</LastPgInd>
      </MsgPgntn>
    </GrpHdr>
    <Stmt>
      <Id>0554789741.2018-05-16</Id>
      <ElctrncSeqNb>000000000000000097</ElctrncSeqNb>
      <CreDtTm>2018-05-22T10:20:21.41+02:00</CreDtTm>
      <Acct>
        <Id>
          <IBAN>NL38ABNA0554789741</IBAN>
        </Id>
        <Ccy>EUR</Ccy>
        <Svcr>
          <FinInstnId>
            <BIC>ABNANL2A</BIC>
          </FinInstnId>
        </Svcr>
      </Acct>
      <Bal>
        <Tp>
          <CdOrPrtry>
            <Cd>PRCD</Cd>
          </CdOrPrtry>
        </Tp>
        <Amt Ccy="EUR">000000000000149.87</Amt>
        <CdtDbtInd>CRDT</CdtDbtInd>
        <Dt>
          <Dt>2018-05-15</Dt>
        </Dt>
      </Bal>
      <Bal>
        <Tp>
          <CdOrPrtry>
            <Cd>CLBD</Cd>
          </CdOrPrtry>
        </Tp>
        <Amt Ccy="EUR">000000000000120.39</Amt>
        <CdtDbtInd>CRDT</CdtDbtInd>
        <Dt>
          <Dt>2018-05-16</Dt>
        </Dt>
      </Bal>
      <Ntry>
[...]

PR #174, opened 2018-08-17
[11.0][IMP] CAMT: allow import of batch

astirpe:
Following the https://github.com/OCA/bank-statement-import/issues/157#issuecomment-411404477 by @thomaspaulb, here is a proposal to allow to skip the loading the TxDtls details. The enabling of this option is configurable.

### DIFF
--- a/account_bank_statement_import_camt_oca/models/parser.py
+++ b/account_bank_statement_import_camt_oca/models/parser.py
@@ -19,13 +19,13 @@ class CamtParser(models.AbstractModel):
         sign = 1
         amount = 0.0
         sign_node = node.xpath('ns:CdtDbtInd', namespaces={'ns': ns})
-        if not sign_node:
+        if not sign_node and not self.env.user.company_id.camt_import_batch:
             sign_node = node.xpath(
                 '../../ns:CdtDbtInd', namespaces={'ns': ns})
         if sign_node and sign_node[0].text == 'DBIT':
             sign = -1
         amount_node = node.xpath('ns:Amt', namespaces={'ns': ns})
-        if not amount_node:
+        if not amount_node and not self.env.user.company_id.camt_import_batch:
             amount_node = node.xpath(
                 './ns:AmtDtls/ns:TxAmt/ns:Amt', namespaces={'ns': ns})
         if amount_node:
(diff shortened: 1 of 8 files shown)


thomaspaulb:
Haha, apparently we have done this work in parallel... #175 

Let's compare the code.

astirpe:
@thomaspaulb the only relevant difference I can see is that, in your solution, the user can enable the batch mode in a flag present on the wizard; from the wizard you pass the flag value through the various parsing methods.

Instead we had an internal discussion and we decided tho adopt a different approach for the configuration. As you can see, in our solution, we just adopted a global (per company) configuration flag. This because we assume that the user does non need to switch between batch mode or detailed mode: the user will only adopt the same configuration for every import. The flag in the wizard could be confusing in case the user forgets how to set it correctly.

What is your opinion about that?

thomaspaulb:
@astirpe I think the configuration option makes more sense.

yung-wang (approved):
Downloaded the changes and tested it locally. Option true will only create 1 line. Option false will create multiple line. Works fine here.

astirpe:
@OCA/community-maintainers I just triggered a rebuild of runbot, but it still fails:
```
Unable to find image 'oca-bank-statement-import:6f620c2_3' locally
docker: Error response from daemon: repository oca-bank-statement-import not found: does not exist or no pull access.
See 'docker run --help'.
```

account_bank_statement_import_camt_oca/models/parser.py
```diff
@@ -133,6 +133,8 @@ def parse_entry(self, ns, node):
             transaction = transaction_base.copy()
             self.parse_transaction_details(ns, node, transaction)
             yield transaction
+            if self.env.user.company_id.camt_import_batch:
+                break
```
thomaspaulb, line 137: Another difference I've found between our codes is that yours takes the payment reference of the first transaction in the batch and ours uses `Btch/PmtInfId`. I have no idea what is better, but thought I'd mention.

thomaspaulb (approved):

astirpe:
Runbot relaunched. This time with success :sweat_smile:

PR #175, opened 2018-08-17
[8.0][IMP] Add option to aggregate batches in CAMT

thomaspaulb:
A checkbox is added on the import wizard, when pressed, the CAMT import imports batch transactions as one transaction with the cumulative amount. Bank account number is left empty.

### DIFF
--- a/account_bank_statement_import_camt/README.rst
+++ b/account_bank_statement_import_camt/README.rst
@@ -8,6 +8,20 @@ Module to import SEPA CAMT.053 Format bank statement files.
 
 Based on the Banking addons framework.
 
+Configuration
+=============
+
+The user can configure the way CAMT bank statements are imported:
+
+* Go to *Accounting* -> *Configuration* -> *Journals* -> *Journals*
+* Find the Journal that is related to the bank account you are importing for
+* Set the *Aggregate batch transactions (CAMT)* checkbox
+
+If the checkbox is false, the import will load every single line of the TxDtls details;
+instead if it's true, it will load only the total amount of each batch of lines.
+
+To be able to access the configuration settings, the user must enable Technical features.
+
 Known issues / Roadmap
 ======================
 
@@ -30,6 +44,9 @@ Contributors
 
 * Stefan Rijnhart <srijnhart@therp.nl>
 * Ronald Portier <rportier@therp.nl>
+* Andrea Stirpe <a.stirpe@onestein.nl>
+* Tom Blauwendraat <tom@sunflowerweb.nl>
+* Dan Kiplangat <dan@sunflowerweb.nl>
 
 Maintainer
 ----------
--- a/account_bank_statement_import_camt/models/account_bank_statement_import.py
+++ b/account_bank_statement_import_camt/models/account_bank_statement_import.py
@@ -1,25 +1,10 @@
 # -*- coding: utf-8 -*-
+# Copyright 2013-2015 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 """Add process_camt method to account.bank.statement.import."""
-##############################################################################
-#
-#    Copyright (C) 2013-2015 Therp BV <http://therp.nl>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as published
-#    by the Free Software Foundation, either version 3 of the License, or
-#    (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+
 import logging
-from openerp import models
+from openerp import api, models
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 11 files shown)


thomaspaulb (commented):
LGTM

thomaspaulb:
@astirpe Adopted your version, but with some changes because of the old codebase.

account_bank_statement_import_camt/models/account_bank_statement_import.py
```diff
@@ -33,7 +18,9 @@ def _parse_file(self, cr, uid, data_file, context=None):
         parser = self.pool['account.bank.statement.import.camt.parser']
         try:
             _logger.debug("Try parsing with camt.")
-            return parser.parse(data_file)
+            user_id = self.pool('res.users').browse(cr, uid, uid)
+            camt_import_batch = user_id.company_id.camt_import_batch
```
astirpe: @thomaspaulb thanks!
I think you can reduce a lot the code of your PR by avoid to pass `camt_import_batch` through the various methods.
Instead of reading `camt_import_batch = user_id.company_id.camt_import_batch` here and then passing `camt_import_batch` to the parser methods, you can just read the camt_import_batch configuration in the place where you need it. So the above two lines could be moved directly to `def parse_amount()` and `def parse_entry()`.
This way the signatures of the methods will remain untouched.

astirpe (changes requested):

account_bank_statement_import_camt/models/account_bank_statement_import.py
thomaspaulb: That is true, but do you have any idea how could we get to `user_id` from the parser without refactoring the whole parser? There is not even `self.env` in the functions inside the parser.

astirpe: @thomaspaulb you're right! This part of code uses the old v7 api and the uid is not propagated through the parser. Let's keep the solution you adopted, it's ok as it is. Thanks!

account_bank_statement_import_camt/models/parser.py
```diff
@@ -179,10 +180,14 @@ def parse_entry(self, ns, node, transaction):
         if len(details_nodes) == 0:
             yield transaction
             return
+        # elif len(details_nodes) > 1 and camt_import_batch:
```
astirpe: Could you remove this commented code?

astirpe (changes requested):

thomaspaulb (commented):
Could you also update the README file

account_bank_statement_import_camt/tests/test_import_bank_statement.py
```diff
@@ -61,6 +45,25 @@ def test_statement_import(self):
...
+        """Test aggregate import of batch statement."""
+        transactions = [
+            {
+                # remote bank account will be left empty
+                'transferred_amount': -664.05,
+                'value_date': '2014-01-05',
+                'ref': 'TESTBANK/NL/20141229/01206408',
+            }
+        ]
+        self.env.user.company_id.camt_import_batch = True
```
thomaspaulb: We will need to change this to the journal.

account_bank_statement_import_camt/models/account_journal.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# © 2013-2016 Therp BV <http://therp.nl>
```
thomaspaulb: We need to change the sign to the word "Copyright".

account_bank_statement_import_camt/models/parser.py
```diff
@@ -150,7 +135,7 @@ def parse_transaction_details(self, ns, node, transaction):
                     'remote_account'
                 )
 
-    def parse_entry(self, ns, node, transaction):
+    def parse_entry(self, ns, node, transaction, camt_import_batch):
```
thomaspaulb: When changing function definitions like this, we should always expect that there is some module out there that called the original function and now may break. So we should add a default value, like `camt_import_batch=None`, and then also call the function with `camt_import_batch=..`

thomaspaulb:
@Kiplangatdan Found another way, tests are now running again. Could you take care of remaining nitpicks, README and check whether the transaction (partner name, bank account etc) is now loading in the way customer wants?)

thomaspaulb:
@Kiplangatdan

account_bank_statement_import_camt/models/account_journal.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields
+
+
+class AccountJournal(models.Model):
+    _inherit = 'account.journal'
+
+    camt_import_batch = fields.Boolean()
```
thomaspaulb: Can we name this "Aggregate batch transactions (CAMT)"  and translate it to "Samenvoegen batch transacties (CAMT)" in Dutch?

thomaspaulb (commented):
@Kiplangatdan

account_bank_statement_import_camt/models/account_bank_statement_import.py
```diff
@@ -43,7 +43,9 @@ def _complete_statement(self, stmt_vals, journal_id, account_number):
                 else:
                     if batch not in batch_transactions:
                         transaction.pop('partner_name')
```
thomaspaulb: maybe also `, None` here, to be sure.

thomaspaulb:
@astirpe As you can see we refactored this one a lot to be configurable per bank account / journal, what do you think of it now.

account_bank_statement_import_camt/README.rst
```diff
@@ -8,6 +8,19 @@ Module to import SEPA CAMT.053 Format bank statement files.
 
 Based on the Banking addons framework.
 
+Configuration
+=============
+
+The user can configure the way CAMT bank statements are imported:
+
+* Go to *Settings* -> *Configuration* -> *Accounting*
```
astirpe: Here the readme refers to the general settings but you actually allow the configuration from the journal instead.

In our solution for 11.0 (https://github.com/OCA/bank-statement-import/pull/174) we decided to allow the configuration from the general settings because we are convinced that the configuration per journal is not needed.
Did you get that kind of requirements from any customer?

thomaspaulb: Yes we did, because different banks treat batch payments differently.

yung-wang:
@astirpe On the journal / bank account is even more flexible. So if this is possible it would be a great change.

astirpe:
@yung-wang Ok, thanks. I will update https://github.com/OCA/bank-statement-import/pull/174 accordingly.

@thomaspaulb the solution is ok for me. Please update the configuration paragraph in the readme.

thomaspaulb:
@astirpe done.

astirpe (approved):
Thanks!

yung-wang (approved):
looking good

yung-wang (approved):
looking good

yung-wang (approved):
looking good

account_bank_statement_import_camt/models/parser.py
```diff
@@ -182,7 +172,8 @@ def parse_entry(self, ns, node, transaction):
         transaction_base = transaction
         for i, dnode in enumerate(details_nodes):
             transaction = copy(transaction_base)
-            self.parse_transaction_details(ns, dnode, transaction)
+            self.parse_transaction_details(
+                ns, dnode, transaction)
```
hbrunn: what's your rationale for this reformatting?

account_bank_statement_import_camt/models/parser.py
```diff
@@ -247,7 +238,9 @@ def parse_statement(self, ns, node):
         transactions = []
         for entry_node in entry_nodes:
             transaction = statement.create_transaction()
-            transactions.extend(self.parse_entry(ns, entry_node, transaction))
+            transactions.extend(self.parse_entry(
+                ns, entry_node, transaction)
+            )
```
hbrunn: and this one?

account_bank_statement_import_camt/models/account_bank_statement_import.py
```diff
@@ -28,15 +13,49 @@ class AccountBankStatementImport(models.TransientModel):
...
             # Not a camt file, returning super will call next candidate:
-            _logger.debug("Statement file was not a camt file.",
-                          exc_info=True)
+            _logger.debug("Statement file was not a camt file.", exc_info=True)
             return super(AccountBankStatementImport, self)._parse_file(
-                cr, uid, data_file, context=context)
+                data_file)
+
+    @api.model
+    def _complete_statement(self, stmt_vals, journal_id, account_number):
```
hbrunn, line 30: this looks complicated. Is this really something we need to decide per journal? Otherwise, the way #174 does it seems much more straightforward to me

hbrunn (changes requested):
please don't reformat code if not absolutely necessary.

Can we agree on a way to configure this, here it's on journals as opposed to the company in v11.

CC @astirpe

account_bank_statement_import_camt/models/parser.py
thomaspaulb: It was probably that I indented the whole block before with some `if`, then tried to satisfy flake8, and then decided the `if` was not necessary, and de-indented. Let me undo it, also the other one.

account_bank_statement_import_camt/models/account_bank_statement_import.py
thomaspaulb: This was discussed with the customer, but I forgot the exact reason. Have asked him to remind us.

astirpe: @thomaspaulb any news from your customer?

thomaspaulb: Yes: for one bank (ING), a batch payment is given and then they get back a CAMT with 1 line per batch. For another bank (Triodos), the batch comes back with all transactions separately, which is always around 100 lines which then have to be reconciled all 100 with the original batch. That is a lot of extra work. So for that bank, they use the checkbox.

Issue #176, opened 2018-09-24
CAMT parser: Remittance information hidden in note field


sbidoul:
Hi,

This is a question of CAMT users.

In this code:

https://github.com/OCA/bank-statement-import/blob/6fc3202c311761ebc78e98013c11b9cf2b4f5bb2/account_bank_statement_import_camt_oca/models/parser.py#L56-L67

we see that the Remittance information is put in the note field, while the Additional Tx Information is put in the name field.

In reconciliation and visualisation of the bank statement in Odoo, the note field is visible only by hovering over the i icon. However the most relevant information for the user is the remittance information (which incidentally is also the communication field as used by OCA/bank-payment).

What do people think about this?

astirpe:
We have the same issue. In all our customers we installed a patched module, in which `note` and `name` are switched. I'm going to open a PR with the patch soon.

astirpe:
I created 3 PRs for 9.0, 10.0 and 11.0.

@yung-wang I know you are interested in this!

PR #178, opened 2018-10-02
[10.0][FIX] account_bank_statement_import_camt: switch name and note

astirpe:
Fixes https://github.com/OCA/bank-statement-import/issues/176


sbidoul (approved):

yung-wang (approved):
It works great for us

ThomasBinsfeld (approved):

PatrickMallee (approved):
works fine

mkusmic:
Looks like this would solve the same problem we had with CAMT.053 import and <RmtInf><Ustrd> not showing up in name/communication field.

pedrobaeza:
It should be forward-ported to v11

astirpe:
@pedrobaeza here is the PR https://github.com/OCA/bank-statement-import/pull/179

PR #179, opened 2018-10-02
[11.0][FIX] account_bank_statement_import_camt: switch name and note

### LINKED ISSUE #176: CAMT parser: Remittance information hidden in note field

astirpe:
Cherry-picked from V10.0.
Fixes #176


yung-wang (approved):
It works great for us

ThomasBinsfeld (approved):

PatrickMallee (approved):
works fine

astirpe:
I think this one can be merged

PR #183, opened 2018-10-17
[11.0 BUGFIX] Pass bytes to Ofxparse

llacroix:
The ofx files have the correct charset inside their headers and
trying to convert the file to StringIO using a particular encoding
is subject to failures when the file isn't exactly an ascii file.

For example a file encoded with CP1252 with accents on letters like
é or è will not be parser as utf-8 and will fail to load.

Also since OfxParse is supposed to receive a file handle, it is
correctly reading the file header and choosing the appropriate charset
to read the rest of the file. For this reason, pass the bytes as a
ByteIO that doesn't care about the encoding.

### DIFF
--- a/account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
+++ b/account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
@@ -39,7 +39,7 @@ class AccountBankStatementImport(models.TransientModel):
         if not OfxParser:
             return False
         try:
-            ofx = OfxParser.parse(io.StringIO(data_file.decode('utf-8')))
+            ofx = OfxParser.parse(io.BytesIO(data_file))
         except Exception as e:
             _logger.debug(e)
             return False
--- a/account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
+++ b/account_bank_statement_import_ofx/wizard/account_bank_statement_import.py
@@ -38,7 +38,12 @@ class AccountBankStatementImport(models.TransientModel):
     def _check_ofx(self, data_file):
         if not OfxParser:
             return False
-        ofx = OfxParser.parse(io.BytesIO(data_file))
+
+        try:
+            ofx = OfxParser.parse(io.BytesIO(data_file))
+        except ValueError:
+            return False
+
         return ofx
 
     @api.model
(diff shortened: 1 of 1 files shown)


llacroix:
Well that's annoying... Any idea why the projects says build passing but the account move line tests fails while the ofx ones do not fail. 

It's not like my changes have anything related to the other module.

pedrobaeza:
Uhm, somehow this change is affecting the other module, as I have rebuilt the main branch to see if there's a change upstream and it continues still in green.

llacroix:
It's really strange

llacroix:
The only guess I have is that may be it somehow reach that code with something that was supposed to fail but my code make it not fail as bytes could be loaded correctly. So the error never get raised and doesn't return False  which could mean the tests are flawed somehow and the code shouldn't silence the exceptions. I'll give a try cleaning the try/except

llacroix:
Hmmm, yeah... Apparently the tests were passing on the assumption that the `decode` call would raise an exception sometimes... I'll remote the try/except to figure out what's happening.

llacroix:
It's strange, the latest tests failed on the OFX parser as expected as I removed the try except but they passed the create_move_line tests and the log pointing out that they enter my part of code were never executed so I'm completely clueless on this

llacroix:
what the?!? :confused:  I basically removed the `_logger.debug` and now it works... For some reasons, it does feel like there is something wrong with the tests.

moylop260:
it seems like a flaky error unrelated to this change.
I have created a DUMMY PR with your first commit:
 - https://github.com/OCA/bank-statement-import/pull/184

And the result is green.

Could you revert all commits and just store the first one?

Maybe a third-party dependency was failing in the moment of the tests...

Let's to see so close the new reproduction of these kinds of errors... for now we can merge it with minimal changes.

moylop260:
Merged
  - https://github.com/OCA/bank-statement-import/commit/dc22fa39314d6d41b6ef0cda1805fb7c3d2aa466

Thanks!

The new stable travis build is green too:
 - https://travis-ci.org/OCA/bank-statement-import/builds/443777397

llacroix:
Cool thank you. Considering it was a really small fix it was pretty strange

PR #185, opened 2018-10-25
[IMP] Let user choose date format of .qif file import

apollet:
Improve and fix qif import date problem.

Problem was:
If date is "04/10/2018" (4 october) in your bank file, the date of the line is "2018-04-10" (10 april).
Date format is not fixed for qif files so we have to manage all cases.

This PR let user set the date format of the import (day first, month first or year first)


pedrobaeza:
Can you point any QIF format document where states that thing? It's very strange that a format lets this undefined.

apollet:
> Can you point any QIF format document where states that thing? It's very strange that a format lets this undefined.

I did not find any documents where dates are mentioned :(

pedrobaeza:
@alexis-via you were using this also. Any clue about this?

gdgellatly:
I've used it and yh qif has no defined format, and yh I get this error also.

gdgellatly:
A quick google shows that even quicken has a non us date format option for
importing, and myob as well. Safe to say this is a real issue.

pedrobaeza:
OK, then let's go for including this, but you should do it at journal level to be configured once per bank, not when importing the file.

apollet:
@pedrobaeza Good idea !
I will do that ;)

apollet:
Anyone to review this PR ?

PierrickBrun (approved):

bealdav (approved):

pedrobaeza:
Thanks for the extra tests on this!

Can you please forward-port the patch to v11?

PR #189, opened 2018-12-17
[11.0][FIX] account_bank_statement_import_camt_oca: add tag AddtlInf to the name field

astirpe:
This is the fix for https://github.com/OCA/bank-statement-import/issues/187, module `account_bank_statement_import_camt_oca` version 11.0

Issue #187, opened 2018-11-21
[10.0][11.0] account_bank_statement_import_camt_oca: Please add tag AddtlInf to the name field


yung-wang:
Impacted versions: [10.0][11.0]
Module: account_bank_statement_import_camt_oca

When the SDD is not successful (because not enough balance) the bank will send some information in the XML tag AddtlInf when you import the XML CAMT file. In the module this information is not written anywhere. It would be nice to add the information in the name field of Odoo when you import the bankstatement.

```
<NtryDtls>
<TxDtls>
<Refs>
<EndToEndId>DLDLASF-SDD-75114</EndToEndId>
<MndtId>PC.CBA.4897</MndtId>
</Refs>
<BkTxCd>
<Prtry>
<Cd>108</Cd>
</Prtry>
</BkTxCd>
<RltdPties>
<Dbtr>
<Nm>COMPANY BV</Nm>
<PstlAdr>
<AdrTp>ADDR</AdrTp>
<Ctry>NL</Ctry>
<AdrLine>STRAAT 2-6</AdrLine>
<AdrLine>3500 AA UTRECHT</AdrLine>
</PstlAdr>
</Dbtr>
<DbtrAcct>
<Id>
<IBAN>NL78ABNA0123456789</IBAN>
</Id>
</DbtrAcct>
</RltdPties>
<RltdAgts>
<DbtrAgt>
<FinInstnId>
<BIC>ABNANL2A</BIC>
</FinInstnId>
</DbtrAgt>
</RltdAgts>
<RmtInf>
<Ustrd>
Information for the customer
</Ustrd>
</RmtInf>
<RtrInf>
<Rsn>
<Cd>AM04</Cd>
</Rsn>
<AddtlInf>Not enough balance</AddtlInf>
</RtrInf>
<AddtlTxInf>/CSID/NL90ZZZ012345678912</AddtlTxInf>
</TxDtls>
</NtryDtls>
```

The easiest way would change the parser.py and change the line from
`'./ns:RmtInf/ns:Ustrd'`
to
`'./ns:RmtInf/ns:Ustrd|./ns:RtrInf/ns:AddtlInf',`
The information will be written after the Ustrd. Can you have a look if there is a better solution?

pedrobaeza:
How is SEPA related to this repository?

yung-wang:
Sorry forgot to say that when you import the XML CAMT file, the bank will sent extra information in the tag AddtlInf. Do you mean this?

[...]

PR #190, opened 2018-12-18
[10.0][FIX] account_bank_statement_import_camt: add tag AddtlInf to the name field

astirpe:
Backport of https://github.com/OCA/bank-statement-import/pull/189 to V10.
This is the fix for #187, module account_bank_statement_import_camt version 10.0.


astirpe:
@yung-wang

yung-wang (approved):
works fine and the additional info is visible

thomaspaulb (approved):

PR #192, opened 2018-12-23
[FIX] [10.0] CAMT: Add tag AddtlTxInf to the name field (for Triodos)

thomaspaulb:


thomaspaulb:
@astirpe @yung-wang Would this be a problem for you in practice? We need it for Triodos bank, but am not sure if it will mess up other bank's behaviour. Otherwise I have to make it into a journal option, like I did with the batch parsing.

yung-wang:
@thomaspaulb This info is now put in the 'note' field. I think if we add this also to the 'name' field, the information in the field will be very long and maybe too much.

thomaspaulb:
@yung-wang Our problem is that the Triodos bank of Netherlands sometimes fails to supply either of `RmtInf/Ustrd`, `RtrInf/AddtlInf`, `AddtlNtryInf` and then `name` will fall back to `Refs/ns:InstrId`, which (in their version at least) contains only bank-technical gibberish instead of something recognizable to the user. We communicated with them and they say that in such cases `AddtlTxInf` should be used.

My question to you would be if you have encountered any real-life cases for your customer's banks in which all of the things are missing and the `name` field comes to be filled with `Refs/ns:InstrId`. If there are not, this might be a safe edit, as then in your case the `AddtlTxInf` will never be used, only for the Triodos bank. If there are cases, maybe you can tell me which info in that case would be in the `InstrId` and if it's of any help to the customer.

yung-wang:
@thomaspaulb I checked some bankstatements, but cannot find any cases where all the info is missing. So this should be a safe edit. I'm not a developer, so cannot tell if the code is correct.

thomaspaulb:
@yung-wang Thanks

ecino (approved):
I tested it and it works fine. LGTM :+1: 

Can you give me a review back please? #199

PR #199, opened 2019-02-07
[10.0][camt_details] Add more address lines information when parsing camt

ecino:
In some CAMT statement lines, the address can either be well formatted into street, zip, city XML tags, but sometimes it's not the case and it's only provided inside more generic <AdrLine> tags.

This small addition enable to parse more address information from the statement for the later case without changing the current behaviour.

### DIFF
--- a/account_bank_statement_import_camt_details/models/camt_parser.py
+++ b/account_bank_statement_import_camt_details/models/camt_parser.py
@@ -33,6 +33,13 @@ class CamtDetailsParser(models.AbstractModel):
                 './ns:BldgNb', namespaces={'ns': ns})
             if building_node:
                 address_values['building'] = building_node[0].text
+            generic_nodes = address_node[0].xpath(
+                './ns:AdrLine', namespaces={'ns': ns})
+            if generic_nodes:
+                address_values.update({
+                    'address_line_' + str(i): node.text
+                    for i, node in enumerate(generic_nodes)
+                })
             zip_node = address_node[0].xpath(
                 './ns:PstCd', namespaces={'ns': ns})
             if zip_node:
@@ -58,7 +65,8 @@ class CamtDetailsParser(models.AbstractModel):
         """
         Hook for formatting the partner address read in CAMT bank statement.
         :param address_values: dict: all address values found in statement
-            Possible keys are ['street', 'building', 'zip', 'city']
+            Possible keys are ['street', 'building', 'address_line_xxx',
+                               'zip', 'city']
             Not all keys may be present.
         :return: str: formatted address
         """
--- a/account_bank_statement_import_camt/test_files/test-camt053
+++ b/account_bank_statement_import_camt/test_files/test-camt053
@@ -262,8 +262,8 @@
                             <Dbtr>
                                 <Nm>3rd party Media</Nm>
                                 <PstlAdr>
-                                    <StrtNm>SOMESTREET 570-A</StrtNm>
-                                    <TwnNm>1276 ML HOUSCITY</TwnNm>
+                                    <AdrLine>SOMESTREET 570-A</AdrLine>
+                                    <AdrLine>1276 ML HOUSCITY</AdrLine>
                                     <Ctry>NL</Ctry>
                                 </PstlAdr>
                             </Dbtr>


account_bank_statement_import_camt_details/models/camt_parser.py
```diff
@@ -41,6 +41,13 @@ def parse_transaction_details(self, ns, node, transaction):
                 './ns:TwnNm', namespaces={'ns': ns})
             if city_node:
                 address_values['city'] = city_node[0].text
+            generic_nodes = address_node[0].xpath(
+                './ns:AdrLine', namespaces={'ns': ns})
+            address_values.update({
+                'address_line_' + str(i): node.text
+                for i, node in enumerate(generic_nodes)
+                if generic_nodes
```
thomaspaulb: If you really mean `if generic_nodes`, why not place that in front of the whole update statement? Now it seems as if you want to check something on each node.

account_bank_statement_import_camt_details/models/camt_parser.py
```diff
@@ -41,6 +41,13 @@ def parse_transaction_details(self, ns, node, transaction):
                 './ns:TwnNm', namespaces={'ns': ns})
             if city_node:
                 address_values['city'] = city_node[0].text
+            generic_nodes = address_node[0].xpath(
+                './ns:AdrLine', namespaces={'ns': ns})
+            address_values.update({
+                'address_line_' + str(i): node.text
```
thomaspaulb: You may want to update [these examples](https://github.com/OCA/bank-statement-import/pull/199/files#diff-eb6787e4d3e1251b117823745cd8d271R68) with your new keys 'address_line_xxxx'.

About the order of things: I don't know if `AdrLine` and eg. `TwnNm` sometimes appear in the same statement, but in case it happens, I would add these `Adrline` fields between `Street` and `Zip` and not at the end. Otherwise it would look weird.

account_bank_statement_import_camt_details/models/camt_parser.py
```diff
@@ -41,6 +41,13 @@ def parse_transaction_details(self, ns, node, transaction):
...
+                for i, node in enumerate(generic_nodes)
+                if generic_nodes
+            })
```
thomaspaulb: Final suggestion: it would be very easy to add test coverage for your work by just changing [these two lines](https://github.com/OCA/bank-statement-import/blob/10.0/account_bank_statement_import_camt/test_files/test-camt053#L265) to `<AdrLine>` fields instead of StrtNm and TwnNm.

thomaspaulb (changes requested):

ecino:
@thomaspaulb Thanks for the very good suggestions! I made the corrections that you suggested.

thomaspaulb (approved):

ecino:
Can we merge this?

thomaspaulb:
It needs one more positive review and then we can ping a maintainer to merge. @astirpe ?

ecino:
@pedrobaeza ?

pedrobaeza:
I can't say about this format, so merging this blindly is risky.

ecino:
@p-tombez @hbrunn

thomaspaulb:
@woepwoep

thomaspaulb:
@yung-wang

yung-wang:
@thomaspaulb @ecino I would love to review this, but i dont have real life cases of these camt files. Which bank is using this kinda format, maybe i can ask around?

ecino:
[camt.053_P_XXXX_302231941_0_2019040201212234.zip](https://github.com/OCA/bank-statement-import/files/3042263/camt.053_P_XXXX_302231941_0_2019040201212234.zip)
Here is a sample we receive from Postfinance SA in Switzerland.

sbidoul:
This looks good. I've seen other CAMT files with this kind of Postal Address.

@ecino, thanks for this PR. Have you considered extending the test case to cover this, instead of changing the test file?

thomaspaulb:
By changing one test case, the tests as a whole now cover both cases. Why set up a whole new test case?

sbidoul:
@thomaspaulb @ecino correct, I had not noticed there other entries with StrtNm and TwnNm.

sbidoul (approved):

davidwul:
anybody here available to merge this?

PR #205, opened 2019-04-01
[ADD] account_bank_statement_import_mt940_nl_mollie

hbrunn:
MT940 import for files generated by https://mollie.com

The dialect is very similar to what the Rabobank produces, but not exactly the same, so I yanked some code from there without making the two depending on each other.


NL66278 (approved):
The example mt940 file contains a space in the filename, otherwise looks good to me.

bosd (approved):
(Functional test of demo file)
LGTM! :tada:

PR #212, opened 2019-05-15
[12.0][ADD] account_bank_statement_import_online: online bank statements

alexey-pelykh:
This module provides base for building online bank statements providers.

To configure online bank statements provider:

#. Go to *Invoicing > Configuration > Bank Accounts*
#. Open bank account to configure and edit it
#. Set *Bank Feeds* to *Online*
#. Select online bank statements provider in *Online Bank Statements (OCA)*
   section and configure provider-specific settings.

or, alternatively:

#. Go to *Invoicing > Overview*
#. Open settings of the corresponding journal account
#. Switch to *Bank Account* tab
#. Set *Bank Feeds* to *Online*
#. Select online bank statements provider in *Online Bank Statements (OCA)*
   section and configure provider-specific settings.


pedrobaeza:
Why not including a main provider in this base module - the same as `currency_rate_update` (or at least in the same PR but on other module) for testing all together?

alexey-pelykh:
@pedrobaeza as of now, I don't have simple-enough candidates to be included. It's either transferwise, or paypal, or payoneer, etc. - yet neither of those are simple enough to maintain within base module.

alexey-pelykh:
but, it's likely to be included as separate module in this PR, we'll see in couple of weeks

pedrobaeza:
OK, choose one of them (the simplest one?) for being able to test all together. I'm wondering if it's worth to have this base module that is only adding the option. Shouldn't be more convenient to directly add an option for each provider in the "Bank Feed" selection?

alexey-pelykh:
Too many copy-paste, this module is far for completion, so far it's just a draft. Likely it will have base code / provider code ratio as `currency_rate_update`. Let's see how it goes, but testing will definitely be done with a reference implementation

account_bank_statement_import_online/data/account_bank_statement_import_online.xml
```diff
@@ -0,0 +1,21 @@
...
+<!--
+    Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+    Copyright 2019 Dataplug (https://dataplug.io)
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+
+    <record model="ir.cron" id="ir_cron_account_pull_online_bank_statements">
+        <field name="name">Pull Online Bank Statements</field>
+        <field name="interval_number">5</field>
```
NL66278: Better make this by default once every hour. At the most. The companies I work with usually import bank-statements once a day. Performance can suffer if there are too many very frequent cron jobs.

alexey-pelykh: Thanks, good point, adjusted.

alexey-pelykh:
For testing purposes, use ?debug - it allows to select Dummy as provider

account_bank_statement_import_online/models/account_journal.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+    @api.multi
+    def _update_online_bank_statement_provider_id(self):
+        OnlineBankStatementProvider = (
+            self.env['online.bank.statement.provider']
+        )
+        for journal in self.filtered('id'):
+            old_provider_id = journal.online_bank_statement_provider_id
+            journal.online_bank_statement_provider_id = False
```
NL66278, line 58: I think there will be a lot of cases where the record for the OnlineBankStatementProvider will be unlinked, and with the same values recreated again, causing a lot of unneeded db I/O and misleading information in date_created. Better first check wether there is an existing record that satisfies the values in the journal and only unlink+ create when needed,

NL66278: This will also force the user the redo the configurations he/she did on the statement provider.

NL66278 (changes requested):
See comments

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -0,0 +1,331 @@
...
+                date_until = provider.next_run
+                provider._pull(date_since, date_until)
+
+        _logger.info('Scheduled pull of online bank statements complete.')
+
+    @api.multi
+    def _obtain_statement_data(
+        self, date_since, date_until
+    ):  # pragma: no cover
+        """Hook for extension"""
```
etobella, line 385: It would be interesting an example of how data should be returned

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -0,0 +1,331 @@
...
+                    )
+                    break
+                statement_date = provider._get_statement_date(
+                    statement_date_since,
+                    statement_date_until,
+                )
+                if not data:
+                    statement_date_since = statement_date_until
+                    continue
+                lines_data, statement_values = data
```
etobella, line 204: Maybe, this could be a list of tuples. I mean, sometimes you receive the data from multiple bank accounts on the same call. For example, if you ask the bank all the information of your accounts he could give you all the related bank accounts of your user, that could be several for company or even, all the companies at the same time.

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -0,0 +1,331 @@
...
+                )
+                date_until = provider.next_run
+                provider._pull(date_since, date_until)
+
+        _logger.info('Scheduled pull of online bank statements complete.')
+
+    @api.multi
+    def _obtain_statement_data(
+        self, date_since, date_until
+    ):  # pragma: no cover
```
etobella: Why the no cover? You can test with the dummy you added.

account_bank_statement_import_online/models/account_journal.py
alexey-pelykh: Thanks, valid point!

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: Thanks!

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: There's a Dummy provider that generates sample data, added a hint

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: But it is expected to receive data for one specific journal only, thus making this a list would over-complicate the code IMHO

etobella: It might be harder, but this can happen (at least is what we found on our companies :cry: )

alexey-pelykh: In case you make overhead calls to banking API and filtering - how much the impact is? (In timing, or in cost per API call)? And I still don't have a good idea about how to enhance, taking into account that provider is attached to a journal, and thus having one provider for multiple journals would make this too complex to maintain

etobella: It might be true, but sometimes they do not allow this. It is just my experience with some banks.

etobella: However, this is not blocking, but it was interesting to consider.

alexey-pelykh: it would be great to hear out some more details on these specific banks, see API reference etc.

alexey-pelykh:
@NL66278 please re-review

NL66278 (approved):
:+1: Thanks for creating this module!

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: As from what I've seen so far, everything is account-centric or at least holder-centric

alexey-pelykh:
@jbeficent @tarteo if you don't mind spending a spare minute for this module review 🙏

account_bank_statement_import_online/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+        'Brainbean Apps, '
+        'Dataplug, '
+        'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/bank-statement-import/',
+    'license': 'AGPL-3',
+    'category': 'Accounting',
+    'summary': 'Online bank statements update',
+    'depends': [
+        'account',
+        'account_bank_statement_import',
```
daramousk, line 18: You need to add the widget dependency here that you are using

daramousk (changes requested):

account_bank_statement_import_online/__manifest__.py
alexey-pelykh: Thanks, fixed!

alexey-pelykh:
@daramousk @etobella please re-review

alexey-pelykh:
@pedrobaeza I guess this one is ready to be reviewed in details

alexey-pelykh:
Sorry @aheficent @mreficent but maybe you may chime in here as well

artemsenko (approved):
functional test

alexey-pelykh:
@OCA/banking-maintainers I'd like to bring this module (and derivatives) to your attention

alexey-pelykh:
@sbidoul maybe you can help with this? To my feeling, that's quite a demanded feature

account_bank_statement_import_online/__init__.py
```diff
@@ -0,0 +1,5 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import models
+from . import wizards
+from .tests import online_bank_statement_provider_dummy
```
etobella, line 5: Why this?

alexey-pelykh: Otherwise that model won't be available during tests. It's either this (and placing it under tests) or moving it to models

AndrewPivnenko (approved):

Issue #209, opened 2019-05-04
wrong import for CAMT files containing statements for multiple accounts


sbidoul:
When a CAMT file contains multiple `Stmt` elements for different accounts, Odoo creates all statements for the last account found in the file.

Odoo apparently does not support this, since `_parse_file` returns only one account number.

Until this is supported upstream, the CAMT parser should raise an error in this situation instead of creating incorrect statements.

sbidoul:
Upstream issue: https://github.com/odoo/odoo/issues/33162

PR #213, opened 2019-05-20
Manage multi-bank-account in statement import

sbejaoui:
This PR propose a fix for the issue #209.

With this PR:

- Consider the bank account number related to the bank journal on which the file is imported. all statements for other account numbers are ignored.
- Consider the currency related to the bank journal on which the file is imported. all statements in a different currencies are ignored.
- Consider the same use cases when the zip file contain files with statements on different bank account numbers or currencies.
- Manage the case when the CAMT file contains more that one document. (Known issue in BNP France CAMT files)

### DIFF
--- a/account_bank_statement_import_camt_oca/models/account_bank_statement_import.py
+++ b/account_bank_statement_import_camt_oca/models/account_bank_statement_import.py
@@ -4,6 +4,7 @@ import logging
 from io import BytesIO
 import zipfile
 from odoo import api, models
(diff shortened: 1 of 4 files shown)


sbejaoui:
@sbidoul can you relabel this PR in Need review

alexey-pelykh:
not the person, but can do

sbejaoui:
> not the person, but can do

Thanks :)

Issue #28, opened 2015-06-11
Wrong start and end balance


sanderkruger:
The OFX importer doesn't calculate the start and end balance correctly. Logically, the balance on a statement is the balance at the statement date. All mutations on the statement must have preceded the statement date (you can't see the future). Hence, the balance at the statement date is the end balance.

The code calculates start and end balance:

``` python
            'balance_start': ofx.account.statement.balance,
            'balance_end_real':
                float(ofx.account.statement.balance) + total_amt,
```

But it should rather do:

``` python
            'balance_end_real': ofx.account.statement.balance,
            'balance_start':
                float(ofx.account.statement.balance) - total_amt,
```

pedrobaeza:
For which version?

sanderkruger:
This is in the latest version, 8.0.1

pedrobaeza:
This is a backport from v9 version Odoo version, so you should report in odoo/odoo the problem.

sanderkruger:
Thanks, for the info. This issue is a duplicate of https://github.com/odoo/odoo/issues/3003

PR #216, opened 2019-06-26
[FIX][9.0] balance calculation inverted on ofx import

remi-filament:
Start and end balance for imported statement were incorrectly calculated

Balance retrieved from ofx file is the the final one. You cet the starting zone by removing transactions from final balance.
Note : this is also alignment with the behaviour of this module in v10.0 onwards

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ sudo: false
 cache: pip
 
 addons:
+  # By default postgresql-9.1 is installed but there is issue related:
+  #  https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935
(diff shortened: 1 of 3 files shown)


remi-filament:
Fixes issue #28

vbarrier:
Hello, when this issue will be merged?

remi-filament:
Hi @pedrobaeza, I am trying to understand why Travis CI fails, but I can't figure it out !
Any help would be welcome... :sos: 

So far I understand that Travis build is not able to connect to database, I have checked wrt the latest valid build on bank-statement-import in v9 was made 3 months ago with a completely different environment (Ubuntu Trusty iso Ubuntu Xenial now). Then I am lost...

PR #217, opened 2019-07-03
[12] [FIX] exception when using multiple files type

vbarrier:
fix exception when using multiple files type (ofx for example)

### DIFF
--- a/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
+++ b/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
@@ -66,7 +66,11 @@ class AccountBankStatementImport(models.TransientModel):
 
     @api.model
     def _check_paypal(self, data_file):
-        data_file = self._get_paypal_str_data(data_file)
+        try:
+            data_file = self._get_paypal_str_data(data_file)
+        except Exception as e:
+            _logger.debug(e)
+            return False
         if not self.paypal_map_id:
             return False
         headers = self.mapped('paypal_map_id.map_line_ids.name')
--- a/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
+++ b/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
@@ -66,12 +66,12 @@ class AccountBankStatementImport(models.TransientModel):
 
     @api.model
     def _check_paypal(self, data_file):
+        if not self.paypal_map_id:
+            return False
         try:
             data_file = self._get_paypal_str_data(data_file)
         except UnicodeDecodeError as e:
             _logger.debug(e)
-            return False
-        if not self.paypal_map_id:
             return False
         headers = self.mapped('paypal_map_id.map_line_ids.name')
         file_headers = data_file.split('\n', 1)[0]
--- a/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
+++ b/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
@@ -68,7 +68,7 @@ class AccountBankStatementImport(models.TransientModel):
     def _check_paypal(self, data_file):
         try:
             data_file = self._get_paypal_str_data(data_file)
-        except Exception as e:
+        except UnicodeDecodeError as e:
             _logger.debug(e)
             return False
         if not self.paypal_map_id:


pedrobaeza:
What is the error you get?

vbarrier:
I got an error using this module with another (bank-statement-import-ofx) 

```
Erreur:
Odoo Server Error

Traceback (most recent call last):
  File "/odoo/odoo-server/odoo/http.py", line 654, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/odoo/odoo-server/odoo/http.py", line 312, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/odoo/odoo-server/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/odoo/odoo-server/odoo/http.py", line 696, in dispatch
    result = self._call_function(**self.params)
  File "/odoo/odoo-server/odoo/http.py", line 344, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/odoo/odoo-server/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/odoo/odoo-server/odoo/http.py", line 337, in checked_call
    result = self.endpoint(*a, **kw)
  File "/odoo/odoo-server/odoo/http.py", line 939, in __call__
    return self.method(*args, **kw)
  File "/odoo/odoo-server/odoo/http.py", line 517, in response_wrap
    response = f(*args, **kw)
  File "/odoo/odoo-server/addons/web/controllers/main.py", line 966, in call_button
    action = self._call_kw(model, method, args, {})
  File "/odoo/odoo-server/addons/web/controllers/main.py", line 954, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/odoo/odoo-server/odoo/api.py", line 749, in call_kw
    return _call_kw_multi(method, model, args, kwargs)
  File "/odoo/odoo-server/odoo/api.py", line 736, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/odoo/odoo-server/addons/account_bank_statement_import/account_bank_statement_import.py", line 37, in import_file
    currency_code, account_number, stmts_vals = self.with_context(active_id=self.ids[0])._parse_file(base64.b64decode(self.data_file))
  File "/odoo/custom/addons/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py", line 207, in _parse_file
    paypal = self._check_paypal(data_file)
  File "/odoo/custom/addons/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py", line 69, in _check_paypal
    data_file = self._get_paypal_str_data(data_file)
  File "/odoo/custom/addons/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py", line 53, in _get_paypal_str_data
    data_file = data_file.decode(self._get_paypal_encoding())
  File "/odoo/python_env/lib/python3.6/encodings/utf_8_sig.py", line 23, in decode
    (output, consumed) = codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb0 in position 7919: invalid start byte
```

This happens,  when I upload an ofx file. Odoo tries to use use paypal import module before trying to use the ofx module and obviously it fails.

pedrobaeza:
Maybe you can restrict the except part to that error then

vbarrier:
Are you sure? 

I'm not fully agree, I think any module should failed silently at this state. It is only to "check if I should use the module for this kind of file" if not let try another module if exits.

pedrobaeza:
Yes, because you can possibly mask other kind of exception in case of legitimate Paypal files, like the one having a paypal map mapping in the journal but the header doesn't conform, or a date format mismatch.

vbarrier:
Ok, Thank you for the review @pedrobaeza :)

account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
```diff
@@ -66,7 +66,11 @@ def _paypal_convert_amount(self, amount_str):
 
     @api.model
     def _check_paypal(self, data_file):
-        data_file = self._get_paypal_str_data(data_file)
+        try:
+            data_file = self._get_paypal_str_data(data_file)
+        except UnicodeDecodeError as e:
+            _logger.debug(e)
+            return False
         if not self.paypal_map_id:
```
pedrobaeza: I'm thinking that it's better also to put this at the beginning. Would you mind?

pedrobaeza (approved):
Thanks for this

StefanRijnhart (approved):

PR #222, opened 2019-08-29
Add German date format

wildi1:


pedrobaeza (approved):

benbrich (approved):
Just tested and it works fine! How about merging this one?

benbrich:
@pedrobaeza is there a reason, why not merging this one? Any Problems with the code in other countries?

pedrobaeza:
I'm not anymore OCA's community manager and I'm not monitoring PRs. This should corresponds to the PSC (as in the past, but I served as wildcard). I can do it one last time, but now this PR is contaminated with 4 commits more than when I approved it.

pedrobaeza (changes requested):
Remove extra commits, that also performs other things.

benbrich:
Hi and thanks, I understand! Wanted to address @StefanRijnhart first, but thought you are on the PSC and on the topic. Couldn't find a maintainer in the readme. Will remember next time...

benbrich:
@wildi1 would be great if you could fix this soon! :-)

account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal.py
```diff
@@ -156,7 +156,7 @@ def _post_process_statement_line(self, raw_lines):
             # transaction of currency change if yes merge the transaction
             # as for odoo it's only one line
             cline = currency_change_lines.get(line['origin_transaction_id'])
-            if cline:
+            if cline and (cline['amount'] * line['amount'] > 0):
```
StefanRijnhart, line 159: I'm slightly puzzled by this line. It looks like you encountered a statement line with the currency amount having the wrong sign. Can you clarify this a bit, or even better, add a test for this situation?

wildi1:
@pedrobaeza The first PR was made in Aug 19. I thought it will be merged soon after you approved it and so I forgot that it is not merged yet. Therfore my other commits have contaminated this PR.

@StefanRijnhart This schould be part of an other Pull Request I already made. I have split them now in two separate PRs.

Here is the new PR for this request.
#256

wildi1:
@StefanRijnhart Here is the other PR and the explenation of the behaviour with the wrong sign.
 #258

wildi1:
@benbrich I have fixed it soon :) When you are using Paypal with different currencies you should also have a look at #258

PR #224, opened 2019-09-06
[10.0][FIX] prevent assigning parner with default name

i-vyshnevska:
if statement has entries with empty ref (which assign to '/' in odoo) reconciliation widget cannot properly find partner_id, this PR prevents such assignment so user will see initial name given by statement

### DIFF
--- a/account_bank_statement_import_camt/models/account_bank_statement_line.py
+++ b/account_bank_statement_import_camt/models/account_bank_statement_line.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class AccountBankStatementLine(models.Model):
+
+    _inherit = "account.bank.statement.line"
+
+    @api.multi
+    def write(self, vals):
+        """
+        Purpose of this hook is catch updates for records with name == '/'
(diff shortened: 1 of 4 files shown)


account_bank_statement_import_camt/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,30 @@
...
+        """
+        Purpose of this hook is catch updates for records with name == '/'
+
+        In reconciliation_widget_preprocess, there is attempt to assign
+        partner into statement line, this assignment relies on name,
+        during import name setup to '/' for records without it
+        and this makes search results wrong and partner assignment randomly
+        """
+        recordset = self
+        if self.env.context.get('no_reassign_empty_ref'):
```
grindtildeath: If https://github.com/odoo/odoo/blob/c7c68bab842aba5300f3a9deec28df6e17b40559/addons/account/models/account_bank_statement.py#L328 is the update you want to avoid, maybe better do so to avoid the write altogether? (we shouldn't have more than one record in self anyway)
```
if (
    self.env.context.get('no_reassign_empty_ref') 
    and len(self) == 1 
    and len(vals.keys()) == 1 
    and 'partner_id' in vals and 
    and self.name == '/'
): 
    return True
```

i-vyshnevska: @grindtildeath thanks, updated

PR #225, opened 2019-09-06
[12.0][FIX] account_bank_statement_import_paypal: set right current and partner in tests

ernestotejeda:
Set right current and partner in `test_import_paypal_file`

Cc @Tecnativa


pedrobaeza (approved):
Fast tracking this as needed for tests

PR #229, opened 2019-10-01
[12.0][MIG] account_bank_statement_import_mt940_base

feketemihai:


wytzejan:
@feketemihai when do you think this PR is ready? I'm eagerly waiting for this one :)

feketemihai:
@wytzejan If i will have time, i will recheck tomorrow, since i think only some small issues with unit test.

feketemihai:
@sbidoul @hbrunn @astirpe Can you check this one, please. I restructure the module and use RABO as a general implementation, for other formats, we should inherit the methods to give the right returns, added methods for get_header_lines, get_header_regex, get_footer_regex, get_codewords, get_tag_61_regex... to be more easy. I will try to migrate the ING one for example.

hbrunn:
for my customers, MT940 is pretty much obsolete (they all use CAMT now), so I probably won't be of much help here

pedrocs-exo:
Hi, will this PR ever be ready and approved? The code works and it's still useful in some cases. How can I help?

account_bank_statement_import_mt940_base/models/mt940.py
```diff
@@ -0,0 +1,359 @@
...
+            'currency': None,
+            'account_number': None,
+            'statement': None,
+        }
+        if not header_lines:
+            header_lines = self.get_header_lines()
+        for match in matches:
+            self.is_mt940_statement(line=match)
+            iterator = '\n'.join(
+                match.split('\n')[1:-1]).split('\n').__iter__()
```
pedrocs-exo, line 186: Should be:

```
            iterator = '\n'.join(
                match.split('\n')[1:]).split('\n').__iter__()
```

instead. Otherwise the last line will be ignored. One of the banks I'm working on uses it so I guess there could be more.

pedrocs-exo (changes requested):

PR #230, opened 2019-10-01
[13.0][MIG] account_bank_statement_import_mt940_base

feketemihai:


feketemihai:
To be updated after #229 is merged.

PR #231, opened 2019-10-02
account_bank_statement_import_camt_oca: Add check over multiple accounts/currencies in zipped statements

MaxyMoos:
This PR originates from an issue we've experienced when importing zipped CAMT statements.

When importing zipped CAMTs, the current code checks the Journal's bank account details with the _last statement in the zip archive_'s detected currency/account number.
This behavior makes it possible to import statements unrelated to the Journal's bank account without any errors or warnings being thrown: they just have not to be the last file in the archive.

This PR adds a check over all the currencies/account numbers detected in the zipped CAMT statements and raises an exception if multiple accounts or currencies are detected.


MaxyMoos:
Sorry, I hadn't seen #213, which proposes a fix for this issue as well as several improvements to the module.
This PR is more akin to a band-aid in comparison. I'll let you guys decide if closing this is the best approach.

PR #233, opened 2019-10-30
[10.0] account_bank_statement_import_camt: Fix error in currencies

p-tombez:
In the case of a multi currencies CAMT import, the currency set on the payment line will be the currency of the account but the amount is in the foreign currency. This happens because in the node `TxDtls` it is optional to have the conversion information used during the transaction, hence the currency does not match the one set on the account.
The only workaround I found for this case is to skip the parsing of the `TxDtls` node if the currencies does not match between account and transaction. The payment line will need to be manually processed in the GUI after the import.

### DIFF
--- a/account_bank_statement_import_camt/__manifest__.py
+++ b/account_bank_statement_import_camt/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'CAMT Format Bank Statements Import',
-    'version': '10.0.1.1.6',
+    'version': '10.0.1.1.7',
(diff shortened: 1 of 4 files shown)


grindtildeath (commented):
Looks ok, but adding a test would highlight the issue and avoid regressions...

BTW, shouldn't this be forward ported as well ?

p-tombez:
@grindtildeath I added the case in the tests of transaction grouping as this issue was introduced in this part. But I can make a separate test if you think it's better.
Yes, this will be forward ported as soon as it is approved

Tonow-c2c (approved):
Code review

vrenaville (approved):

PR #234, opened 2019-11-02
[12.0][ADD] account_bank_statement_import_online_paypal

alexey-pelykh:
Adds online bank statements from PayPal

### DIFF
--- a/account_bank_statement_import_online_paypal/__init__.py
+++ b/account_bank_statement_import_online_paypal/__init__.py
@@ -1,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import models
--- a/account_bank_statement_import_online_paypal/__manifest__.py
+++ b/account_bank_statement_import_online_paypal/__manifest__.py
@@ -1,0 +1,23 @@
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2019 Dataplug (https://dataplug.io)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Online Bank Statements: PayPal.com',
+    'version': '12.0.1.0.0',
+    'author':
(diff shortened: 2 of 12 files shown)


alexey-pelykh:
@pedrobaeza potentially, online sync is also interesting for Tecnativa

pedrobaeza:
Well, this is used for OCA instance, so @OSevangelist should say if they are going to use this or the traditional file import.

alexey-pelykh:
@OSevangelist it would be awesome to understand if that's useful for you

artemsenko (approved):
functional tests

sweenya (approved):
Functional test

sweenya (approved):
Functional test

rvalyi:
@alexey-pelykh I'm following your online bank statetement import work with interest. But for this particular Paypal case, I was wondering if no logic could be shared between this new module and the existing one which works from the file (and not via API):
https://github.com/OCA/bank-statement-import/tree/12.0/account_bank_statement_import_paypal
Then may be some refactor is possible instead of duplicating the mapping logic. Frankly I don't know, just asking.

alexey-pelykh:
@rvalyi thanks! I've looked into that module in terms of sharing code and there's even a #236 regarding it. My conclusion is: not worthy, data structures are different

rvalyi:
ok thank you for this analysis and the other PR. I pinged @sebastienbeau who is the person most skilled on these modules at Akretion.

alexey-pelykh:
@OCA/banking-maintainers would it be possible to make further steps about this module?

rafaelbn:
Are you using this PR in production?

alexey-pelykh:
@rafaelbn yep, though we’re using PayPal less often than before. On top of that, PayPal has some glitch on every Monday’s start of day that causes false-positive error message, their engineers are looking into that right now

alexey-pelykh:
The issue is confirmed to be on PayPal side, nothing to be fixed in the module itself, so I'll just merge it.

PR #236, opened 2019-11-11
[12.0][IMP] account_bank_statement_import_paypal

alexey-pelykh:
This PR refactors PayPal reports importing:

- support for both statement and activity reports
- extract fees for transactions with more details
- remove hidden currency conversions, as PayPal currencies should be handled by different journals

CC @pedrobaeza @sbidoul @alexis-via


alexey-pelykh:
@pedrobaeza who could help me out with this? Or there's nobody who's interested in this?

pedrobaeza:
Not sure, as I have only one customer using it and it's working, and the other one is OCA itself, which uses also as is.

alexey-pelykh:
Gotcha, in this case would Tecnativa opt-in for checking if this PR improves usecase of the customer?

pedrobaeza:
Please tell me what this improves and what will be need for adapting old mapping to new one.

alexey-pelykh:
It's pretty much as per description in first message:
- support for both statement and activity reports
- extract fees for transactions with more details
- remove hidden currency conversions, as PayPal currencies should be handled by different journals

yet migration of mapping would be tough - it's very different, and not sure if that's even worthy to do that. But having no migration is bad :(

pedrobaeza:
I'm afraid we are not using that features.

rvalyi:
cc @sebastienbeau

YegorOkhotnikov (approved):
Functional test

sweenya (approved):
Functional test

alexey-pelykh:
Discussed with Pedro, since the migration script is there, should merge

PR #244, opened 2019-12-16
Add line for original name from camt

tschanzt:
Adds a line for the original name from the camt in the information box.
The current one is sometimes omitted and this also allows to check the selected partners name against the original one.


i-vyshnevska (approved):

i-vyshnevska:
@elicoidal can you check proposed PR

TDu (approved):

PR #263, opened 2020-03-26
[12.0][IMP] - Add an option to group payment return at camt statements import

sbejaoui:
When importing bank statements, all payment return transactions are imported individually.

To improve the reconciliation performance, it's better to have an option to group all payment return in a one transaction.

Todo:
- [ ] Add unit tests

### DIFF
--- a/account_bank_statement_import_camt_oca/models/account_journal.py
+++ b/account_bank_statement_import_camt_oca/models/account_journal.py
@@ -1,11 +1,15 @@
 # Copyright 2019 ACSONE SA/NV <thomas.binsfeld@acsone.eu>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import models, _
+from odoo import fields, models, _
 
 
 class AccountJournal(models.Model):
     _inherit = "account.journal"
+
+    group_return_statement_at_camt_import = fields.Boolean(
+        string="Group Return Statements At CAMT Import"
+    )
 
     def _get_bank_statements_available_import_formats(self):
         res = super(AccountJournal, self).\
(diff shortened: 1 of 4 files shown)


sbejaoui:
cc/ @sbidoul , @gva-acsone

account_bank_statement_import_camt_oca/models/parser.py
```diff
@@ -130,6 +130,14 @@ def parse_entry(self, ns, node):
             ],
             transaction, 'ref'
         )
+        if self.env.context.get('journal_id'):
```
sbidoul, line 133:
```suggestion
        if 'journal_id' in self.env.context:
```

is more pythonic maybe

account_bank_statement_import_camt_oca/models/parser.py
```diff
@@ -130,6 +130,14 @@ def parse_entry(self, ns, node):
...
+            journal = self.env['account.journal'].browse(
+                self.env.context.get('journal_id')
+            )
+            if journal.group_return_statement_at_camt_import:
+                if amount < 0:
```
sbidoul, line 138:
```suggestion
```

account_bank_statement_import_camt_oca/models/account_journal.py
```diff
@@ -1,12 +1,16 @@
...
-from odoo import models, _
+from odoo import fields, models, _
 
 
 class AccountJournal(models.Model):
     _inherit = "account.journal"
 
+    group_return_statement_at_camt_import = fields.Boolean(
+        string="Group Return Statements At CAMT Import"
+    )
```
sbidoul, line 12: Should we not name this flag "Ignore CAMT transaction details". help="If checked, import one statement line per CAMT entry else import one statement line per transaction detail".

PR #272, opened 2020-04-11
[12.0][IMP] account_bank_statement_import_online: use pre-computed sanitized bank account number

alexey-pelykh:


alexey-pelykh:
Seems Ocabot is having a weekend as well

PR #273, opened 2020-04-11
[12.0][IMP] account_bank_statement_import_paypal: DE mapping; use account number

alexey-pelykh:


BettinaPfeifer (commented):
Checked the german mapping, seems perfect.

PR #278, opened 2020-04-16
[FIX] account_bank_statement_import_online: use wrong date

fmdl:
@alexey-pelykh

it doesn't make sens to use `date_since` and `date_until`

### DIFF
--- a/account_bank_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_bank_statement_import_online/models/online_bank_statement_provider.py
@@ -204,25 +204,7 @@ class OnlineBankStatementProvider(models.Model):
                     statement_date_since = statement_date_until
                     continue
                 lines_data, statement_values = data
-                statement = AccountBankStatement.search([
-                    ('journal_id', '=', provider.journal_id.id),
-                    ('state', '=', 'open'),
-                    ('date', '=', statement_date),
-                ], limit=1)
-                if not statement:
-                    statement_values.update({
-                        'name': provider.journal_id.sequence_id.with_context(
-                            ir_sequence_date=statement_date,
-                        ).next_by_id(),
-                        'journal_id': provider.journal_id.id,
-                        'date': statement_date,
-                    })
-                    statement = AccountBankStatement.with_context(
-                        journal_id=provider.journal_id.id,
-                    ).create(
-                        # NOTE: This is needed since create() alters values
-                        statement_values.copy()
-                    )
+
                 filtered_lines = []
                 for line_values in lines_data:
                     date = fields.Datetime.from_string(line_values['date'])
@@ -246,7 +228,7 @@ class OnlineBankStatementProvider(models.Model):
                                 )
                             )
                         continue
-                    elif date <= statement_date_since or date > statement_date_until:
+                    elif date <= date_since or date > date_until:
                         continue
                     unique_import_id = line_values.get('unique_import_id')
                     if unique_import_id:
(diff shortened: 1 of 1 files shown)


account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -246,7 +246,7 @@ def _pull(self, date_since, date_until):
                                 )
                             )
                         continue
-                    elif date <= date_since or date > date_until:
+                    elif date <= statement_date_since or date > statement_date_until:
```
alexey-pelykh: These cases are covered by clauses above, and this change breaks the logic due to `<` vs `<=` and `>` vs `>=`

alexey-pelykh (changes requested):

account_bank_statement_import_online/models/online_bank_statement_provider.py
fmdl: in the wizard if you set 01/01 to 01/02, if you have in monthly statement:

`_obtain_statement_data(statement_date_since, statement_date_until)`
it upload transaction 01/01 to 01/02
then it upload transaction 01/01 to **01/03**

if you keep your filter `elif date <= date_since or date > date_until:` all transaction between 02/01 to 28/02 are removed, and you balance are false.

that why it is better to use `elif date <= statement_date_since or date > statement_date_until:`

no ?

alexey-pelykh: If you have monthly statement provider, and use wizard to pull Jan 1 to Feb 1, it will pull
Jan 1 - Jan 31 as one statement and Feb 1 - Feb 1 as another statement, that is by design. As you've requested only until Feb 1st, the monthly/weekly/daily selection affects on how transactions are split between statements.

alexey-pelykh: On top of that, `date < statement_date_since` is handled properly 20 lines above the check that you refer to

fmdl: ok, but the issues is it create an empty statement

alexey-pelykh: It's not an issue, since for requested date statement IS empty

alexey-pelykh: When/if that statement will be having any line, it will become not empty. On top of that, statement date is the beginning of the period that is covered by statement, so it HAS to be created.

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -204,25 +204,7 @@ def _pull(self, date_since, date_until):
                     statement_date_since = statement_date_until
                     continue
                 lines_data, statement_values = data
-                statement = AccountBankStatement.search([
```
alexey-pelykh: This will break logic of appending monthly statements within a month.

alexey-pelykh (changes requested):

account_bank_statement_import_online/models/online_bank_statement_provider.py
fmdl: I have modified the commit, now it create an empty bank statement with the good balance.

alexey-pelykh: What do you mean by good balance? And what is has to do with what we've discussed above?

alexey-pelykh:
Please describe your usecase on what you're trying to achieve. Since you're fixing a by-design logic with no explanation at all.

fmdl:
Get monthly from 01 january to 15 february
- first request : 01 january to 31 janury return 
```
- start balance : 0
- end balance : 100
- transac : [{'name':'transac 1', 'amount':100, 'date':'10 january'}]
```

- seconde request : 01 february to 28 februray return
```
- start balance : 100
- end balance : 150
- transac : [{'name':'transac 1', 'amount':50, 'date': '20 february'}]
```

system create 2 bank statement:
- first : 
```
- start balance : 0
- end balance : 100
- 1 transaction
```

- second
```
- start balance : 100
- end balance : 150
- 0 transaction
```

Why : because the filter to update the balance is not good, it is 1 to 28 febuary : `statement_date_since` and `statement_date_until`. It should be 1 to 15 february

It sould be `date_since` and `date_until`, same filter to include transactions.

alexey-pelykh:
The module should not behave like that because of this block of code: https://github.com/OCA/bank-statement-import/blob/12.0/account_bank_statement_import_online/models/online_bank_statement_provider.py#L239

it subtracts all removed transactions. ~So, either that block has a bug or your information is not correct. But your fix is 100% incorrect.~ On second thought, it may be the case. But still I don't agree to the fix.

fmdl:
user set `date_since = 10 february`, `date_until 20 february`
if it is montly
statement_date is set like this :
`statement_date_since = 1 february`, `statement_date_until = 28 february`

the provider get transaction from `statement_date_since`  to `statement_date_until`
then the system remove transaction from `date_since` to `statement_date_since` and from `statement_date_until` to `date_until`
and it recompute the balance from `statement_date_since` to `statement_date_until`.

It should recompute from `date_since` to `date_until`.

or the provider get transaction from max(`date_since`,`statement_date_since`)  to min(`date_until`, `statement_date_until`).

no ?

alexey-pelykh:
https://github.com/OCA/bank-statement-import/pull/280 should be like this.

alexey-pelykh:
You indeed found a bug, yet it's fix according to your usecase (I've added a test case for that) is different.

PR #279, opened 2020-04-16
[ADD] account_bank_statement_import_online_ponto

fmdl:
@alexey-pelykh 

Ponto is a bank aggregate (For FR, BE, NL banks).


fmdl:
cc @alexis-via

account_bank_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
```diff
@@ -0,0 +1,244 @@
...
+        return self._ponto_obtain_statement_data(date_since, date_until)
+
+    def _get_statement_date(self, date_since, date_until):
+        self.ensure_one()
+        if self.service != 'ponto':
+            return super()._get_statement_date(
+                date_since,
+                date_until,
+            )
+        return date_since.astimezone(pytz.timezone('Europe/Paris')).date()
```
alexey-pelykh: Why do you convert UTC to other timezone? Also, it's better to return datetime to base module for more correct sorting

account_bank_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
```diff
@@ -0,0 +1,244 @@
...
+        self._ponto_synchronisation(account_id)
+
+        transaction_lines = self._ponto_get_transaction(account_id,
+                                                        date_since,
+                                                        date_until)
+
+        new_transactions = []
+        sequence = 0
+        for transaction in transaction_lines:
+            if StatementLine.sudo().search(
```
alexey-pelykh: This filtering is done in base module, and import ID is prefixed always different

account_bank_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
```diff
@@ -0,0 +1,244 @@
...
+                'name': re.sub(' +', ' ', ref) or '/',
+                'ref': attributes.get('remittanceInformation', ''),
+                'unique_import_id': transaction['id'],
+                'amount': attributes['amount'],
+            }
+            new_transactions.append(vals_line)
+
+        if new_transactions:
+            date = max([t['date'] for t in new_transactions])
+            start_balance = Statement.search(
```
alexey-pelykh: Please stop faking statement balance if you don't have one!

alexey-pelykh: Please stop faking statement balance if you don't have one!

alexey-pelykh (changes requested):

account_bank_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
fmdl: If you have an transaction with this datetime : `2020-04-16 - 23:30:00 UTC`, In France UTC+2, the datetime is `2020-04-17 - 1:30:00 UTC+2`, the accounting date is the **17** of april, and not the 16 of april. (Statement work with date and not date_time)

I think there is an issue on your code when you replace hours, minute by zero to find the date, it is not the correct way to find the date. On each provider the tz should be set (when provider send date in UTC format), to convert date_time to date.

Ponto/Qonto are based in Europe/Paris, that why I convert in tz `Europe/Paris` to find the good date.

alexey-pelykh: Correct me if I'm wrong, but you should be using timezone of company, not timezone of provider, since if the provider would be PayPal, doubtfully you'd use the US EST timezone, right?

account_bank_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
fmdl: iI understand, if not you should write by hand the balance for each statement.

alexey-pelykh: That's what Odoo does with manual statement sand that's why it makes difference between `balance_end` and `balance_end_real` - to require used to double check the data unless it's provided. And faking the data is only misleading users that it's correct and verified.

account_bank_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
fmdl: the company have not a `tz` field.

It is more logical to use the timezone of the bank, like that you have the same date between odoo and the bank paper statement (date of bank paper statement == date of accounting).

I think for Paypal it is different, because it is not a bank. And use the timezone of company (keep the user chosse).

alexey-pelykh: When you get statement from Ponto in form of PDF, what timezone it uses?

alexey-pelykh: > date of bank paper statement == date of accounting

I'm not confident about this, yet I'm no expert to claim what's right or wrong in this case. Odoo statement import module always uses the date as-is supplied by the provider, w/o conversions, if I recall correctly. @pedrobaeza maybe you have some deeper knowledge on the matter?

fmdl: Ponto is not a bank, there are no PDF, it agregate data from banks.

On PDF Qonto only the date is written (but by the api you have date and time), Qonto is based in France.

alexey-pelykh: and the PDF in Qonto contains date that corresponds the API response around the midnight or requires conversion to French timezone?

fmdl: For this transaction : "2020-04-15T18:22:30.296Z", I was created at 20h22 (French Timezone).

for exemple in OFX file the timezone is write. https://www.ibm.com/developerworks/xml/library/x-ofxv2/

With cfonb file (french horrible bank file) there is only the date.

pedrobaeza: I don't understand what do you want from me

alexey-pelykh: :D A slice of common sense, as usual! The debate is over this: if provider provides transaction timestamp in UTC full date-time format, what date should be saved in statement? Options:

- UTC date
- Date from a timestamp in assumed timezone of provider (Europe/Paris in this case)
- Your option :)

alexey-pelykh: Frankly speaking, I can find pros and cons for each option so far, because usually API returns in UTC time, but PDF from same service can sometimes be in some fixed timezone or users timezone, or etc. timezone. Maybe it would be even worth adding an option "convert to XXX timezone"

fmdl: It is an accounting question:
Some banks doesn't give a `date` of transaction but a `datetime`, and some time without the timezone (just in UTC).
In Odoo the statement use only the `date`.

Question : how convert the `datetime` to `date` (use the TZ of company, TZ of bank, TZ of user) ?

I think we should use the TZ of bank, because on the PDF bank report there is the date at the bank TZ.

alexey-pelykh: Ultimately, we can give user option to select, since otherwise it's just an assumption, any of options.

alexey-pelykh: I've created #281 to expose timezone configuration to the user in order to avoid such magic within the provider. If the mentioned PR would me merged, provider should return with whatever timezone it has (naive datetime will be treated as UTC)

alexey-pelykh:
Would it be possible to increase test coverage?

patel1008b:
Hello @fmdl and @alexey-pelykh 

Thanks for this PR and functionality is it possible to have it for v11 soon?

Regards,

rafaelbn:
Hello @fmdl , thanks for this great contribution!

Are you working in production with this module? Your customers are already syncing with ponto?

Thank in advance!
Regards, Rafa

fmdl:
I use every days

fmdl:
Yes I have some customer use ponto (sometime, the connection with bank is difficult).

account_bank_statement_import_online_ponto/README.rst
```diff
@@ -0,0 +1,102 @@
...
+#. Save the bank account
+#. Click on provider and configure provider-specific settings.
+
+or, alternatively:
+
+#. Go to *Invoicing > Overview*
+#. Open settings of the corresponding journal account
+#. Switch to *Bank Account* tab
+#. Set *Bank Feeds* to *Online*
+#. Select *Qonto.eu* as online bank statements provider in
```
pedrobaeza:
```suggestion
#. Select *MyPonto.com* as online bank statements provider in
```

account_bank_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
pedrobaeza: I'm getting this error trying to fetch things from Ponto:
```
ValueError: astimezone() cannot be applied to a naive datetime
```
Please remove this conversion

pedrobaeza:
Tried on a live instance and seems to be working well. I can do the needed changes if you want.

fmdl:
@pedrobaeza you can make the needed change.

pedrobaeza:
Solved the tz problem. I have pushed an improved README with fragments, and a new icon.

@alexey-pelykh please change your review.

pedrobaeza (approved):
Tested on my own instance with a Spanish bank.

alexey-pelykh:
@pedrobaeza are you planning to adjust tz code?

alexey-pelykh (approved):

pedrobaeza:
You have already seen my change :wink: 

/ocabot merge nobump

PR #281, opened 2020-04-17
[12.0][IMP] account_bank_statement_import_online: tz

alexey-pelykh:
To accommodate different timezone cases of provider, add specific setting

### DIFF
--- a/account_bank_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_bank_statement_import_online/models/online_bank_statement_provider.py
@@ -2,13 +2,16 @@
 # Copyright 2019-2020 Dataplug (https://dataplug.io)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+from datetime import datetime
 from dateutil.relativedelta import relativedelta, MO
 from decimal import Decimal
 import logging
+from pytz import timezone, utc
 from sys import exc_info
 
 from odoo import models, fields, api, _
 from odoo.addons.base.models.res_bank import sanitize_account_number
+from odoo.addons.base.models.res_partner import _tz_get
 
 _logger = logging.getLogger(__name__)
 
@@ -42,6 +45,15 @@ class OnlineBankStatementProvider(models.Model):
     )
     account_number = fields.Char(
         related='journal_id.bank_account_id.sanitized_acc_number'
+    )
+    tz = fields.Selection(
+        selection=_tz_get,
+        string='Timezone',
+        default=lambda self: self.env.context.get('tz'),
+        help=(
+            'Timezone to convert transaction timestamps to prior being'
+            ' saved into a statement.'
+        ),
     )
     service = fields.Selection(
         selection=lambda self: self._selection_service(),
@@ -156,6 +168,7 @@ class OnlineBankStatementProvider(models.Model):
             )
         AccountBankStatementLine = self.env['account.bank.statement.line']
         for provider in self:
+            provider_tz = timezone(provider.tz) if provider.tz else utc
             statement_date_since = provider._get_statement_date_since(
                 date_since
             )
@@ -225,7 +238,14 @@ class OnlineBankStatementProvider(models.Model):
                     )
                 filtered_lines = []
                 for line_values in lines_data:
-                    date = fields.Datetime.from_string(line_values['date'])
+                    date = line_values['date']
+                    if not isinstance(date, datetime):
+                        date = fields.Datetime.from_string(date)
+
+                    if date.tzinfo is None:
+                        date = date.replace(tzinfo=utc)
+                    date = date.astimezone(utc).replace(tzinfo=None)
+
                     if date < statement_date_since or date < date_since:
                         if 'balance_start' in statement_values:
                             statement_values['balance_start'] = (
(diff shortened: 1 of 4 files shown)


account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -225,7 +238,9 @@ def _pull(self, date_since, date_until):
                     )
                 filtered_lines = []
                 for line_values in lines_data:
-                    date = fields.Datetime.from_string(line_values['date'])
+                    date = line_values['date']
+                    if isinstance(date, str):
```
fmdl: I think don't need here.

but add a check to check that line_values['date'] is a datetime with no tz (like datime used in odoo).

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -246,6 +261,13 @@ def _pull(self, date_since, date_until):
                                 )
                             )
                         continue
+
+                    if isinstance(date, datetime):
```
fmdl: it is the wrong place to do convert. Because you convert dateime UTC to 
you sould do that here : 
https://github.com/OCA/bank-statement-import/blob/12.0/account_bank_statement_import_online/models/online_bank_statement_provider.py#L342
create an other to convert line_values['date'] in date:
```python
def _convert_datetime_in_date(self, date):
     return date.astimezone(self).date()
```
`line_values['date'] = self._convert_datetime_in_date(date)`

update this methode to have the good date in the statement.
```python
    def _get_statement_date(self, date_since, date_until):
        self.ensure_one()
        return date_since.astimezone(provider_tz).date()
```

no ?

I think for better reading of the code, `date_since` and `date_until` could be a `date` and not `datetime`, and `_obtain_statement_data` could be return line with `date` (and not `datetime`).

fmdl (commented):
I think for better reading of the code, date_since and date_until could be a date and not datetime, and _obtain_statement_data could be return line with date (and not datetime).

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: It's done to adapt to some providers that have only date and can send that as string

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: Thanks, regarding statement date it is a valid point - added. As for converting to date prior to returning data - we could do that, yet it's done automatically by Odoo, and just in case there would be a module that changes date field to be datetime, this would make it compatible.

account_bank_statement_import_online/models/online_bank_statement_provider.py
fmdl: you don't need to adapt `from_string` adapt automaitcely `datetime` or `str` and check that there are no timezone.
https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L1741
all date should be pass by `from_string`

alexey-pelykh: That's not for adapting for Odoo, but parsing for processing within the module to allow provider to provide either date or datetime or string.

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -344,6 +368,10 @@ def _get_statement_date(self, date_since, date_until):
         # NOTE: Statement date is treated by Odoo as start of period. Details
         #  - addons/account/models/account_journal_dashboard.py
         #  - def get_line_graph_datas()
+        tz = timezone(self.tz) if self.tz else utc
```
fmdl, line 374: don't need this code, because `statement_date_since` is an datetime without tz (withou tz == utc).
https://github.com/OCA/bank-statement-import/blob/12.0/account_bank_statement_import_online/models/online_bank_statement_provider.py#L199

```python
def _get_statement_date(self, date_since, date_until):
     date_since = date_since.astimezone(provider_tz)
     return date_since.date()
```

alexey-pelykh: Valid point, fixed!

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -225,7 +238,9 @@ def _pull(self, date_since, date_until):
                     )
                 filtered_lines = []
                 for line_values in lines_data:
-                    date = fields.Datetime.from_string(line_values['date'])
+                    date = line_values['date']
+                    if isinstance(date, str):
+                        date = fields.Datetime.from_string(date)
```
fmdl, line 243: if `line_values['date']` is a `date`, you should convert in `datetime` if not after you compare `date` and `datetime`.
`from_string` make this

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -246,6 +261,13 @@ def _pull(self, date_since, date_until):
                                 )
                             )
                         continue
+
+                    if isinstance(date, datetime):
+                        if date.tzinfo is None:
+                            date = date.replace(tzinfo=utc)
+                        date = date.astimezone(provider_tz).replace(tzinfo=None)
+                        line_values['date'] = date
```
fmdl: `line_values['date'] = self._get_statement_date(date, None)`

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: Great point, thanks

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: Confusing optimisation, and defies flexibility of `_get_statement_date` override

alexey-pelykh:
@fmdl please re-review, it should be more appropriate now.

PR #293, opened 2020-05-15
[13.0] MIG account_bank_statement_import_save_file

alexis-via:
The changes in account_bank_statement_import implied a full re-write of the module.


alexis-via:
Problem: with this implementation, it only works when the save_file module is "above" the format-specific files. Because the format-specific files don't call super() when the format of the file is their format.
I'm not sure we can find a solution. Or maybe monkey-patching.

StefanRijnhart:
Can you use the fact that `_create_bank_statements` is called once for each attachment, in the order of self.attachment_ids? You can add a mutable (i.e. a dictionary of list) in the context in an override of `import_file` and use it to encode a counter to keep track of calls to `_create_bank_statements` so that you know which attachment to match with which bank statement.

alexis-via:
@StefanRijnhart I tried to do that (inject info in the context of import_file and use it later in the code), but it failed because what I injected in the context was not present later on. Maybe I made a mistake in the implementation. If you manage to do it, feel free to propose a better solution.
The "good" news is that the module account_bank_statement_import is not part of Odoo Community in v14, so we'll have a full control and we'll be able to have this feature in the base module itself.

PR #301, opened 2020-06-09
[12.0][account_bank_statement_import_txt_xlsx, account_bank_statement_import_paypal] fixes broken views

JordiBForgeFlow:
Adjust wizard after https://github.com/OCA/server-ux/pull/165 has been merged.

cc @alexey-pelykh

### DIFF
--- a/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal_mapping_wizard.py
+++ b/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal_mapping_wizard.py
@@ -164,5 +164,5 @@ class AccountBankStatementImportPayPalMappingWizard(models.TransientModel):
             'res_id': mapping.id,
             'view_mode': 'form',
             'view_id': False,
-            'target': 'new',
+            'target': 'current',
         }
--- a/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal_mapping_wizard.xml
+++ b/account_bank_statement_import_paypal/wizards/account_bank_statement_import_paypal_mapping_wizard.xml
@@ -143,12 +143,8 @@
                     </group>
                 </group>
             </xpath>
-            <xpath expr="//div[@name='final_buttons']/button" position="attributes">
-                <attribute name="class">btn-default</attribute>
-                <attribute name="string">Cancel</attribute>
-            </xpath>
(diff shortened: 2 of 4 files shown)


alexey-pelykh (approved):

alexey-pelykh:
image
@JordiBForgeFlow it seems that it needs a bit more changes

alexey-pelykh (changes requested):
"Import" button then is no longer needed and it's handling may be moved to "Next" button

JordiBForgeFlow:
I'm confused on why do we need to use the multi_step_wizard in those modules, if in this case we just need one step. Am I missing something here?

alexey-pelykh:
@JordiBForgeFlow there are two steps: file import and mapping configuration

JordiBForgeFlow:
@alexey-pelykh ok. Will look into it then.

JordiBForgeFlow:
Ok, I now show the import statement button only in the final step. Does it look right to you now @alexey-pelykh ?

JordiBForgeFlow:
@alexey-pelykh It's a bit weird that after importing the file and creating the mapping the wizard window is not closed. But I'm not sure if this was already happening before, or it's a regression introduced by this PR.

alexey-pelykh:
@JordiBForgeFlow or it's a multi_step_wizard regression, I don't recall such behaviour, it should've been closed, naturally

alexey-pelykh:
Try adding `return {'type': 'ir.actions.act_window_close'}` to `import_mapping` maybe, and please adjust the commit history (or even split the PRs) so that changes would be atomic, I guess

JordiBForgeFlow:
@alexey-pelykh I now have split into two separate commits to clearly distinguish the changes to each module.

JordiBForgeFlow:
@alexey-pelykh instead of doing a window close I am redirecting the user to the new record once the import has happened.

JordiBForgeFlow:
I'm merging this one as currently stable 12.0 is broken.

JordiBForgeFlow:
@alexey-pelykh I guess that you need to approve the changes you requested before I can merge.

alexey-pelykh (approved):

alexey-pelykh:
Ocabot just sleeps probably :) It reacts sometimes hours later

PR #305, opened 2020-06-19
[12.0][account_bank_statement_import_txt_xlsx] fix error parsing field 'original_amount'

JordiBForgeFlow:
At this moment it is not possible to leave the field 'original_amount' excluded from the mapping. It attempts to parse to a decimal a null value.


alexey-pelykh (approved):

alexey-pelykh:
A forward-port to v13 would be much appreciated

JordiBForgeFlow:
Thanks! Will forward-port to 13

PR #306, opened 2020-06-20
[12.0][FIX] account_bank_statement_import_online: correct range filtering

alexey-pelykh:


alexey-pelykh:
cc @fmdl indeed, this was worth more attention

PR #310, opened 2020-06-25
[13.0][IMP]notes and partner_id field mapping.

patel1008b:
Currently, if we set the notes it will not add into the statement lines, Also the same for the partner_id field.

### DIFF
--- a/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
+++ b/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
@@ -351,9 +351,14 @@ class AccountBankStatementImportSheetParser(models.TransientModel):
             note = note.strip()
         if note:
             transaction["note"] = note
+        if notes:
+            transaction['note'] = notes
 
         if partner_name:
             transaction["partner_name"] = partner_name
+            partner_id = self.env["res.partner"].search([("name", "=", partner_name)], limit=1)
+            if partner_id:
+                transaction["partner_id"] = partner_id.id
         if bank_account:
             transaction["account_number"] = bank_account
 


account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
```diff
@@ -351,9 +351,14 @@ def _convert_line_to_transactions(self, line):
...
         if note:
             transaction["note"] = note
+        if notes:
+            transaction['note'] = notes
 
         if partner_name:
             transaction["partner_name"] = partner_name
+            partner_id = self.env["res.partner"].search([("name", "=", partner_name)], limit=1)
+            if partner_id:
+                transaction["partner_id"] = partner_id.id
```
alexey-pelykh, line 361: This is not that base module expects: https://github.com/odoo/odoo/blob/66182a5645980886eb118057c5b0e95e978998c1/addons/account_bank_statement_import/account_bank_statement_import.py#L109

By design, it happens during reconcile:
https://github.com/odoo/odoo/blob/ed47f597bf8cfa18021c66d577c29e541ab4d00d/addons/account/models/reconciliation_widget.py#L118

If you want to resolve partner, you may implement this feature as a separate module, so that it would be available to all other importing modules.

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
```diff
@@ -351,9 +351,14 @@ def _convert_line_to_transactions(self, line):
             note = note.strip()
         if note:
             transaction["note"] = note
+        if notes:
+            transaction['note'] = notes
```
alexey-pelykh, line 355: This is wrong, as you're overwriting previous value. That should have been fixed by replacing
```
note = "{}\n{}".format(note, note.strip())
```
with
```
note = "{}\n{}".format(note, notes.strip())
```

alexey-pelykh (changes requested):

alexey-pelykh:
On top of that, CI fails

patel1008b:
Thanks, @alexey-pelykh for review this PR. I will make it as mention by you. And re-submit new PR.

Issue #312, opened 2020-07-15
UnknownTimeZoneError on paypal PayPal Activity


estekadigital:
I get UnknownTimeZoneError error importing a CSV file downloaded from Paypal. The reason is that Paypal's statement tz_value is CEST, which is not recognized by python3.7, therefor making a small change in the model would solve the issue. I just write the modification here instead of sending a PR as I do not have permission to send them.

Line 269 in file /account_bank_statement_import_paypal/models/account_bank_statement_import_paypal_parser

```
    @api.model
    def _normalize_tz(self, value):
        if value in ['PDT', 'PST']:
            return 'PST8PDT'
        elif value in ['CEST']:
            return 'CET'
        return value

```

alexey-pelykh:
@estekadigital would it be possible to provide an anonymized CSV sample file to be embedded into the tests? And the country from which it originated.

estekadigital:
Here goes a test, I replaced confidential data. The country is Spain.

[paypal_test.csv.txt](https://github.com/OCA/bank-statement-import/files/4930315/paypal_test.csv.txt)

alexey-pelykh:
Thanks! I wonder why Paypal changed the format, as there was a Spanish statement in tests

pedrobaeza:
I don't think the format has changed, but it has been exported from other place.

estekadigital:
I followed the instructions on the readme to download the reports. This test was generated by Paypal from Reports > Activity download menu.

PR #313, opened 2020-07-17
[12.0][FIX] account_bank_statement_import_paypal: adjust TZ mapping

alexey-pelykh:
Addressing #312

### DIFF
--- a/account_bank_statement_import_paypal/models/account_bank_statement_import_paypal_parser.py
+++ b/account_bank_statement_import_paypal/models/account_bank_statement_import_paypal_parser.py
@@ -269,5 +269,7 @@ class AccountBankStatementImportPayPalParser(models.TransientModel):
     @api.model
     def _normalize_tz(self, value):
         if value in ['PDT', 'PST']:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
CEST is not the same as CET. Why do you change it?

alexey-pelykh:
@pedrobaeza because there's no such timezone in tz database:
image

pedrobaeza:
https://stackoverflow.com/questions/10913986/correctly-parse-date-string-with-timezone-information

alexey-pelykh:
image

pedrobaeza:
I don't get you... The solution I saw in the link is to put "Europe/Madrid" for example as tz

alexey-pelykh:
Ah, I didn't get that right. You mean to use Paris or Madrid as a replacement?

pedrobaeza:
Yes, I think it can work.

alexey-pelykh:
Done

pedrobaeza (approved):

PR #315, opened 2020-08-05
[13.0][FIX] account_bank_statement_import_transfer_move: Solved module cras…

CarlosRoca13:
…h when is charged earlier than account_bank_statement_import_camt_oca

@pedrobaeza this is the fix for the crash

### DIFF
--- a/account_bank_statement_import_transfer_move/tests/test_statement.py
+++ b/account_bank_statement_import_transfer_move/tests/test_statement.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import base64
+import unittest
 
 from odoo.modules.module import get_module_resource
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking for avoiding the problem:

/ocabot merge nobump

@i-vyshnevska please take care of not creating such hidden dependencies, and the ideal thing is to not design tests this way.

i-vyshnevska:
@pedrobaeza getting sources from account_bank_statement_import_camt_oca should be a dependency problem no?
even if so then we should copy paste a data to current one addon than just skip the test 
@CarlosRoca13  can you pls send me a link on failed build?

pedrobaeza:
@i-vyshnevska here in OCA and in this branch, you are not going to find the problem, as all modules are installed together. The problem comes when you don't have that module available, or when migrating this module before having the other one available. It doesn't serve to copy the file, as the importer won't be available if the module is not installed.

PR #316, opened 2020-08-06
[12.0] Deeply wrong dependencies specification for `account_bank_statement_import_txt_xlsx`

Byloth:
Hello, @alexey-pelykh.
Hello, everyone.

We're trying to use the module named `account_bank_statement_import_txt_xlsx`.  
It works great when you download it and copy it directly where you need it!

**BUT...**  
Our technological stack includes `pipenv` as package manager for Odoo, its modules and its dependencies.

So...  
When we try to install this module (also using `pip`), it looks for a Python module named `csv` on the package indexes.  
Actually, it would be great if `csv` was a package you can really install but... You know... It is already included in Python's Standard Library and it's deeply wrong to include something that comes by default with the technology you're using.

Here the commit involved: https://github.com/OCA/bank-statement-import/commit/a605bde5f533c55f954eae9199c15343265c010b#diff-1a406bafb14916f5c8fac3ea34cefaf6R24

---

This PR, simply, solves this problem.

### DIFF
--- a/account_bank_statement_import_txt_xlsx/__manifest__.py
+++ b/account_bank_statement_import_txt_xlsx/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Account Bank Statement Import TXT XLSX",
     "summary": "Import TXT/CSV or XLSX files as Bank Statements in Odoo",
-    "version": "12.0.2.0.2",
+    "version": "12.0.2.0.3",
     "category": "Accounting",
     "website": "https://github.com/OCA/bank-statement-import",
     "author":
@@ -21,7 +21,6 @@
     ],
     "external_dependencies": {
         "python": [
-            "csv",
             "xlrd",
         ]
     },


pedrobaeza:
This is not managed that way, but through python_overrides (or similar name) in setup folder.

@sbidoul once again I need your knowledge about all the pip stuff.

account_bank_statement_import_txt_xlsx/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Account Bank Statement Import TXT XLSX",
     "summary": "Import TXT/CSV or XLSX files as Bank Statements in Odoo",
-    "version": "12.0.2.0.2",
+    "version": "12.0.2.0.3",
```
sbidoul, line 8: Normally you don't change the version manually, as the bot does it automatically when merging.

sbidoul (approved):
That's correct :+1:

sbidoul:
/ocabot merge patch

Thanks for the fix.

pedrobaeza:
@sbidoul removing the dependency in the manifest is the correct step?

sbidoul:
@pedrobaeza since `csv` is part of the standard python library there is no need to declare it as an external dependency.

pedrobaeza:
OK then

bodedra (approved):

alexey-pelykh:
I guess this needs to be forward-ported to v13 as well

alexey-pelykh:
@sbidoul this is incorrect as well then https://github.com/OCA/reporting-engine/blob/12.0/report_csv/__manifest__.py

sbidoul:
@alexey-pelykh yes, that one needs fixing too.

Byloth:
@pedrobaeza:
> This is not managed that way, but through python_overrides (or similar name) in setup folder.

Well... Yes... But actually, no! 😅

It's managed in that way only if, of course, you specify the `external_dependencies_override` key in the `setup.py` file...  
Sadly, not in this case: [`account_bank_statement_import_txt_xlsx/setup.py`](https://github.com/OCA/bank-statement-import/blob/12.0/setup/account_bank_statement_import_txt_xlsx/setup.py)

Here, the involved code: [`setuptools_odoo/core.py#L217-L225`](https://github.com/acsone/setuptools-odoo/blob/master/setuptools_odoo/core.py#L217-L225)

---

Thanks to everyone! ❤

pedrobaeza:
Yeah, understood.

account_bank_statement_import_txt_xlsx/__manifest__.py
Byloth: Oh, sorry!  
My bad... 😥

PR #318, opened 2020-08-24
[FIX] account_bank_statement_import_camt_oca: support DtTm and Dt

DonatasV:
* Some banks use DtTm instead of Dt for dates

### DIFF
--- a/account_bank_statement_import_camt_oca/models/parser.py
+++ b/account_bank_statement_import_camt_oca/models/parser.py
@@ -129,7 +129,9 @@ class CamtParser(models.AbstractModel):
     def parse_entry(self, ns, node):
         """Parse an Ntry node and yield transactions"""
         transaction = {"name": "/", "amount": 0}  # fallback defaults
-        self.add_value_from_node(ns, node, "./ns:BookgDt/ns:Dt", transaction, "date")
+        self.add_value_from_node(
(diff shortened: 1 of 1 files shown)


gaikaz:
Could this PR be approved or commented by someone from @OCA/banking-maintainers please :slightly_smiling_face: 
It's a very small non-breaking change.

alexey-pelykh (approved):
LGTM

StefanRijnhart:
I'm afraid those timestamps in XML have a timezone. To convert to a date in an appropriate timezone, you have to apply some logic. One way would be like this:
```
(Pdb) import pytz
(Pdb) from odoo import fields
(Pdb) from dateutil.parser import isoparse  # datetime.datetime.fromisoformat from Python 3.7 on
(Pdb) dt = node.xpath("./ns:BookgDt/ns:DtTm", namespaces={"ns": ns})[0].text
(Pdb) dt
2014-01-04T23:00:00.000Z'  # Timestamp is 4th of Januari 2014, 23:00 hours in UTC
(Pdb) dt_no_tz = isoparse(dt).astimezone(pytz.utc).replace(tzinfo=None)
(Pdb) self.env.user.tz
'Europe/Brussels'
(Pdb) date = fields.Date.context_today(self, timestamp=dt_no_tz)
(Pdb) date
datetime.date(2014, 1, 5)  # Date is 5th of Januari 2014
```

You see that in this example, the date comes out differently from the truncated timestamp which is what the code does now. Can you apply something like this when you encounter the DtTM value?

PR #332, opened 2020-09-30
[12.0][REBRAND] account_bank_statement_import_paypal

alexey-pelykh:


pedrobaeza (changes requested):
You can't add your new copyright to each of the files, as you are not touching them. You can replace old one or only touch the manifest.

PR #334, opened 2020-09-30
[11.0][REBRAND] account_bank_statement_import_online

alexey-pelykh:


account_bank_statement_import_online/README.rst
```diff
@@ -27,6 +27,8 @@ Online Bank Statements
 
 This module provides base for building online bank statements providers.
 
+For support please contact `CorporateHub <mailto:support@corphub.eu>`__.
```
pedrobaeza, line 30: You can't direct to private support, so please undo this, and the same comment about copyright.

alexey-pelykh: Why?

alexey-pelykh: We're maintaining and supporting this module. If we weren't maintaining it - we wouldn't offer the support

pedrobaeza: Because it's not allowed to advertise in OCA.

pedrobaeza: Sorry, but it's not allowed this direct one. You are putting an email as part of the contributors. You can receive private contacts there, but directly state that the support must be done to that email instead of putting an issue here is other thing.

alexey-pelykh: Would phrasing "For support please submit issues at https://github.com/OCA/bank-statement-import and tag @corporatehub" be better? I'm looking for a more straightforward way of maintaining the modules we author.

pedrobaeza: No, you can't simply put that. It's not put in any module.

alexey-pelykh: Why not? It's not "instead of putting an issue here"?

pedrobaeza: It's still advertising. Sorry, but if you want to do that, a debate must be opened in contributors list for adding such option, which is not allowed nowadays.

alexey-pelykh: Fair enough, I'll do the cleanup

PR #361, opened 2020-11-18
Move from 12.0 branch and change version in manifest file

kwoychesko:
The 12.0 branch code works in v11, so shall we pull it into the 11.0 branch? I have only changed the version in the manifest file for this PR.


alexey-pelykh (changes requested):
Please preserve commit history as per https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

PR #363, opened 2020-12-06
[MIG][14.0] account_bank_statement_import_oca_camt54

i-vyshnevska:
depends 
 - [X] https://github.com/OCA/bank-statement-import/pull/362
- [X] https://github.com/OCA/bank-statement-import/pull/367

PR #370, opened 2021-01-17
[14.0][FIX]camt import

luc-demeyer:
The current 14.0 OCA camt parser doesn't work.
This PR fixes a couple of major bugs:
1) Missing payment_ref in the transactions
2) Missing sequences in the transactions

I also moved the 'AddtlNtryInf' to the 'narration' field since this is what Odoo does in the Odoo Enterprise parser.

Remark:
In Odoo Enterprise the narration field also contains the Counterparty Name.
I don't think we need to do the same in the OCA parser since we have the 'partner_name' field correctly populated.
We can consider to make the 'partner_name' field visible in the User Interface, but that should be handled via a separate PR/discussion.


account_statement_import_camt/models/parser.py
```diff
@@ -217,6 +219,8 @@ def parse_statement(self, ns, node):
             result["date"] = sorted(
                 transactions, key=lambda x: x["date"], reverse=True
             )[0]["date"]
+        for seq, vals in enumerate(transactions, start=1):
+            vals["sequence"] = seq
```
StefanRijnhart, line 223: This looks generic. Is this something that should be added to account_bank_import instead, for any imported statements with transactions that do not contain a sequence number?

--edit-- See here: https://github.com/OCA/bank-statement-import/blob/4c82f280adeb7fdef20fd7e0889482c13c2423c8/account_statement_import/wizard/account_statement_import.py#L274

luc-demeyer:
@StefanRijnhart 
I agree that the sequence field seems generic hence can be moved to account_statement_import.
But I prefer we do this via seperate PR because we have a merged but not operational camt parser in 14.0 hence we should fix this asap and add improvements on top of that later.

StefanRijnhart:
But it's the same repo and it's a really simple change. Can you spare the 5 minutes?

luc-demeyer:
@StefanRijnhart 
I moved the sequencing from the camt parser to account_statement_import.
@alexis-via @i-vyshnevska 
Can you also review this PR so that we merge is asap ?

luc-demeyer:
@jgrandguillaume 
Joel,
travis fails on my PR because in the meanwhile you merged https://github.com/OCA/bank-statement-import/pull/363.
But this PR doesn't fly and merging failing modules is not very good for the reputation of the OCA.
The problem is the same as what I fixed via this PR.

If you try to import a camt.054 file via the account_statement_import_oca_camt54 module than it will not work.
Can you revert the merge of https://github.com/OCA/bank-statement-import/pull/363 and reopen it so that @i-vyshnevska can fix it ?

I also think we should not allow mock.patch in unit tests, cf.

    @mock.patch(
        "odoo.addons.account.models.sequence_mixin."
        "SequenceMixin._constrains_date_sequence",
        side_effect=False,
    )

@i-vyshnevska 
I also suggest to rename the module to account_statement_import_camt54 in order to stay consistent with naming of the 14.0 bank statement import modules.

StefanRijnhart:
Sorry about my completely flawed estimation of the change I requested. I'm proposing a superseding PR (https://github.com/OCA/bank-statement-import/pull/374) that is a rebased version of this and which applies the same change to the camt54 module.

About mock.patch, I am not sure what the fundamental problem is using this in the tests of OCA modules but I did encounter errors in this case because `side_effect` was set to `False` instead of `None` (or left out entirely). This change I also applied in the superseding PR. Maybe the error only surfaces depending on the Python version used, not sure.

As for the module name, I think we all agree and in fact it already has the suggested name ;-)

StefanRijnhart:
@luc-demeyer Thanks for clarifying in our private chat that the usage of mock in this particular case masks a problem that actually occurs in real world usage. I've updated my PR.

PR #374, opened 2021-01-31
[14.0][FIX] Adapt camt parsers to data model change (s/name/payment_ref/)

StefanRijnhart:
Continuation of https://github.com/OCA/bank-statement-import/pull/370 by @luc-demeyer 

As of Odoo 14.0, `account.bank.statement.line` has a delegated relation on `account.move`. Each transaction's reference needs to be stored in `payment_ref` now instead of the `name` field, which is reserved for the move sequence number. Setting `name` to the transaction reference would trigger `_constrains_date_sequence` which was disabled in these modules' tests but would still be triggered in real world usage.

This change also setd the sequence number for statement lines of all import formats in bank_statement_import.


luc-demeyer (approved):

luc-demeyer:
Thanks Stefan.
I have tested this latest update with import of a couple of camt files.
This one should be merged asap.

hbrunn (approved):

PR #378, opened 2021-02-26
[14.0][FIX] avoid conflict with enterprise module account_bank_statement_import

fredzamoabg:
Modules OCA **account_statement_import** and ENTERPRISE **account_bank_statement_import** adds a link in **account.journal** that call the same method name **import_statement**. 
When user click any import button, the **import_statement** method of ENTERPRISE module is called.
This PR changes the name of the method in **account_statement_import** module to avoid conflict with ENTERPRISE one.

### DIFF
--- a/account_statement_import/models/account_journal.py
+++ b/account_statement_import/models/account_journal.py
@@ -22,7 +22,7 @@ class AccountJournal(models.Model):
             rslt.append(("file_import", _("Import") + "(" + import_formats_str + ")"))
         return rslt
 
-    def import_statement(self):
+    def import_account_statement(self):
         """return action to import bank/cash statements.
         This button should be called only on journals with type =='bank'"""
         action = (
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I would say that both modules are incompatible (or not compatibility warranty), isn't it, @alexis-via ?

alexis-via:
Yes, compatibility is not warranty, but we tried to make it compatible by changing the name of the module and the name of the wizard. Here, we're talking about the button on account.journal dashboard, so I think it's a good idea to change the name of the method to avoid conflict.

alexis-via (approved):
I would prefer to rename the method "import_account_statement".

pedrobaeza:
Yes, please rename it and let's avoid this one, but I remember you told about a blocking point, Alexis. If not, good to have choices.

fredzamoabg:
> I would prefer to rename the method "import_account_statement".

done

pedrobaeza (approved):

alexis-via:
Thanks for your work on this PR!

PR #384, opened 2021-03-30
[13.0][MIG][IMP] adyen statement import

NL66278:
Migration of account_bank_statement_import_adyen, but extended with possibility to import csv files, and with a module to automatically import them form the Adyen website.

This PR also contains some changes to other modules, that were made necessary by the unfortunate logic of the import modules to try the import in turns and then call super when failing. Not all modules where handling this in a proper way.

It would be better if, like the original modules in 7.0, an import was always linked to a certain format, so this dubious try..catch..pass on logic could be avoided.


StefanRijnhart:
Looks great, you've solved all the technical debt from the 12.0 version and greatly enhanced the functionality! Are you suggesting we keep all the individual commits or can some of them be squashed?
Just curious: if you don't enable the clearing account mechanism for your adyen imports, what does your accounting setup looks like? In our setup, I think the amounts are transferred by batch so each statement represents one actual bank transaction, corresponding to the outstanding amount on the clearing account. In any case, please let me know what your considerations are not to use the clearing account mechanism.

NL66278:
@StefanRijnhart with the customer that needed the migration, it is not yet clear how they want to handle the downloaded statements. The clearing account is still in the migration. I just removed the dependency from the adyen import. But there is no obstacle to use the modules in combination.

I will squash the commits where logical.

StefanRijnhart:
Mmm... In our case, we implemented the import module first, then when testing it turned out that the financial logic required this clearing account mechanism. So I'm not sure removing the dependency is a good idea.

NL66278:
@StefanRijnhart Zou je kunnen leven met een dringend advies in de README om ook de clearing_account module te installeren ipv een technische dependency?

StefanRijnhart:
I'm curious if the Adyen module is useful without the clearing account mechanism, and if we don't know, then what is the reason that you decoupled those?

NL66278:
@StefanRijnhart I discussed it and will put back the clearing_account dependency. I will also go through all the commits once more to minimize them even more and rectify a few oversights.

lfreeke:
@NL66278 I think this should be an OCA icon -> account_bank_statement_import_online_adyen/static/description/icon.png

NL66278:
@lfreeke I did not change the icon, do you expect problems with this?

lfreeke:
> 


@NL66278  It has the Ponto icon now.

NL66278:
@lfreeke I replaced the Ponto logo with the Adyen logo

NL66278:
Runboat sunk?

+ bash /runboat/runboat-clone-and-install.sh
+ rm -fr /mnt/data/initialized
+ rm -fr /mnt/data/odoo-addons-dir
+ mkdir -p /mnt/data/odoo-addons-dir
+ cd /mnt/data/odoo-addons-dir
+ curl -sSL https://github.com/oca/bank-statement-import/tarball/5885781ce65e877ce51e8bb83f095d73b517abd8
+ tar zxf - --strip-components=1
+ INSTALL_METHOD=oca_install_addons
+ [[ oca_install_addons == \o\c\a\_\i\n\s\t\a\l\l\_\a\d\d\o\n\s ]]
+ oca_install_addons
+ setuptools-odoo-get-requirements --include-addons --addons-dir /mnt/data/odoo-addons-dir
/root/.local/pipx/shared/lib/python3.8/site-packages/_distutils_hack/__init__.py:17: UserWarning: Distutils was imported before Setuptools, but importing Setuptools also replaces the `distutils` module in `sys.modules`. This may lead to undesirable behaviors or errors. To avoid these issues, avoid using distutils directly, ensure that setuptools is installed in the traditional way (e.g. not an editable install), and/or make sure that setuptools is always imported before distutils.
  warnings.warn(
/root/.local/pipx/shared/lib/python3.8/site-packages/_distutils_hack/__init__.py:30: UserWarning: Setuptools is replacing distutils.
  warnings.warn("Setuptools is replacing distutils.")
+ cat test-requirements.txt
csv
odoo13-addon-multi_step_wizard
odoo13-addon-web_widget_dropdown_dynamic
odoo>=13.0a,<13.1dev
xlrd
+ pip install -r test-requirements.txt
Looking in indexes: https://wheelhouse.odoo-community.org/oca-simple-and-pypi
ERROR: Could not find a version that satisfies the requirement csv (from versions: none)
ERROR: No matching distribution found for csv

lfreeke (commented):
:+1:  Functional test, works well for our customer.

NL66278:
@StefanRijnhart Could you re-open, and possibly merge this one? Working  very well for our customer for a long time now. And I will be migrating it to 16.0.

StefanRijnhart (approved):
Great work, thanks!

StefanRijnhart:
Please rebase and try to fix linting: https://github.com/OCA/bank-statement-import/actions/runs/6296116373/job/17090503942

NL66278:
@StefanRijnhart Thanks for your actions and comments. I did the rebase and the linting problems should be fixed now.

StefanRijnhart:
Alright!

/ocabot merge nobump

PR #386, opened 2021-05-04
[13.0][IMP] account_bank_statement_import_camt_oca: support DtTm

naglis:
Related PR: #318, see comment: https://github.com/OCA/bank-statement-import/pull/318#issuecomment-769912488

### DIFF
--- a/account_bank_statement_import_camt_oca/models/parser.py
+++ b/account_bank_statement_import_camt_oca/models/parser.py
@@ -4,9 +4,11 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 import re
 
+import pytz
+from dateutil.parser import isoparse
 from lxml import etree
 
-from odoo import models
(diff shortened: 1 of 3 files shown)


gaikaz (approved):
LGTM
Tested functionally

@alexey-pelykh You approved the other PR, this is an improved one, could you please check it out as well?

alexey-pelykh (approved):
Code LGTM

gaikaz:
It just so happens that today one of our clients got the same issue this PR fixes on v12.
I checked if this fix could work on v12 and it does, but because `isoparse` is used, it requires dateutil of version `2.7.0+`, but on Odoo 12.0 `requirements.txt` it was only `2.5.3`, so no `isoparse` to use. 
@naglis @StefanRijnhart Any ideas how we could back-port this fix to v12?
I guess just adding `2.7.3` version of dateutil to `external_depends` wouldn't be acceptable?

PR #394, opened 2021-06-09
[12.0][IMP] account_bank_statement_import_txt_xlsx: Skip initial rows

MateuGForgeFlow:
Adds a field that allows to introduce an integer. This integer will be the number of initial rows that will be skipped in the imported file.
cc @LoisRForgeFlow

### DIFF
--- a/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
+++ b/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
@@ -13,9 +13,9 @@ class AccountBankStatementImportSheetMapping(models.Model):
         required=True,
     )
 
-    delete_rows = fields.Integer(
+    skipped_rows = fields.Integer(
         string="Ignore First Rows",
-        help="The import will ignore the specified initial rows amount",
+        help="The number of lines to skip at the beginning of the file.",
         default=0,
     )
 
--- a/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
+++ b/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
@@ -12,6 +12,13 @@ class AccountBankStatementImportSheetMapping(models.Model):
     name = fields.Char(
         required=True,
     )
+
+    delete_rows = fields.Integer(
+        string="Ignore First Rows",
+        help="The import will ignore the specified initial rows amount",
+        default=0,
+    )
+
     float_thousands_sep = fields.Selection(
         string='Thousands Separator',
         selection=[
--- a/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
+++ b/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
@@ -56,8 +56,8 @@ class AccountBankStatementImportSheetParser(models.TransientModel):
             journal.code,
             path.basename(filename),
         )
-        if mapping.delete_rows > 0 and name.lower().endswith('.csv'):
-            for i in range(mapping.delete_rows):
+        if mapping.skipped_rows > 0 and name.lower().endswith('.csv'):
+            for i in range(mapping.skipped_rows):
                 aux = data_file.decode("utf-8")
                 index = aux.find("\n")
                 data_file = aux[index + 1:].encode('utf-8')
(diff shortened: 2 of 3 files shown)


account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
```diff
@@ -56,6 +56,12 @@ def parse(self, mapping, data_file, filename):
             journal.code,
             path.basename(filename),
         )
+        if mapping.delete_rows > 0 and name.lower().endswith('.csv'):
```
alexey-pelykh: What's wrong with skipping XLS rows?

LoisRForgeFlow: agree, it is better to skip if possible. Was there any impediment to skip?

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
```diff
@@ -12,6 +12,13 @@ class AccountBankStatementImportSheetMapping(models.Model):
     name = fields.Char(
         required=True,
     )
+
+    delete_rows = fields.Integer(
```
LoisRForgeFlow: actually, I'm seeing now it is only a naming issue...
image

Can you change this field no somthing more accurate: `skipped_rows` or similar

MateuGForgeFlow: Done! Thank you @alexey-pelykh and @LoisRForgeFlow.

LoisRForgeFlow (approved):
LGTM

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
```diff
@@ -12,6 +12,13 @@ class AccountBankStatementImportSheetMapping(models.Model):
     name = fields.Char(
         required=True,
     )
+
+    skipped_rows = fields.Integer(
+        string="Ignore First Rows",
+        help="The import will ignore the specified initial rows amount",
```
alexey-pelykh: > The number of lines to skip at the beginning of the file.

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
```diff
@@ -12,6 +12,13 @@ class AccountBankStatementImportSheetMapping(models.Model):
     name = fields.Char(
         required=True,
     )
+
+    skipped_rows = fields.Integer(
```
alexey-pelykh, line 16: `rows_to_skip_start`? What if someone would add _end later

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
```diff
@@ -56,6 +56,12 @@ def parse(self, mapping, data_file, filename):
             journal.code,
             path.basename(filename),
         )
+        if mapping.skipped_rows > 0 and name.lower().endswith('.csv'):
```
alexey-pelykh, line 59: If you skip rows in `_parse_lines` why this block is needed?

alexey-pelykh (changes requested):
And travis fails

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
MateuGForgeFlow: Hi @alexey-pelykh, in _parse_lines() the xls lines are skipped, here is the code for the .csv

MateuGForgeFlow:
Hi @alexey-pelykh! Travis Fixed, thank you for the review!

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
MateuGForgeFlow: Done!

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_mapping.py
chienandalu: I'd go for a more precise `header_offset`

account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py
chienandalu: This is not the best way to detect a csv, which in the end is just a plain text file. Why don't you just skip the proper offset in this `else`: https://github.com/OCA/bank-statement-import/blob/2e6a37f91e21d0d81b0062f7726bd6f427271f31/account_bank_statement_import_txt_xlsx/models/account_bank_statement_import_sheet_parser.py#L123-L128

Something like:
```python
     for i in range(mapping.skipped_rows):
         next(csv_or_xlsx)
     header = [value.strip() for value in next(csv_or_xlsx)]
```

chienandalu (commented):
This would be a cool feature

Some comments though (also some tests would be nice)

PR #399, opened 2021-07-14
[13.0][FIX] account_bank_statement_import_paypal: Parser bug for required fields

CarlosRoca13:
The fields required for the paypal parser are set as None and don't throw the
ValueError. With this changes we achieve to filter the data and throw the exception
when the fields are required.

cc @Tecnativa TT29944

please @joao-p-marques @pedrobaeza review this

### DIFF
--- a/account_bank_statement_import_paypal/models/account_bank_statement_import_paypal_parser.py
+++ b/account_bank_statement_import_paypal/models/account_bank_statement_import_paypal_parser.py
@@ -72,7 +72,7 @@ class AccountBankStatementImportPayPalParser(models.TransientModel):
         )
 
     def _data_dict_constructor(self, mapping, header):
-        list_of_content = [
+        required_list = [
             "date_column",
             "time_column",
             "tz_column",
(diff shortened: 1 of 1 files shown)


joao-p-marques (approved):
:+1: 

This is specially important when dealing with other parser modules, where this one should fail and delegate the parsing to another modules in the chain when it encounters something it can0t process.

pedrobaeza (commented):
I think it's more readable to do 2 lists: one required and one optional, and put the code according that.

CarlosRoca13:
@pedrobaeza changed

account_bank_statement_import_paypal/models/account_bank_statement_import_paypal_parser.py
```diff
@@ -93,7 +95,9 @@ def _data_dict_constructor(self, mapping, header):
             "bank_account_column",
         ]
         data_dict = {}
-        for key in list_of_content:
+        for key in required_list:
+            data_dict[key] = header.index(getattr(mapping, key))
```
pedrobaeza, line 99: What about the raise if not found?

joao-p-marques: AFAICS, if the element is not found, the `index()` method will raise the exception, which won't be catched, so it is what we want...

pedrobaeza: OK

pedrobaeza (approved):

PR #295, opened 2020-05-15
[13.0][MIG] account_bank_statement_import_ofx

alexis-via:

PR #403, opened 2021-08-07
[13.0] pre-commit sanitize of #295

rpsjr:
#295 account_bank_statement_import_ofx read to merge


rvalyi:
Hello @rpsjr thanks for trying to help. However there is an OCA migration process to enforce (mostly to preserve the module history so one can track bugs and fixes across Odoo series/branches) that you can find here https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

So basically you could either do:

1.  checkout the branch https://github.com/OCA/bank-statement-import/pull/295 and add an extra commit to pass the pre-commit test and submit that new branch in a new PR like here (but with all the proper git history then). In the case the previous commit was trivial, you could amend it (commit with --amend).
2. send us a PR to our branch at Akretion against https://github.com/akretion/bank-statement-import/tree/13-mig-ofx . In the case of somebody working actively on the PR you are trying to help/fix that would be the best. But in this case, since Akretion works on v12 and v14 and not v13, it is true we would likely not be very reactive. So solution 1) would be better. 

But submitting a migrated module in a new branch in a single big commit is not the way to go sorry.

Did I cleared your doubts?

rpsjr:
Got it, thx @rvalyi

PR #407, opened 2021-08-19
[FIX] account_statement_import_online_ponto: fix vals dict to create a bank statement

ernestotejeda:
Cc @Tecnativa TT31034

- Wrong key in the vals dict passed to create method of bank statements
- Improve tests


pedrobaeza (approved):

pedrobaeza:
Let's fast-track it as this is not working without the fix:

/ocabot merge patch

PR #408, opened 2021-08-19
[14.0][IMP] account_statement_import_online: don't import empty statements.

ernestotejeda:
Cc @Tecnativa TT31034

Sometimes and depending on how the 'Statement Creation Mode' field is configured in the provider model, empty bank statements will be created. This PR is to avoid that.

Main changes:
- Avoid create empty statements.
- Add unit tests.

### DIFF
--- a/account_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_statement_import_online/models/online_bank_statement_provider.py
@@ -95,6 +95,7 @@ class OnlineBankStatementProvider(models.Model):
     certificate_public_key = fields.Text()
     certificate_private_key = fields.Text()
     certificate_chain = fields.Text()
+    allow_empty_statements = fields.Boolean(string="Allow empty statements")
 
     _sql_constraints = [
         (
@@ -204,6 +205,8 @@ class OnlineBankStatementProvider(models.Model):
             )
         if not data:
             data = ([], {})
+        if not data[0] and not data[1] and not self.allow_empty_statements:
+            return
         lines_data, statement_values = data
         if not lines_data:
             lines_data = []
--- a/account_statement_import_online/README.rst
+++ b/account_statement_import_online/README.rst
@@ -56,6 +56,10 @@ or, alternatively:
 #. Save the bank account
 #. Click on provider and configure provider-specific settings.
 
+If you want to allow empty bank statements to be created every time the
+information is pulled, you can check the option "Allow empty statements"
+at the provider configuration level.
+
 **NOTE**: To access these features, user needs to belong to
 *Show Full Accounting Features* group.
 
--- a/account_statement_import_online/readme/CONFIGURE.rst
+++ b/account_statement_import_online/readme/CONFIGURE.rst
@@ -19,5 +19,9 @@ or, alternatively:
 #. Save the bank account
 #. Click on provider and configure provider-specific settings.
 
+If you want to allow empty bank statements to be created every time the
+information is pulled, you can check the option "Allow empty statements"
+at the provider configuration level.
+
 **NOTE**: To access these features, user needs to belong to
 *Show Full Accounting Features* group.
--- a/account_statement_import_online/static/description/index.html
+++ b/account_statement_import_online/static/description/index.html
@@ -406,6 +406,9 @@ section</li>
 <li>Save the bank account</li>
 <li>Click on provider and configure provider-specific settings.</li>
 </ol>
+<p>If you want to allow empty bank statements to be created every time the
+information is pulled, you can check the option “Allow empty statements”
+at the provider configuration level.</p>
 <p><strong>NOTE</strong>: To access these features, user needs to belong to
 <em>Show Full Accounting Features</em> group.</p>
 </div>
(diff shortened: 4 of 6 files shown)


alexey-pelykh (changes requested):
Thanks for the contribution! It was intended behaviour actually. Since it's a breaking change, please make that optional and off by default. E.g. "Create empty statements" with default "on"

pedrobaeza:
I don't get why this can be a feature. It's totally distracting and adding nothing.

alexey-pelykh:
@pedrobaeza Because that way it clearly and explicitly shows which weeks/days/months had no turnovers. If for your usecases it's unwanted - make it optional and disable it.

pedrobaeza:
That's a nonsense, and more when you have monthly statement. You are not maintaining this module in this version, so you should be the one that if you want such behavior, propose it as an option. For us in v14 this shouldn't be here.

alexey-pelykh:
@pedrobaeza may I suggest that it's your _opinion_ that it is nonsense not a fact? Altering behaviour in favor of one group of users vs another w/o providing "the others" way to continue using as before is bad. It's a breaking change, w/o any chance of migration. Thus should be optional and explicitly enabled, and in v15 it could make sense making it default, making both groups of users happy with proper migration script.

pedrobaeza:
The question is that you are the only one wanting that behavior, and it's not favoring anything with it. You have introduced other breaking changes in the past (like Paypal one, that I suffered, as migration scripts wasn't enough) without changing version, and v14 module has been just merged. Don't twist rules on your benefit.

alexey-pelykh:
> you are the only one wanting that behavior

I would assume that you have data to make such a statement? Since it seems not to count in my deployments.

> You have introduced other breaking changes in the past (like Paypal one, that I suffered, as migration scripts wasn't enough) without changing version

I did made those breaking changes:

- it was marked as refactor and improvement
- there was a huge banner about that in README
- at least I did what I could to avoid bad experience for users going through upgrade
- that refactor brought in _extended_ features that were impossible to introduce w/o refactor that unfortunately caused breaking changes

That case is hardly comparable to current one:

- tweaking behaviour to a specific liking (since you favour your "many" over my "one", which is still your estimation)
- not introducing warning/note about that (makes that sneaky breaking change)
- no real reasoning of the breaking change (e.g. features that need that) except for opinion

> just merged

Not really, 3 weeks ago.

> Don't twist rules on your benefit.

Which rules do I twist in this case?

pedrobaeza:
None of my customers want empty statements.

OK, we are going to add such option at config level (Don't create empty statements), but enabled by default.

alexey-pelykh:
That sounds like an awesome trade-off! It allows all users regardless of their views to happily use the module the way they like it with a tick of a checkbox.

ernestotejeda:
@pedrobaeza @alexey-pelykh changes done, check it out please :)

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -95,6 +95,10 @@ class OnlineBankStatementProvider(models.Model):
     certificate_public_key = fields.Text()
     certificate_private_key = fields.Text()
     certificate_chain = fields.Text()
+    dont_create_empty_statements = fields.Boolean(
+        string="Don't create empty statements",
+        default=True,
```
alexey-pelykh: Default should be false, OR it should be `allow_empty_statements` - to avoid sneaky change.

pedrobaeza: No, we want this default. You can change it in your side.

alexey-pelykh: What about proper mentioning in readme? So that users would know about the sneaky change?

alexey-pelykh: Also, wouldn't it be more readable to do `allow_empty_statements` with  default false then?

alexey-pelykh: I'd really expect a major version version bump with this also, since breaking change, in addition to explaining in README.

alexey-pelykh: > No, we want this default. You can change it in your side.

Ok

ernestotejeda:
@pedrobaeza @alexey-pelykh changes done

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -204,6 +205,8 @@ def _create_or_update_statement(
             )
         if not data:
             data = ([], {})
+        if (not data[0]) and (not data[1]) and (not self.allow_empty_statements):
```
alexey-pelykh: Not sure if `(` and `)` are needed

alexey-pelykh (approved):

ernestotejeda:
you are right, according to this https://docs.python.org/3.6/reference/expressions.html#operator-precedence it is not necessary, I'm going to change it. Thanks.

ernestotejeda:
ping @pedrobaeza @alexey-pelykh

pedrobaeza (approved):

PR #413, opened 2021-09-17
[12.0][IMP] account_bank_statement_import_online_transferwise: support SCA

alexey-pelykh:


JordiBForgeFlow:
@FerranCforgeFlow can you review?

FerranCforgeFlow:
The URL modifications and the Secure Customer Authentication look good to me. However, I do not actually see the utility of the _widget_dynamic_dropdown_ in the _transferwise_profile_ field.  Shouldn't it be better to allow the user to freely write the ProfileID?

alexey-pelykh:
@FerranCforgeFlow 

> I do not actually see the utility of the widget_dynamic_dropdown in the transferwise_profile field

Being user-friendly, there's no way of getting that ID from TransferWise UI, known to me. Originally, this module was the sole reason why `widget_dynamic_dropdown` was created.

> Shouldn't it be better to allow the user to freely write the ProfileID?

No benefit of that, since API allows access only to profiles it returns via API call. And module lists those in a convenient way via the widget.

FerranCforgeFlow:
I have now understood the actual working of the widget. Thanks! For the rest, it LGTM 👍

alexey-pelykh:
@FerranCforgeFlow thanks! Don't forget to 👍 the PR itself :)

FerranCforgeFlow (approved):

PR #415, opened 2021-09-23
[13.0][MIG] account_bank_statement_import_online_paypal

alexey-pelykh:


hbrunn (approved):
if you agree with https://github.com/OCA/bank-statement-import/pull/418, you might want to add this here too

alexey-pelykh:
Thanks for the review! Not sure I understand it enough to agree or disagree, since I'm using this module a lot and haven't had issues, so our use-case must be different. Once we'll sort out that one positively, I'll just do an extra migration

PR #417, opened 2021-09-27
[14.0] [MIG] account_bank_statement_import_txt_xlsx → account_statement_import_txt_xlsx

primes2h:


account_statement_import_txt_xlsx/README.rst
```diff
@@ -1,6 +1,6 @@
-==========================================
-Account Bank Statement Import TXT/CSV/XLSX
-==========================================
+==================================
+Bank Statement TXT/CSV/XLSX Import
```
TheMule71, line 2: Account Statement TXT/CSV/XLSX Import

Is keeping Bank and removing Account the intented result?

TheMule71:
Pls see https://github.com/TheMule71/bank-statement-import/commit/9bfb79dd9279cf58d2314d311a6a0e647baf8077

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
```diff
@@ -0,0 +1,378 @@
...
+        if bank_account:
+            note += _("Account: %s; ") % (bank_account,)
+        if transaction_id:
+            note += _("Transaction ID: %s; ") % (transaction_id,)
+        if note and notes:
+            note = "{}\n{}".format(note, note.strip())
+        elif note:
+            note = note.strip()
+        if note:
+            transaction["payment_ref"] = note
```
TheMule71: replace if with:
```transaction["payment_ref"] = reference or description or note or transaction.get("unique_import_id") or "/"```

"payment_ref" must be present. The order of choice is open for discussion. In practical examples I've seen, often there's only a description line.  Which one to choose if note, reference and description isn't an obvious pick.

Here's the rationale for my choice: let the user pick from actual data first, then provide generated data (on which the user has less control) as fallback.

**Comment by TheMule71** [comment]
Pls see https://github.com/PNLUG/bank-statement-import/pull/4

**Comment by primes2h** [comment]
> Pls see [PNLUG#4](https://github.com/PNLUG/bank-statement-import/pull/4)

Merged, thanks!

**Comment by primes2h** [comment]
> Pls see [TheMule71@9bfb79d](https://github.com/TheMule71/bank-statement-import/commit/9bfb79dd9279cf58d2314d311a6a0e647baf8077)

Thanks, done.

File: account_statement_import_txt_xlsx/README.rst (hunk 1 above)
Comment (primes2h) [reply account_statement_import_txt_xlsx/README.rst:2]: Yes.

File: account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py (hunk 2 above)
Comment (primes2h) [reply account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py]: Thanks, merged with https://github.com/PNLUG/bank-statement-import/pull/4.

**Comment by TheMule71** [comment]
One possible fix for the name vs sequence problem we experienced.
https://github.com/PNLUG/bank-statement-import/pull/5

I'm not entirely sure this is the best approach. I'm going to try something different.

A technical analysis of what's going one here (since @primes2h and I discussed this in person):

At some point, account.bank.statement.line objects are created, whose name contains a number, and that triggers a check (inside the sequence.mixin) of name vs date, which fails (even a name like "entry n.20" fails, since 20 is interpreted as 2020 and that doesn't match the date 2021).  See the test I've added for details.

In order to circumvent that, I simply remove the 'name' key from the transaction dictionary, which creates unnamed "/" entries.

**Comment by ernestotejeda** [comment]
Hello @primes2h , do you plan to continue with this PR? if you can't, I would like to continue in a new PR, what do you say? . Thanks.

**Comment by primes2h** [comment]
> Hello @primes2h , do you plan to continue with this PR? if you can't, I would like to continue in a new PR, what do you say? . Thanks.

Hi @ernestotejeda 
PR is ready now. Feel free to review it, thanks. :wink: 

P.S. [This](https://github.com/OCA/bank-statement-import/pull/417#issuecomment-948917958) particular issue is not directly related to this migration, it probably occurs in v.13 too.

File: account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py [hunk 3]
```diff
@@ -0,0 +1,389 @@
...
+            if currency:
+                transaction["currency_id"] = currency.id
+
+        if transaction_id:
+            transaction["unique_import_id"] = "{}-{}".format(
+                transaction_id,
+                int(timestamp.timestamp()),
+            )
+
+        transaction["name"] = description or _("N/A")
```
ernestotejeda:
```suggestion
        transaction["payment_ref"] = description or _("N/A")
```
Now in v14 the model `account.bank.statement.line` inherits from `account.move` (composition-based inheritance) so the name field is the same as defined in the `account.move` model. In other words, it will be the number of the related journal entry, so we shouldn't change it. However, the column that is now displayed with the name 'Label' is the one corresponding to the `payment_ref` field.

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
```diff
@@ -0,0 +1,389 @@
...
+            note = "{}\n{}".format(note, note.strip())
+        elif note:
+            note = note.strip()
+        transaction["payment_ref"] = (
+            note
+            or reference
+            or description
+            or transaction.get("unique_import_id")
+            or "/"
+        )
```
ernestotejeda:
```suggestion
        if note:
            transaction["narration"] = note
```
I think now in v14 `narration` field is the substitute for `note` field, so I think we should only change `note` to `narration` and leave everything as it was before, what do you think?

ernestotejeda (changes requested):
@primes2h thanks for the migration. I have some suggestions, @TheMule71 give me your opinion about them, thank you.

ernestotejeda:
ping @pedrobaeza @carlosdauden

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
TheMule71: I maintain that we should try and populate `payment_ref` to the best of our possibilities. Contrary to other modules, we don't have control over the data we are provided. `description` may, or may not, be populated (that depends on the map).

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
TheMule71: Well in my tests I've found that the first `account.bank.statement.line` does get a meaningful name/number like BNK1/2021/.../..., the following ones don't. If the name field isn't displayed in the most used views, I agree to leave it alone, even if I'd like the records created to have proper names.

Anyway, I think we agree here. I decided to simply not pass the name at all:
```      # avoid errors due to sequence checking
        del transaction["name"]
```
since transaction is a dictionary that is (cutting a long story short) used as values in a create().  

The final point (where the sequence checking happens and the error is triggered) is here:

https://github.com/OCA/OCB/blob/c0a79cdda56bf0aa4ee6741dce7eaff0179f857b/addons/account/models/account_bank_statement.py#L859

I've tried to simply disable the sequence mechanism, by overriding the mixin methods in `account.bank.statement.line`, that doesn't seem to work, in that call is seems the methods in `account.move` are called directly (I've traced it). I've stopped investigating further because I'm not sure if  `account.bank.statement.line` is inheriting from `account.move` just for convenience (to avoid code repetition) or those records are actually and fully intended to be `account.move`'s (and I mean by standard accounting practice).  To put it differently, if it was a python programmer's idea or an accountant's idea.  If the second is true, and they are supposed to be real `account.move`'s, probably they _should_ have proper sequence numbers as names and disabling the checks isn't the appropriate solution.


All this to say I like your approach more. In short:
- don't create the key `name` in the first place, at all
- use `payment_ref` instead (and put just the `description_column` in it)
- use `narration` for notes


BTW the "long story" I've mentioned before goes like this:

the _parse_file() method in `account.statement.import` (which is an abstract method you should override is sub-modules) returns a whose data structures, in particulare the transactions list is what we're interested in:

https://github.com/OCA/bank-statement-import/blob/98618c06117f9f6ebdecd361e741897d79393523/account_statement_import/wizard/account_statement_import.py#L135-L163
note that the comment mentions `payment_ref` and not `name`.

The transactions undergo to some checking and sanitation, the finally they are turned into `line_ids` here:
https://github.com/OCA/bank-statement-import/blob/98618c06117f9f6ebdecd361e741897d79393523/account_statement_import/wizard/account_statement_import.py#L354-L358

So my assessment was, we either provide meaningful names or don't even have the key in our dictionaries. I was unable to find a way to provide a meaningful name, so I opted for the second route, and I decided to remove the key at the end. Like I said, I agree it's best not to create it in the first place.

ernestotejeda: Thanks to answer @TheMule71. I really do not know if taking the lines of the bank statements as journal entries is the best idea or not, but we have no choice to deal with it, hehe.

Anyway `del transaction["name"]` throws an error, because the key 'name' is not in the dictionay `transaction`, that's why I just propused to change `transaction["name"]` to `transaction["payment_ref"]`.

ping @primes2h

ernestotejeda (approved):
LGTM :+1:

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
primes2h: > Now in v14 the model `account.bank.statement.line` inherits from `account.move` (composition-based inheritance) so the name field is the same as defined in the `account.move` model. In other words, it will be the number of the related journal entry, so we shouldn't change it. However, the column that is now displayed with the name 'Label' is the one corresponding to the `payment_ref` field.

Done, thanks!

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
primes2h: > I think now in v14 `narration` field is the substitute for `note` field, so I think we should only change `note` to `narration` and leave everything as it was before, what do you think?

You are right, fixed.
Thanks for your review!

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
primes2h: @TheMule71
Name/Sequence is not an issue anymore. 
Do you think it still worth adding [this test](https://github.com/PNLUG/bank-statement-import/pull/5/commits/a0fb268a99f9dd04a904553d968e7d4b663b9443#diff-3c2f9c9f9192ba5e4d53d38654893c14ae6596292af17e96967742a2b9f8a2df)? Otherwise https://github.com/PNLUG/bank-statement-import/pull/5/ can be closed.

ernestotejeda: No, I think it's not necessary to add that test, because the only thing that it tested (that has not been tested in other tests) is the issue with Name/Sequence and we no longer have it.
ping @TheMule71

TheMule71: I agree that having a second import_csv test adds nothing and can be removed, but I'd still put a line with a number in the description as part of the regular test_import_csv_file() test we already have.

The cost is negligible and we're covering a case that did prove problematic in the past.

Closing https://github.com/PNLUG/bank-statement-import/pull/5

primes2h: > I agree that having a second import_csv test adds nothing and can be removed, but I'd still put a line with a number in the description as part of the regular test_import_csv_file() test we already have.

Done, thank you!

pedrobaeza:
/ocabot merge nobump

/ocabot migration account_bank_statement_import_txt_xlsx > account_statement_import_txt_xlsx

PR #421, opened 2021-09-30
[FIX] account_bank_statement_import_online_ponto: verify TLS

yajo:
Not sure why, cert validation was disabled. That is insecure.

@Tecnativa TT31034


alexey-pelykh (approved):

pedrobaeza:
Duplicated of #420

PR #423, opened 2021-10-16
[14.0] account_statement_import_online: fixing empty  assuming  is a good default value

### LINKED ISSUE #422: [14.0] account_statement_import_online: payment_ref can be empty
I have a case with ponto where my `payment_ref` is empty. Pull ended with:

```
L'opération ne peut pas être terminée:
- Créer / mettre à jour: un champ obligatoire n'est pas défini.
- Supprimer: un autre modèle nécessite la suppression de l'enregistrement. Si possible, archivez-le.

Modèle: Bank Statement Line (account.bank.statement.line), Champ : Label (payment_ref)
```

I create a pull request assuming if empty, we use `ref` as `payment_ref`.

njeudy:
- Fix #422

### DIFF
--- a/account_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_statement_import_online/models/online_bank_statement_provider.py
@@ -299,6 +299,8 @@ class OnlineBankStatementProvider(models.Model):
                         ),
                     }
                 )
+            if not line_values.get("payment_ref"):
+                line_values["payment_ref"] = line_values.get("ref")
             filtered_lines.append(line_values)
         return filtered_lines
 


account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -299,6 +299,9 @@ def _get_statement_filtered_lines(
                         ),
                     }
                 )
+
```
pedrobaeza: Please don't add extra empty lines inside a method.

pedrobaeza (approved):

account_statement_import_online/models/online_bank_statement_provider.py
njeudy: ok done :)

PR #426, opened 2021-10-19
[14.0][IMP] account_statement_import_camt: fix multi currency support

TeoGoddet:
It now ensure that the amount found is in the correct (account) currency.
It also tries to have the amount in foreign currency and the foreign currency set.

I may need to extend the tests set


anedisi:
is there any plan to merge this. im facing the same problem

TeoGoddet:
I could finally add the required tests :)
@anedisi did you test it ? does it fit ?

renzo1974 (approved):

anedisi:
@TeoGoddet sorry i did not have time to test this, i needed a quick fix in my code so bank statements would work so i dealt with it another way.

account_statement_import_camt/models/parser.py
```diff
@@ -13,25 +13,73 @@ class CamtParser(models.AbstractModel):
...
             amount_node = node.xpath(
                 "./ns:AmtDtls/ns:TxAmt/ns:Amt", namespaces={"ns": ns}
             )
-        if amount_node:
-            amount = sign * float(amount_node[0].text)
-        return amount
+        if (
+            not amount_node
+            or currency is not None
+            and amount_node[0].xpath("./@Ccy", namespaces={"ns": ns})[0] != currency
```
StefanRijnhart: Looks like you may be missing a couple of parenthesis here. Also, you could condense the code (making it more readable) by actually iterating over the various paths and breaking if the conditions are met.
```
for path in [ ... ]
    if
        break
else
    return (None, None, None))
```

I would feel safer if you'd raise in the situation that you *did* encounter one or more amount nodes but none of them were in the right currency.

account_statement_import_camt/models/parser.py
```diff
@@ -13,25 +13,73 @@ class CamtParser(models.AbstractModel):
     _name = "account.statement.import.camt.parser"
     _description = "Account Bank Statement Import CAMT parser"
 
-    def parse_amount(self, ns, node):
+    def parse_amount(self, ns, node, currency=None):
         """Parse element that contains Amount and CreditDebitIndicator."""
```
StefanRijnhart: Can you document the (new) return format here?

StefanRijnhart (changes requested):
Thanks, this is a good step towards further completeness in covering the Camt specifications. Please have a look at my comments.

account_statement_import_camt/models/parser.py
TeoGoddet: done

account_statement_import_camt/models/parser.py
TeoGoddet: done

TeoGoddet:
Hi ! @StefanRijnhart, thanks for the review, 
I updated according to your comments.

TeoGoddet:
I think it's ready now, 
@StefanRijnhart could you please review, especially the test I needed to change on camt54. (balance = None instead of 0 after this when no amount found like in camt54)
I tested your test-camt54 file on this [runboat statement](http://oca-bank-statement-import-14-0-pr426-16c99f0c02ef.runboat.odoo-community.org/web?debug=1#id=2&action=215&active_id=2&model=account.bank.statement&view_type=form&cids=1&menu_id=) but I have no other camt54 file to test.

StefanRijnhart (approved):

PR #427, opened 2021-11-04
[14.0] [IMP] account_statement_import_camt: unique statement names

hailangvn:
* Most import formats provide unique transaction ids, which allows to
ensure no redundant import happens
(https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import/models/account_bank_statement_line.py#L14)
CAMT format doesn't, so we add a new feature to check for uniqueness at
the statement name level

* The block `Bank & Cash` is only visible to `group_account_user`
(https://github.com/odoo/odoo/blob/14.0/addons/account/views/res_config_settings_views.xml#L448).
But admin is `group_account_manager` so he cannot view the block
(https://github.com/odoo/odoo/blob/14.0/addons/account/security/account_security.xml#L25).
That is why `Bank Statement Import` block is added for new config
parameter.

### DIFF
--- a/account_statement_import_camt/models/res_config_settings.py
+++ b/account_statement_import_camt/models/res_config_settings.py
@@ -1,0 +1,17 @@
+# Copyright 2021 CampToCamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+MODULE = "account_statement_import_camt"
+CHECKING_NAME_UNIQUE = MODULE + ".checking_bank_statement_name_unique"
+
+
+class ResConfigSetting(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    checking_bank_statement_name_unique = fields.Boolean(
+        string="Unique bank statement name",
+        config_parameter=CHECKING_NAME_UNIQUE,
+        default=False,
+    )
--- a/account_statement_import_camt/views/account_bank_statement_import.xml
+++ b/account_statement_import_camt/views/account_bank_statement_import.xml
@@ -11,6 +11,11 @@
                 <li>CAMT</li>
                 <li>zipped CAMT</li>
             </ul>
+            <ul id="statement_format" position="after">
+                <p
+                    id="checking_name_unique"
+                >Unique bank statement name checking (needed by CAMT importing) can be turned on at Settings > Invoicing > Bank Statement Import > Bank Statement Import.</p>
+            </ul>
         </field>
     </record>
 </odoo>
--- a/account_statement_import_camt/__manifest__.py
+++ b/account_statement_import_camt/__manifest__.py
@@ -8,5 +8,8 @@
     "website": "https://github.com/OCA/bank-statement-import",
     "category": "Banking addons",
     "depends": ["account_statement_import"],
-    "data": ["views/account_bank_statement_import.xml"],
+    "data": [
+        "views/account_bank_statement_import.xml",
+        "views/res_config_settings.xml",
+    ],
 }
--- a/account_statement_import_camt/models/__init__.py
+++ b/account_statement_import_camt/models/__init__.py
@@ -1,3 +1,4 @@
+from . import res_config_settings
 from . import parser
 from . import account_bank_statement_line
 from . import account_statement_import
--- a/account_statement_import_camt/models/account_statement_import.py
+++ b/account_statement_import_camt/models/account_statement_import.py
@@ -5,6 +5,9 @@ import zipfile
 from io import BytesIO
 
 from odoo import models
+from odoo.exceptions import ValidationError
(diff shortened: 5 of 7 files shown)


account_statement_import_camt/models/res_config_settings.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 CampToCamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+MODULE = __name__.split(".")[2]
```
sebalix: Seems a bit wobbly, I think you can hardcode the module name here
```suggestion
MODULE = "account_statement_import_camt"
```
There is surely a better approach but I think it's fine by doing so.

hailangvn:
Thanks @sebalix for the suggestion. This PR has been updated.

simahawk:
@sebalix can you get back to this?

@StefanRijnhart could you have a look at this?

TeoGoddet (approved):
Hello ! Thanks for this really nice feature !
I'm wondering if the checking_bank_statement_name_unique fields would no be better on the account.journal ?

hailangvn:
Thanks @TeoGoddet for your comment. IMHO, it is simpler to just set at one place to have consistent behavior instead of setting multiple journals. I will check the journal option when situation is changed.

sebalix:
@simahawk sorry for the delay, can you re-open this so I can approve? We should port this in next versions I guess

account_statement_import_camt/models/res_config_settings.py
```diff
@@ -0,0 +1,17 @@
...
+
+MODULE = "account_statement_import_camt"
+CHECKING_NAME_UNIQUE = MODULE + ".checking_bank_statement_name_unique"
+
+
+class ResConfigSetting(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    checking_bank_statement_name_unique = fields.Boolean(
+        string="Unique bank statement name",
```
StefanRijnhart, line 14: Because the scope of the parameter is limited to CAMT statements, I would prefer it if tthe field name and string referred to CAMT to make this clear to the user.

account_statement_import_camt/views/account_bank_statement_import.xml
```diff
@@ -11,6 +11,11 @@
                 <li>CAMT</li>
                 <li>zipped CAMT</li>
             </ul>
+            <ul id="statement_format" position="after">
+                <p
+                    id="checking_name_unique"
+                >Unique bank statement name checking can be turned on at Settings > Invoicing > Bank Statement Import > Bank Statement Import.</p>
```
StefanRijnhart:
```suggestion
                >Unique bank statement name checking for CAMT statements can be turned on at Settings > Invoicing > Bank Statement Import > Bank Statement Import.</p>
```

account_statement_import_camt/models/res_config_settings.py
sebalix: @hailangvn can you address this please? So we can merge :)

hailangvn: Thanks @StefanRijnhart for your suggestion. I agree that the field name and string should reflect the logic. But in the code to check duplicated names, I could not find any logic specific to camt. The code only check for generic field 'name'. It may confuses because it is in camt module and can be split to another module. But that work is for future when we see another module needs the logic.

account_statement_import_camt/views/account_bank_statement_import.xml
hailangvn: Thanks for the suggestion. It has been applied with a little difference.

hailangvn:
Thanks @StefanRijnhart and @sebalix for your review. This PR has been updated and re-based.

account_statement_import_camt/views/account_bank_statement_import.xml
```diff
@@ -11,6 +11,11 @@
                 <li>CAMT</li>
                 <li>zipped CAMT</li>
             </ul>
+            <ul id="statement_format" position="after">
+                <p
+                    id="checking_name_unique"
+                >Unique bank statement name checking (needed by CAMT importing) can be turned on at Settings > Invoicing > Bank Statement Import > Bank Statement Import.</p>
```
StefanRijnhart, line 17: This still sounds ambiguous to me. It sounds like it would work for all bank statement formats, and is required for CAMT statements, whereas effectively it only works for CAMT statements and it is not actually required for CAMT statements to be imported.

account_statement_import_camt/models/res_config_settings.py
StefanRijnhart: OK, we clearly don't agree. I'll abstain from further review.

PR #431, opened 2021-12-13
[14.0] [IMP] account_statement_import_online[_ponto] improvements.

NL66278:
1. Find partner if there already is a bank account with the right IBAN;
2. Store (and display) raw import data to help in problem determination;
3. Make it easy to extend the parsing of import data;
4. Apply some clean coding principles.

### DIFF
--- a/account_statement_import_online/models/account_bank_statement_line.py
+++ b/account_statement_import_online/models/account_bank_statement_line.py
@@ -1,0 +1,18 @@
+# Copyright 2021 Therp BV <https://therp.nl>.
+# @author: Ronald Portier <ronald@therp.nl>.
+# Licence LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0).
+"""Add raw data to statement line, to solve import issues."""
+
+from odoo import fields, models
+
+
+class AccountBankStatementLine(models.Model):
+    """Add raw data to statement line, to solve import issues."""
+
+    _inherit = "account.bank.statement.line"
+
+    raw_data = fields.Text(
+        help="The complete data retrieved online for this transaction",
+        readonly=True,
+        copy=False,
+    )
--- a/account_statement_import_online/models/account_bank_statement_line.py
+++ b/account_statement_import_online/models/account_bank_statement_line.py
@@ -11,7 +11,7 @@ class AccountBankStatementLine(models.Model):
 
     _inherit = "account.bank.statement.line"
 
-    raw_data = fields.Text(
+    online_raw_data = fields.Text(
         help="The complete data retrieved online for this transaction",
         readonly=True,
         copy=False,
--- a/account_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_statement_import_online/models/online_bank_statement_provider.py
@@ -292,13 +292,11 @@ class OnlineBankStatementProvider(models.Model):
                     continue
             bank_account_number = line_values.get("account_number")
             if bank_account_number:
-                line_values.update(
-                    {
-                        "account_number": (
-                            self._sanitize_bank_account_number(bank_account_number)
-                        ),
-                    }
+                sanitized_account_number = self._sanitize_bank_account_number(
+                    bank_account_number
                 )
+                line_values["account_number"] = sanitized_account_number
+                self._update_partner_from_account_number(line_values)
             if not line_values.get("payment_ref"):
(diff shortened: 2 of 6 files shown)


account_statement_import_online/models/account_bank_statement_line.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo import fields, models
+
+
+class AccountBankStatementLine(models.Model):
+    """Add raw data to statement line, to solve import issues."""
+
+    _inherit = "account.bank.statement.line"
+
+    raw_data = fields.Text(
```
pedrobaeza:
```suggestion
    online_raw_data = fields.Text(
```
?

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -365,6 +363,14 @@ def _sanitize_bank_account_number(self, bank_account_number):
         self.ensure_one()
         return sanitize_account_number(bank_account_number)
 
+    def _update_partner_from_account_number(self, line_values):
+        partner_bank = self.env["res.partner.bank"].search(
+            [("acc_number", "=", line_values["account_number"])], limit=1
```
pedrobaeza: No `company_id` restriction?

account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
```diff
@@ -156,34 +156,32 @@ def _ponto_get_transaction(self, account_id, date_since, date_until):
...
-                        if not page_next or (page_next and not latest_identifier):
-                            latest_identifier = current_transactions[0].get("id")
-                        transaction_lines.extend(current_transactions)
-            else:
+            if response.status_code != 200:
                 raise UserError(
                     _("Error during get transaction.\n\n%s \n\n %s")
                     % (response.status_code, response.text)
                 )
+            if params.get("before"):
```
pedrobaeza, line 164: Why changing this order? Please try to minimize diff

account_statement_import_online/models/account_bank_statement_line.py
NL66278: I will change the name

account_statement_import_online/models/online_bank_statement_provider.py
NL66278: I took this from the not online account_statement_import. res.partner.bank statements with company_id filled are own accounts, not partner accounts. Maybe I should add that company_id should be False.

account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
NL66278: I thought it was cleaner, saving a lot of indented code and getting rid of an `else` branch. But if you still object I will change it back.

account_statement_import_online/models/online_bank_statement_provider.py
pedrobaeza: Can you have the same bank account number in several companies? If so, then you should check by company_id = company of the statement or company_id = False.

account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
pedrobaeza: Although I can agree on the principle, this makes more difficult to review the improvements. Maybe you can add an initial commit doing this refactoring, and another with the improvements.

account_statement_import_online/models/online_bank_statement_provider.py
NL66278: O wait, that was in in old version of Odoo.

NL66278: Still, I do not see a need to check for company. You can have a partner without a company_id, meaning valid everywhere, and then when adding a bank_account, the res.partner.bank record gets the company_id of the logged in user. I think that is illogical, and unneeded, as it is still not a required field.

pedrobaeza: I insist: the problem is that you have the same bank account number in several companies, and with this, you select the incorrect bank account belonging to other company.

NL66278:
@pedrobaeza I think I attended to all your comments now.

alexis-via (approved):
code review, no test

rlizana (approved):
Code review, @pedrobaeza comments is satisfied, not test!

PR #437, opened 2022-01-08
[13.0][IMP]account_bank_statement_import_camt_oca - line note

luc-demeyer:


luc-demeyer:
The OCA camt parser doesn't contain sufficient information in the note field to facilitate reconciliation.
Cf. alsohttps://github.com/odoo/odoo/commit/668b0c71fd680b3a0f0dfab2e7482f1373426d25

TeoGoddet:
I fully agree we need more infos, 
actually, I was doing some work for this, i pushed as WIP in #438 

We had very similar approach :)

We need to define the fields we want and the approach to make a single PR ?

TeoGoddet:
Also I'm not sure we should use translation because it may make the regex fail depending on the language context when the import is done.

account_bank_statement_import_paypal/__manifest__.py
```diff
@@ -18,7 +18,6 @@
         "multi_step_wizard",
         "web_widget_dropdown_dynamic",
     ],
-    "external_dependencies": {"python": ["csv"]},
```
TeoGoddet, line 21: This is not related to this PR

BenjaHe (approved):
Thank you

luc-demeyer:
@TeoGoddet 

>>> We need to define the fields we want and the approach to make a single PR ?

I have added the info parsed by the current (13.0) oca camt parser in the line notes field.
I agree that many fields defined by the camt protocol are not parsed currently but adding these should imho be part of a seperate PR, hence I think we should use your PR  https://github.com/OCA/bank-statement-import/pull/438 for this purpose.

>>> Also I'm not sure we should use translation because it may make the regex fail depending on the language context when the import is done.

Imho we must have translations on this field. 
I have created this PR for the german subsidiary of one of our belgian customers. 
The lady that processes the bank statements doesn't understand a single word of english, hence for the 'sponsoring' customer of this PR translation is a must. I am sure there are many persons in the Odoo end user community without english language skills. The info that we add into the Notes (e.g. additional transaction info will in german when downloading german camt files, in dutch for camt files from dutch banks, ... ).

>>> "external_dependencies": {"python": ["csv"]},

Indeed not related to this PR but required to 'greenify' this PR.

TeoGoddet:
Ok, 
nevertheless, the diff between 13 and 14 is too big to have the same pr ported
I'll keep #438 for the new fields in v14 and include the fields you added.


Ok for the translation, let's merge it as is. What do you think of smth like Translated field name (iso field code): value ? 
I'm afraid it's too long but will propose it for v14

TeoGoddet (approved):

luc-demeyer:
@TeoGoddet 
>>> What do you think of smth like Translated field name (iso field code): value ?
I think regex remains possible since you have added the Camt XML tags in the narration field, the translator groups in the OCA will not translate these tags.
Since you are based in Suisse you still may have the problem that the bank generates a different Camt file depending on the language of the user downloading it, hence creating certain regex rules will require a bit of specialised regex knowledge to support multiple languages for the same type of transaction.

vanderperre (approved):

luc-demeyer:
I think this one is ready for merge. Can somebody launch the ocabot merge ?

PR #444, opened 2022-02-03
[13.0][ADD] account_bank_statement_import_online_nordigen

cormaza:
Nordigen Integration

@ForgeFlow

PR #445, opened 2022-02-03
[13.0][FIX] account_bank_statement_import_online: Default value for active field

cormaza:
@ForgeFlow

### DIFF
--- a/account_bank_statement_import_online/models/account_journal.py
+++ b/account_bank_statement_import_online/models/account_journal.py
@@ -61,6 +61,7 @@ class AccountJournal(models.Model):
                 OnlineBankStatementProvider.create({
                     "journal_id": journal.id,
                     "service": journal.online_bank_statement_provider,
+                    "active": True,
                 })
             # fmt: on
 


pedrobaeza:
I would say this default=False is on design, for avoiding to see non yet enabled providers. What are the steps to reproduce an error on this?

cormaza:
> I would say this default=False is on design, for avoiding to see non yet enabled providers. What are the steps to reproduce an error on this?

On this PR #444 

When I was testing, I have a problem with controller searching provider, I check that provider is inactivated, I think that is a little weird to create a register in inactive state, because that I think that is an issue, but if initial design is make record inactive by default, I remove this PR, anyway I pass active_test context for search inactive records, WDYT?

pedrobaeza:
Controller? The way I activate this is by going to the journal, and selecting the provider, and then it can be configured, and this is working for MyPonto. Can you please check and tell possible steps to reproduce a problem?

ChrisOForgeFlow:
> Controller? The way I activate this is by going to the journal, and selecting the provider, and then it can be configured, and this is working for MyPonto. Can you please check and tell possible steps to reproduce a problem?

In nordigen case I needed to make a controller to store data related with agreement on account of user, and select Iban of acccount, in another implementations like MyPonto everything is doing in backoffice, because that we can create online.bank.statement.provide with active on False and there's no issue, because all process works with m2o field online_bank_statement_provider_id not with search

https://github.com/OCA/bank-statement-import/blob/c648e7fb3afa3ec15ba2bb45fe05d00a1e854f5e/account_bank_statement_import_online/models/account_journal.py#L61-L64

In this lines you can check that every register created is not active by default

https://user-images.githubusercontent.com/94866688/152533091-a3c42154-2eee-4f1a-beab-5af7be41f492.mp4

AaronHForgeFlow:
Is it possible to add way to activate the bank.online.provider by hand?

pedrobaeza:
Then just add `active=True` in such create statement.

ChrisOForgeFlow:
> Then just add `active=True` in such create statement.

Done!, thanks for review

ChrisOForgeFlow:
> Is it possible to add way to activate the bank.online.provider by hand?

There's a way, but I think that is better make active by default on creation from journal configuration process

AaronHForgeFlow (approved):

PR #446, opened 2022-02-03
[12.0][ADD] account_bank_statement_import_online_nordigen

ChrisOForgeFlow:
Integration Nordigen

@ForgeFlow


account_bank_statement_import_online_nordigen/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (C) 2022 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Online Bank Statements: Nordigen",
+    "version": "13.0.1.0.0",
```
MateuGForgeFlow: Hi @ChrisOForgeFlow, maybe you already noticed it, just making sure. The version should be 12.0! Thank you!

ChrisOForgeFlow: Thanks, I solved

rafaelbn (approved):
LGTM, maybe you can add some test to avoid errors 👍 

Thanks

rafaelbn:
Hello @ChrisOForgeFlow @MateuGForgeFlow , do you plan to get this merge? Thank you! 😄

ChrisOForgeFlow:
> Hello @ChrisOForgeFlow @MateuGForgeFlow , do you plan to get this merge? Thank you! smile

Sure!, I'm working in tests to finish PR on every version 12, 13, and 14

account_bank_statement_import_online_nordigen/static/src/lib/nordigen-ui/selector.css
```diff
@@ -0,0 +1,209 @@
+
+* {
```
StefanRijnhart, line 2: This css file is added to `web.assets_backend`. That means it is applied to the whole backend layout, as long as the elements match the selectors. I was going to recommend therefore that you use class names that are a little bit more specific to the module (e.g. `nordigen_institution_*` rather than just `institution_*`).

Moreover, here you are using the universal selector (just like you are using the `html` selector further down). Will that not lead to problems in other parts of the system?

(my web dev skills are notoriously lacking, so let me know if I am missing anything here)

PR #448, opened 2022-02-22
PayPal import fails if there is number in details or description

thomas-work:
The regex matches lines like "Some Caption 39 xx 30"
_sequence_yearly_regex = r'^(?P<prefix1>.*?)(?P<year>((?<=\D)|(?<=^))((19|20|21)?\d{2}))(?P<prefix2>\D+?)(?P<seq>\d*)(?P<suffix>\D*?)$'

The test in _constrains_date_sequence fails and a ValidationError is thrown.

The other patch adds the data to the new transaction and does not update the source data.

### DIFF
--- a/account_statement_import_paypal/models/account_statement_import_paypal_parser.py
+++ b/account_statement_import_paypal/models/account_statement_import_paypal_parser.py
@@ -207,9 +207,9 @@ class AccountBankStatementImportPayPalParser(models.TransientModel):
             note += " (%s)" % payer_email
 
         unique_import_id = "{}-{}".format(transaction_id, int(timestamp.timestamp()))
-        name = (invoice or details or description or "",)
+        name = (invoice or transaction_id or "",)
         transaction = {
-            "name": invoice or details or description or "",
+            "name": name[0],
             "amount": str(gross_amount),
             "date": timestamp,
             "payment_ref": note,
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Hi @thomas-work, thanks for your contribution! Did you manage to submit a CLA already?

account_statement_import_paypal/models/account_statement_import_paypal_parser.py
```diff
@@ -207,18 +207,18 @@ def _convert_line_to_transactions(self, line):
             note += " (%s)" % payer_email
 
         unique_import_id = "{}-{}".format(transaction_id, int(timestamp.timestamp()))
-        name = (invoice or details or description or "",)
+        name = (invoice or transaction_id or "",)
         transaction = {
-            "name": invoice or details or description or "",
+            "name": name[0],
```
StefanRijnhart, line 212: I believe the best way to deal with the API change in Odoo 14, where the statement line inherits from account.move, the `name` attribute should not be set. Instead, the data that was assigned to `name` in earlier versions can be assigned to `ref` instead. See for instance this migration of a paypal module: https://github.com/OCA/bank-statement-import/pull/452/commits/12b7bfa128ac6d1fb313e529478b5e78a4393cf3#diff-84f1950cdde1a8ef9a098ed89c154cb4afd13d955a0088c978a570b7461ab604L295-R295

Do you think that would work in this case too?

PR #449, opened 2022-02-23
[14.0][FIX] account_statement_import: abs lines don't need partner_bank_id field

MiquelRForgeFlow:
Let them use the one from account moves.

Followup of https://github.com/OCA/OpenUpgrade/pull/3150.

NOTE: When merging this PR, use ocabot `major`

### DIFF
--- a/account_statement_import/migrations/14.0.2.0.0/pre-migration.py
+++ b/account_statement_import/migrations/14.0.2.0.0/pre-migration.py
@@ -1,0 +1,26 @@
+# Copyright 2022 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.column_exists(
+        env.cr, "account_bank_statement_line", "partner_bank_id"
+    ):
+        # during v14, a partner_bank_id field was added to statement lines,
+        # but later we realized it is not needed.
+        # As we remove the explicit partner_bank_id in statement line, we need to transfer the
+        # values to the move for getting them through delegated inheritance
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE account_move am
+            SET partner_bank_id = absl.partner_bank_id
+            FROM account_bank_statement_line absl
+            WHERE am.statement_line_id = absl.id
+                AND am.partner_bank_id IS NULL AND absl.partner_bank_id IS NOT NULL""",
+        )
+        openupgrade.lift_constraints(
+            env.cr, "account_bank_statement_line", "partner_bank_id"
+        )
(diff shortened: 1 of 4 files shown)


AaronHForgeFlow (approved):
Code review LGTM

account_statement_import/models/account_bank_statement_line.py
```diff
@@ -12,10 +12,6 @@ class AccountBankStatementLine(models.Model):
     # Ensure transactions can be imported only once
     # (if the import format provides unique transaction ids)
     unique_import_id = fields.Char(string="Import ID", readonly=True, copy=False)
-    # v13 field: bank_partner_id
-    # This field was removed in v14, but it is still used in the code, cf the
-    # method reconcile() !!! So I restore the field here
-    partner_bank_id = fields.Many2one("res.partner.bank", string="Partner Bank Account")
```
pedrobaeza, line 18: It seems to be used here:

https://github.com/OCA/account-reconcile/blob/e5ab5ffc3b696c66a174f6cde7f6bfbcb34e1e46/account_reconciliation_widget/models/account_bank_statement.py#L203

so maybe a change in such module is needed as well.

MiquelRForgeFlow: No, that's a verbatim copy of https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_bank_statement.py#L1210. The statement lines inherit from account moves, so the self.partner_bank_id odoo interprets it as self.move_id.partner_bank_id, no need to create a new explicit field in statement lines. That's the point of my PR.

account_statement_import/migrations/14.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+        env.cr, "account_bank_statement_line", "partner_bank_id"
+    ):
+        # between 14.0.1.0.0 and 14.0.2.0.0, lines had a partner_bank_id field
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE account_move am
+            SET partner_bank_id = COALESCE(absl.partner_bank_id, am.partner_bank_id)
+            FROM account_bank_statement_line absl
+            WHERE am.statement_line_id = absl.id
```
pedrobaeza: For not updating more rows than needed:
```suggestion
            UPDATE account_move am
            SET partner_bank_id = absl.partner_bank_id
            FROM account_bank_statement_line absl
            WHERE am.statement_line_id = absl.id
            AND am.partner_bank_id IS NULL AND absl.partner_bank_id IS NOT NULL
```

account_statement_import/models/account_bank_statement_line.py
pedrobaeza: OK, thanks for the explanations.

pedrobaeza (approved):

MiquelRForgeFlow:
nobump? should be major!

pedrobaeza (approved):

PR #450, opened 2022-02-24
[15.0][FIX] account_statement_import: abs lines don't need partner_bank_id field

MiquelRForgeFlow:
Let them use the one from account moves.

Forward port of https://github.com/OCA/bank-statement-import/pull/449


AaronHForgeFlow (approved):
approval forward port

pedrobaeza (approved):

PR #451, opened 2022-02-25
[15.0][FIX] account_statement_import: Change demo data

Shide:
Change demo data to not override Odoo `account_bank_statement_import` module and let the tests pass if both modules are going to be installed.

### DIFF
--- a/account_statement_import/demo/partner_bank.xml
+++ b/account_statement_import/demo/partner_bank.xml
@@ -1,30 +1,31 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!--
   Copyright 2004-2020 Odoo S.A.
+  Copyright 2022 Moduon Team <edu@moduon.team>
   Licence LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0).
 -->
 <odoo>
 
     <record id="ofx_partner_bank_1" model="res.partner.bank">
-        <field name="acc_number">BE68539007547034</field>
+        <field name="acc_number">BE02631118262640</field>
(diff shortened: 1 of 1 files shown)


yajo (approved):

yajo:
FTR the problem is not that the data is overriding anything, but that there's a unique key in postgres that forbids installing both modules demo data in parallel in an integration environment:

```
2022-02-25 07:42:26,826 27 INFO odoo odoo.modules.loading: loading account_bank_statement_import/demo/partner_bank.xml 
2022-02-25 07:42:26,875 27 ERROR odoo odoo.sql_db: bad query: UPDATE "res_partner_bank" SET "sanitized_acc_number"='BE68539007547034',"write_date"='2022-02-25T07:42:25.452870'::timestamp,"write_uid"=1 WHERE id IN (9)
ERROR: duplicate key value violates unique constraint "res_partner_bank_unique_number"
DETAIL:  Key (sanitized_acc_number, company_id)=(BE68539007547034, 1) already exists.
```

MiquelRForgeFlow (approved):
Yeah.

rafaelbn (approved):
Thank you ! 👍 

@StefanRijnhart @sbidoul could you please take a look? 🙏

StefanRijnhart (approved):

MiquelRForgeFlow:
My question is: Why would you install both `account_statement_import` (OCA) and `account_bank_statement_import` (Enterprise) with demo data? In which case it happens?

vincentius:
@MiquelRForgeFlow On odoo.sh test suites this may happen.

PR #454, opened 2022-03-07
[14.0][FIX] account_statement_import_online: use unittest.mock and don't import tests at install

yajo:
Installing this module failed because `mock` couldn't be found. Indeed there was a migration problem.

However, I was not testing the module; just installing. At install, tests shouldn't be imported.

Fixing both problems here.

@moduon MT-295

### DIFF
--- a/account_statement_import_online/__init__.py
+++ b/account_statement_import_online/__init__.py
@@ -2,4 +2,3 @@
 
 from . import models
 from . import wizards
-from .tests import online_bank_statement_provider_dummy
--- a/account_statement_import_online/__manifest__.py
+++ b/account_statement_import_online/__manifest__.py
@@ -11,6 +11,7 @@
     "license": "AGPL-3",
     "category": "Accounting",
     "summary": "Online bank statements update",
+    "external_dependencies": {"python": ["odoo_test_helper"]},
     "depends": [
         "account",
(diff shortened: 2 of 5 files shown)


rafaelbn:
Hello @cormaza @ChrisOForgeFlow what do you think? Thank you! :-)

rafaelbn:
Hello @NL66278 @njeudy what do you think? 🙏 Thank you!

ChrisOForgeFlow:
Maybe should be good delete account_statement_import_online/tests/online_bank_statement_provider_dummy.py file too

NL66278:
The dummy model is needed for running the tests, as can be seen from all tests failing at the moment. Why not just correct the import of the mock?

yajo:
Let me see if I can fix it with https://github.com/OCA/odoo-test-helper.

yajo:
Let's see how it goes now.

FTR `mock` still has to be installed until https://github.com/OCA/odoo-test-helper/pull/19 is merged, which is a tiny PR. Please review if possible.

yajo:
Tests are ✔️ now.

pedrobaeza (approved):

NL66278 (approved):
:+1: LGTM

PR #455, opened 2022-03-16
[WIP][13.0][IMP] account_bank_statement_import_txt_xlsx: header/footer trimming, separate debit/credit

jjscarafia:
Cherry pick of this PR for v12 https://github.com/OCA/bank-statement-import/pull/296
With some fixes and improvements

Pending fixes/improvements:
* usage of "csv_or_xlsx_lst"
* give the user better information when the mapping is wrong

PR #457, opened 2022-04-03
[14.0][IMP] account_statement_import_txt_xlsx: amount_debit_column & amount_credit_column

ghost:
I have bank statements with two amount columns: one for payments coming IN, and another for payments going OUT.
This PR is adding "amount2_column".
Assign "IN" or "OUT" to "amount", and assign the other one to "amount2".


TeoGoddet:
Check #455

TeoGoddet:
The code is correct, I still need to do a functional review 
But I wonder how you find out the sign in your 2 column file ?

ghost:
Hi @TeoGoddet 
My bank statement has positive numbers for payments coming in, and negative numbers for payments going out. If all numbers would be positive, I think there would be need for another `debit_credit_column2`. Maybe we should rename the other field to `amount_column2`. And put debit_credit_column(2) on the right side of the amount_column(2) in the UI form. What do you think?

TeoGoddet:
Ok, I see, I wasn't expecting this file format, let's keep the credit/debit for another PR
Maybe an easy way would be to allow a regex or "any non empty non zero value" in the credit/debit value

TeoGoddet (approved):
Code review only !

ghost:
@pedrobaeza Would you like to review this PR?

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
```diff
@@ -189,7 +192,9 @@ def _parse_rows(self, mapping, currency_code, csv_or_xlsx, columns):  # noqa: C9
                 if columns["currency_column"] is not None
                 else currency_code
             )
-            amount = values[columns["amount_column"]]
+            amount = (
+                values[columns["amount_column"]] or values[columns["amount2_column"]]
```
pedrobaeza: Shouldn't it be the second column with negative?

norlinhenrik:
My bank statement has negative numbers in the second column. Do you think we should rename the field to `amount_negative_column`?

pedrobaeza:
Why not adding an extra field to indicate if the amount should be reversed, and it makes it more versatile?

norlinhenrik:
I have added `amount2_reverse` boolean field.

account_statement_import_txt_xlsx/models/account_statement_import_sheet_mapping.py
```diff
@@ -69,6 +69,14 @@ class AccountStatementImportSheetMapping(models.Model):
         required=True,
         help="Amount of transaction in journal's currency",
     )
+    amount2_column = fields.Char(
```
alexey-pelykh: I'd use debit_amount and credit_amount and used absolute value of those instead.

alexey-pelykh: That is by far easier to understand w/o looking into code what `amount2` means IMHO

pedrobaeza: Not really, as that means 3 extra fields (instead of 2) or less options (not being able to reverse the value if needed). Absolute values will not cover both options.

alexey-pelykh: @pedrobaeza what would be the point of reversing debit column value? If it's debit, just process its absolute value as debit. Reversing option is not needed at all

pedrobaeza: Well, I'm not sure all cases fit such logic, but if you think so, let's go with it.

alexey-pelykh: Well, unless we can't think of any cases that would fail - I guess it's worth a try. `amount`, `debit_amount`, `credit_amount`. `amount` for compatibility looks at sign and believes it, `debit_amount` ignores sign and treats as debit, `credit_amount` similarly. The only case I can think of that would require `reverse_amount` boolean field is when there's only one column and it's reversed.

alexey-pelykh: Just don't do vague `amount2` please

norlinhenrik:
Hello @alexey-pelykh 
Thank you for the feedback! I think you have good point, and I wish we had this conversation in April when I introduced this PR and started using it in production. Yesterday I spent 4 hours to implement the suggestions from @pedrobaeza, and I don't have more time for this project now. Would you like to add a new commit to this PR to implement your suggestion?

pedrobaeza:
Hi, Henrik, I know that sometimes the community is not fair, having a PR dead for months, and then demanding you more changes that don't bring direct value to you. If @alexey-pelykh is not able to do the change, I will do it for not letting this PR unmerged.

alexey-pelykh:
@pedrobaeza I'm not sure what scares me more, `amount2` field or you doing the change 😝 I honestly don't know when I'll be able to take care of this, hopefully sometime still this year.

norlinhenrik:
I actually need this change now, so I will try to do it.

norlinhenrik:
@alexey-pelykh Would you prefer
```
amount_column
debit_amount_column
credit_amount_column
```
or
```
amount_column
amount_debit_column
amount_credit_column
```
?

alexey-pelykh:
@norlinhenrik thanks for asking, no preference between those. @pedrobaeza you?

pedrobaeza:
Second option for me for having all fields together alphabetically.

norlinhenrik:
Now the PR is ready. The migration will need **merge minor** to work properly.

I see 5 PRs named `account_bank_statement_import_txt_xlsx`. The name should be `account_statement_import_txt_xlsx`, right?

account_statement_import_txt_xlsx/models/account_statement_import_sheet_mapping.py
```diff
@@ -66,9 +66,16 @@ class AccountStatementImportSheetMapping(models.Model):
     )
     amount_column = fields.Char(
         string="Amount column",
-        required=True,
```
pedrobaeza, line 69: Please add an SQL constraint for having or this one, or the other 2. Some similar examples: https://github.com/odoo/odoo/blob/44ccecb994baa0f7bce4f5c39d6d136499e5be2e/addons/account/models/account_move.py#L3337

norlinhenrik: Done

pedrobaeza (approved):
Thanks for the effort!

/ocabot merge major

PR #459, opened 2022-04-10
[13.0][MIG] account_bank_statement_import_online_qonto: Migration to 13.0

pedrobaeza:
- Standard procedure
- Module icon
- Pre-commit stuff
- Adjust naming and domain to .com, as stated in Qonto API page (https://api-doc.qonto.com/docs/business-api/ZG9jOjI5NjA5OQ-getting-started)
- Better line label

@Tecnativa TT35884


account_bank_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
```diff
@@ -95,18 +95,18 @@ def _qonto_prepare_statement_line(
...
-            name = "{} {}".format(name, transaction["reference"])
+        payment_ref_list = [
+            transaction["label"],
+            transaction["reference"],
+        ]
         vals_line = {
             "sequence": sequence,
             "date": date,
-            "name": name,
+            "name": " - ".join([x for x in payment_ref_list if x]) or "/",
```
StefanRijnhart, line 105: Having done some fancy statement matching in my time it seems to me that having the dash in between can lead to some noise when splitting up the line name by whitespace so I have a slight preference for joining by space only.

StefanRijnhart (approved):

account_bank_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
pedrobaeza: I always prefer to have a more distinct joining char, being `-`, `;` or whatever. If not, you don't differentiate if the content comes from one place or another. Do you have an example of why is the space better?

StefanRijnhart: It may be a theoretical case but if you have statement references like "1 - 2" and you have outstanding items with references "3 - 4" and you split your references by whitespace you could have a match on the `dash.`
But like I said, it's not blocking for me.

pedrobaeza: OK, I let others to give their opinion. I took this decision from the 14.0 code from @alexis-via

StefanRijnhart: OK, let's just keep it.

rafaelbn (approved):
Thanks 👍

PR #461, opened 2022-04-12
[12.0][FIX] *_ponto: solve exception on empty name

NL66278:
The name for a transaction imported from Ponto is taken from the remittanceInformation, failing that,
the reference should be used. However in case the transaction attributes do contain the
remittanceInformation key, but the value is False or empty, the failover to the reference fails.


NL66278:
I had to add a second commit, similar to what was done on 14, to solve problems with runboat due to a wrongly defined external dependency in the paypal import module. As a bonus, after merging this PR, all the other 12.0 branches can be rebased to become green if they are correct by themselves.

NL66278:
Another comment: I needed this now on 12.0. But should be forward ported once accepted, as same problem exists in later versions.

rafaelbn:
Thanks @NL66278

alexey-pelykh:
Hi @NL66278, would it be reasonable to split this PR into two to extract paypal change?

alexey-pelykh (approved):

PR #462, opened 2022-04-13
[15.0][MIG]account_statement_import_online

manuelregidor:
Standard migration.

PR #463, opened 2022-04-13
[15.0][MIG]account_statement_import_online_ponto

manuelregidor:
Standard migration.
Depends: https://github.com/OCA/bank-statement-import/pull/462


HaraldPanten (approved):
Tested in production environment. Works like a charm

PR #468, opened 2022-05-17
[FIX] account_bank_statement_import_paypal don't declare standard python

hbrunn:
this fixes runboat builds, as there's no package called `csv` to install, and won't hurt anywhere else because `csv` is part of python's standard lib


NL66278 (approved):
:+1: LGTM And has already been succesfully changed in other places.

pedrobaeza (approved):

Issue #481, opened 2022-08-05
Automatic partner detection on bank statement lines


alexis-via:
I would like to discuss partner detection on bank statement lines when importing bank statements via file import (OFX, Camt, ...) AND by online import (Ponto, Qonto, ...).

The code that handle partner detection:

- in account_statement_import, there is native support for automatic partner detection using account number, cf https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import/wizard/account_statement_import.py#L281 If you want to add other detection methods, you have to inherit _complete_stmts_vals() from account.statement.import, cf https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import/wizard/account_statement_import.py#L289

- in account_statement_import_online (which depends on account_statement_import), there is native support for automatic partner detection using account number, cf https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import_online/models/online_bank_statement_provider.py#L366 If you want to add other detection methods, you have to inherit _get_statement_filtered_lines from online.bank.statement.provider cf https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import_online/models/online_bank_statement_provider.py#L256 (I didn't try it yet, but I don't see a better way).

I know that banks in Belgium (and maybe other countries) give the account_number of the party that send a wire transfer on your bank account, so the match using account_number works well and is very reliable. In France, we never have such information ; the only way we can "try" to detect a partner is by using the bank statement line label.

Partner detection methods other than account_number:

1) in the official "account" module, there is an object "account.reconcile.model.partner.mapping" https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_reconcile_model.py#L13 which is designed to detect a partner using a subset of the bank statement line label. It seems that this code is not used anywhere in the "account" module nor in the OCA module account_reconciliation_widget https://github.com/OCA/account-reconcile/tree/14.0/account_reconciliation_widget

[...]

PR #482, opened 2022-08-07
[14.0][ADD] account_statement_import_base and account_statement_import_file_reconciliation_widget

alexis-via:
The 2 modules account_statement_import_online and account_statement_import depend on account_statement_import_base (and not on each other) and share common code, in particular a hook to update the statement line. So we can now have reconciliation modules that use this hook and therefore work both on file import and online import. More details on https://github.com/OCA/bank-statement-import/issues/481.

Improve bank statement line form view and journal form view.

I still have to write the migration script for online_raw_data -> raw_data (If we are ok to rename this field and move it to the base module).


alexis-via:
Travis error seems unrelated.

pedrobaeza:
Sorry, but I don't agree on this refactoring adding more complexity on the module dependency with not too much added value.

alexey-pelykh:
@alexis-via you should rebase once https://github.com/OCA/bank-statement-import/pull/483 is merged

alexis-via:
@pedrobaeza I implemented what has been discussed in #481, so this refactoring is not a surprise.

IMHO this refactoring fixes the dependency: there was NO reason for account_statement_import_online to depend on account_statement_import. This refactoring fixes this.

The BIG advantage of this refactoring is that we can now have reconciliation modules that work BOTH on file import and online import. It should have been the case long ago, but it was unfortunately not the case until now. I made sure that reconciliation modules that inherited _complete_stmts_vals() from account_statement_import or _get_statement_filtered_lines() from account_statement_import_online will continue to work, because I didn't change the prototype of those methods.

What I can propose to simplify the refactoring: I can avoid the rename of the field online_raw_data to raw_data and the move of the field definition from account_statement_import_online to account_statement_import_base. That would simplify the refactoring while keeping the real advantage for the users (and we'll be able to move+rename this field in v16 for example).

alexis-via:
@alexey-pelykh I made the rebase

alexey-pelykh:
@alexis-via 

> account_statement_import (LGPL-3) depends on account_statement_import_base (AGPL-3)

this will have to be sorted out

alexis-via:
@alexey-pelykh I fixed the licence issue. account_statement_import_base is now LGPL.

alexis-via:
I also added a module account_statement_import_file_reconciliation_widget which is a glue auto-install module between account_statement_import and account_reconciliation_widget. It adds a button "Import and Start to Reconcile" on the wizard that allows to import and jump directly to the reconciliation interface. The associated code was moved from account_statement_import (it was present in this module but never called) to account_statement_import_file_reconciliation_widget.

bank_statement_import_start_reconcile

If you use the version of account_reconciliation_widget from the PR https://github.com/OCA/account-reconcile/pull/487 where I fixed the notifications, when you click on the new button "Import and Start to Reconcile", you will see the warning message when importing an OFX file where some statement lines were already present in Odoo:

bank_statement_v14-notifications

alexis-via:
Tests are red because "account_statement_import_file_reconciliation_widget (Mature) depends on account_reconciliation_widget (Beta)".

I made a PR to switch account_reconciliation_widget to mature https://github.com/OCA/account-reconcile/pull/491
If the PR is not accepted, I'll flag account_statement_import_file_reconciliation_widget to beta.

alexis-via:
Tests are green again (I removed the mature flag from account_statement_import_file_reconciliation_widget).

clementmbr (approved):
I don't understand the "complexity" raised by @pedrobaeza 
However, it looks great to factor common code and clean dependencies between modules.

account_statement_import_base/README.rst
```diff
@@ -0,0 +1 @@
+Will be autogenerated from the readme subdir
```
alexey-pelykh, line 1: What's the point of adding this file then?

account_statement_import_base/models/account_journal.py
```diff
@@ -0,0 +1,69 @@
...
+
+from odoo import api, models
+
+from odoo.addons.base.models.res_bank import sanitize_account_number
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    def _statement_line_import_speeddict(self):
```
alexey-pelykh, line 13: `speeddict`?

account_statement_import_base/views/account_bank_statement_line.xml
```diff
@@ -0,0 +1,61 @@
...
+            <attribute name="invisible">1</attribute>
+        </field>
+        <!--
+            I don't see anything interesting in the 'ref' field of the
+            move linked to the statement line
+            and the field move_id is displayed in the view.
+            But if you think it's useful to see this field,
+            don't hesitate to remove the 2 lines below
+        -->
+        <field name="ref" position="attributes">
```
pedrobaeza: Please don't hide fields in standard modules, and more in form views.

alexey-pelykh: 100% don't do that

pedrobaeza (dismissed):
For doing this in a compatible way with existing installations, you have to put a pre-init hook in the new module that renames the column online_raw_data to raw_data if exists.

account_statement_import_base/views/account_bank_statement_line.xml
alexis-via: ok, I removed the XML instruction for hiding "ref" in statement line form view

account_statement_import_base/README.rst
alexis-via: The tests are red if there is no README.rst at the root of the module.

account_statement_import_base/models/account_journal.py
alexis-via: speeddict is a dict to improve speed. It allows you to pre-fetch some data, which may avoid to execute some search() requests for every statement line to find the relevant data (for partner matching for example). You can find an example of the use of speeddict in my module account_statement_completion_label_simple, cf this code: https://github.com/akretion/bank-statement-reconcile-simple/blob/14-compat-import_base/account_statement_completion_label_simple/models/account_journal.py#L97

account_statement_import_base/views/account_bank_statement_line.xml
```diff
@@ -0,0 +1,51 @@
...
+    <field name="arch" type="xml">
+        <field name="statement_id" position="attributes">
+            <attribute
+                    name="invisible"
+                >not context.get('statement_line_main_view')</attribute>
+        </field>
+        <xpath expr="//field[@name='company_id']/.." position="attributes">
+            <attribute name="col">2</attribute>
+        </xpath>
+        <field name="sequence" position="attributes">
```
pedrobaeza, line 21: We meant all the opinionated hidings like this one and the following one. Please leave the standard as less touched as possible.

account_statement_import_base/views/account_bank_statement_line.xml
```diff
@@ -0,0 +1,51 @@
...
+            <notebook>
+                <page name="narration" string="Notes">
+                    <field name="narration" nolabel="1" />
+                </page>
+                <page name="technical" string="Technical Information">
+                    <group name="tech-fields">
+                        <field name="unique_import_id" />
+                        <field name="partner_name" />
+                        <field name="account_number" />
+                        <field name="is_reconciled" />
```
pedrobaeza, line 40: This field is not something of this module, so you shouldn't add it IMO as it's not directly related.

account_statement_import_base/models/account_journal.py
alexey-pelykh: Gotcha, still `speeddict` as a name feels odd if it requires that explanation 🤔

alexis-via: Well, I name it "speeddict", because it is a python dict to improve speed. But if you find a better name, feel free to suggest it.

account_statement_import_base/views/account_bank_statement_line.xml
alexis-via: unique_import_id is defined in account_statement_import_base.
partner_name, account_number and is_reconciled are defined in the "account" module but the account module doesn't display it in the form view. As those fields are very useful for technical audit of bank statement import (along with the field raw_data), I decided to create a notebook on the view, where the second tab is called "technical info" and it displays those fields. That way, the form view is not over-charged with technical fields which are only useful for technical audit in case of problem.

account_statement_import_base/views/account_bank_statement_line.xml
alexis-via: I will take a screenshot before the PR and after the PR, and you will see that the form view is way better than before.

account_statement_import_base/views/account_bank_statement_line.xml
pedrobaeza: But this is going a bit out of the scope. You see it as "I improve the usability a lot", meanwhile others like me see it as "this is breaking compatibility + not related with the change we are handling".

alexis-via:
**Statement line form view**

Before the PR:

before-PR

Scary, isnt' it ?

After PR:
1) first tab:

after_first_tab

2) second tab:

after-second_tab

I re-organised the form view of account.bank.statement.line because I thought that the form view defined in the "account" module was really too bad and it required a re-organisation.

alexis-via:
But, if the reviewers are against the re-organisation of the bank statement form view, I'll leave the ugly view as-is in OCA (and I'll improve it in one of my modules).

account_statement_import_base/views/account_bank_statement_line.xml
alexis-via: Please show me exactly where you think I "break compatibility". I didn't move any field, I set the initial field as invisible and display it elsewhere AFTER the initial position. So, if you added a field via an inherit written <field name="narration" position="after">xxx</field>, then your added field would still be at the same place (after the field which is now invisible).

 I don't see how adding the 3 fields partner_name, account_number and is_reconciled would break compatibility.

pedrobaeza: There are a lot of ways to break compatibility with this kind of changes. Examples:

- A form can only contains one notebook, and another module adding other notebook will have unpredictable behavior.
- Other modules can hook into any of the changed fields, so the fields won't be placed where wanted. This is not exactly breaking, but also not desired.
- Hiding fields will get mad someone that is expecting and using such fields for other purpose that you haven't predicted.
- In general, any change over the base visually speaking may be limited to the minimum.

bosd:
> But, if the reviewers are against the re-organisation of the bank statement form view, I'll leave the ugly view as-is in OCA (and I'll improve it in one of my modules).

For What It's worth, I like the improved view.

njeudy:
I tested this new proposal and I think this is a good way to go (to keep in sync file on sync import) ! thanks @alexis-via 

**Note:** I prefer the new bank statement line form, simplier to use for non technical user and technical data accessible if needed !

NL66278:
I also like the new view much more. But if it would be blocking to add the notebook, an alternative might be to only show the technical data in debug / technical mode.

account_statement_import/wizard/account_statement_import_view.xml
```diff
@@ -21,11 +21,11 @@
                     <footer>
                         <button
                         name="import_file_button"
-                        string="Import"
+                        string="Import and View"
                         type="object"
                         class="btn-primary"
-                        context="{'return_regular_interface_action': True}"
-                    />  <!-- The context may be temporary... waiting for the port of the reconcile interface -->
+                    />
```
NL66278, line 27: This indent seems off

NL66278 (changes requested):
:+1: Definitely a great improvement, but we do need the migration script for the raw data due to the field rename.

alexis-via:
@NL66278 Sorry for the delay; the migration script is provided now

NL66278 (approved):
:+1: LGTM Thanks for streamlining online and offline imports!

alexis-via:
I think we are ready to merge now. @pedrobaeza are you ok with the modification of the view ? The other reviewers were positive about the new form view of statement lines.

account_statement_import_base/README.rst
alexey-pelykh, line 1: @pedrobaeza would really the absence of the file trigger a red CI?

pedrobaeza: It can be

account_statement_import_base/models/account_journal.py
alexey-pelykh: IDK, even `dict` would've been odd there. Sounds like `_prefill_statement_line_import_data` to be..

PR #484, opened 2022-08-08
[13.0][UPD] Repo template

alexey-pelykh:


.copier-answers.yml
```diff
@@ -9,9 +9,8 @@ odoo_version: 13.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups: []
-repo_description: This repository hosts additionnal parsers and import features for
-    bank statements.
-repo_name: OCA bank statement import modules for Odoo
+repo_description: 'TODO: add repo description.'
+repo_name: bank-statement-import
```
sbidoul, line 13: This change does not look right.

alexey-pelykh: Agreed, I've taken the file from https://github.com/OCA/bank-statement-import/blob/15.0/.copier-answers.yml yet there is was not great as well

sbidoul: To update the template, run `pipx run copier -f update`: it will preserve the previous answers.

PR #490, opened 2022-08-18
[IMP] account_bank_statement_import_txt_xlsx: Import lines to existing bank statement

zaoral:


alexey-pelykh:
I don't think such feature should be implemented at the level of individual file parser. Take a look at https://github.com/OCA/bank-statement-import/tree/12.0/account_bank_statement_import_split maybe it already solves the case that you need.

JordiBForgeFlow:
I agree with @alexey-pelykh that it would be best to add it to https://github.com/OCA/bank-statement-import/tree/13.0/account_bank_statement_import_split.

However in my opinion there's a conflict in: https://github.com/OCA/bank-statement-import/tree/13.0/account_bank_statement_import_split.

In my opinion the decision to group the statements to daily, weekly, monthly should be stored in the journal, similar to how you can now define it in online statement import https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import_online/models/online_bank_statement_provider.py#L78

I will send a PR.

andreampiovesana:
name with version please

PR #493, opened 2022-09-03
[12.0] [IMP] *_online_ponto: reliable and efficiently get data from Ponto

NL66278:
There was/is a problem with the Ponto interface that the online import is organised around periods (date since and date until), while Ponto does not offer retrieval of data by date, but instead uses a pointer based on an id present in data that was previously retrieved if any.

This PR, adds extensive debug logging, but also unties the retrieval of data from the Ponto API, from the subsequent translation in imported bank statement lines. Basically at the start of any pull request, any data not yet retrieved from Ponto will be retrieved and stored. After that the creation of statements will be, as now, based on the date since and date until logic.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,7 +38,6 @@ repos:
     rev: v3.4.1
     hooks:
       - id: flake8
-        language_version: python3.6
         name: flake8 excluding __init__.py
         exclude: __init__\.py
   - repo: https://github.com/pre-commit/mirrors-pylint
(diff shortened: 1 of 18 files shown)


ntsirintanis (approved):

rafaelbn (approved):
LGTM 👍 Thanks!

You don't need to update version, Odoobot make it! 😄

NL66278:
@StefanRijnhart Would you have some time to look at and merge the approved and ready PR's like this one?

PR #498, opened 2022-09-26
[12.0] import online improvements

NL66278:
refactor for easier override and maintenance.

### DIFF
--- a/account_bank_statement_import_online/models/account_journal.py
+++ b/account_bank_statement_import_online/models/account_journal.py
@@ -9,48 +9,44 @@ _logger = logging.getLogger(__name__)
 
 
 class AccountJournal(models.Model):
-    _inherit = 'account.journal'
+    _inherit = "account.journal"
 
     online_bank_statement_provider = fields.Selection(
         selection=lambda self: self.env[
-            'account.journal'
+            "account.journal"
         ]._selection_online_bank_statement_provider(),
     )
     online_bank_statement_provider_id = fields.Many2one(
-        string='Statement Provider',
-        comodel_name='online.bank.statement.provider',
-        ondelete='restrict',
+        string="Statement Provider",
+        comodel_name="online.bank.statement.provider",
+        ondelete="restrict",
         copy=False,
     )
 
     def __get_bank_statements_available_sources(self):
         result = super().__get_bank_statements_available_sources()
-        result.append(('online', _('Online (OCA)')))
+        result.append(("online", _("Online (OCA)")))
         return result
 
     @api.model
     def _selection_online_bank_statement_provider(self):
-        return self.env[
-            'online.bank.statement.provider'
-        ]._get_available_services() + [('dummy', 'Dummy')]
+        return self.env["online.bank.statement.provider"]._get_available_services() + [
+            ("dummy", "Dummy")
+        ]
 
     @api.model
     def values_online_bank_statement_provider(self):
-        res = self.env[
-            'online.bank.statement.provider'
-        ]._get_available_services()
-        if self.user_has_groups('base.group_no_one'):
-            res += [('dummy', 'Dummy')]
+        res = self.env["online.bank.statement.provider"]._get_available_services()
+        if self.user_has_groups("base.group_no_one"):
+            res += [("dummy", "Dummy")]
         return res
 
     @api.multi
     def _update_online_bank_statement_provider_id(self):
-        OnlineBankStatementProvider = (
-            self.env['online.bank.statement.provider']
-        )
-        for journal in self.filtered('id'):
+        OnlineBankStatementProvider = self.env["online.bank.statement.provider"]
+        for journal in self.filtered("id"):
             provider_id = journal.online_bank_statement_provider_id
-            if journal.bank_statements_source != 'online':
+            if journal.bank_statements_source != "online":
                 journal.online_bank_statement_provider_id = False
                 if provider_id:
                     provider_id.unlink()
@@ -61,39 +57,39 @@ class AccountJournal(models.Model):
             if provider_id:
                 provider_id.unlink()
             journal.online_bank_statement_provider_id = (
-                OnlineBankStatementProvider.create({
-                    'journal_id': journal.id,
-                    'service': journal.online_bank_statement_provider,
-                })
+                OnlineBankStatementProvider.create(
+                    {
+                        "journal_id": journal.id,
+                        "service": journal.online_bank_statement_provider,
+                    }
+                )
             )
 
     @api.model
     def create(self, vals):
         rec = super().create(vals)
-        if 'bank_statements_source' in vals \
-                or 'online_bank_statement_provider' in vals:
+        if "bank_statements_source" in vals or "online_bank_statement_provider" in vals:
             rec._update_online_bank_statement_provider_id()
         return rec
 
     @api.multi
     def write(self, vals):
         res = super().write(vals)
-        if 'bank_statements_source' in vals \
-                or 'online_bank_statement_provider' in vals:
+        if "bank_statements_source" in vals or "online_bank_statement_provider" in vals:
             self._update_online_bank_statement_provider_id()
         return res
 
     @api.multi
     def action_online_bank_statements_pull_wizard(self):
-        provider_ids = self.mapped('online_bank_statement_provider_id').ids
+        provider_ids = self.mapped("online_bank_statement_provider_id").ids
         return {
-            'name': _('Online Bank Statement Pull Wizard'),
-            'type': 'ir.actions.act_window',
-            'res_model': 'online.bank.statement.pull.wizard',
-            'views': [[False, 'form']],
-            'target': 'new',
-            'context': {
-                'default_provider_ids': [(6, False, provider_ids)],
-                'active_test': False,
+            "name": _("Online Bank Statement Pull Wizard"),
+            "type": "ir.actions.act_window",
+            "res_model": "online.bank.statement.pull.wizard",
+            "views": [[False, "form"]],
+            "target": "new",
+            "context": {
+                "default_provider_ids": [(6, False, provider_ids)],
+                "active_test": False,
             },
         }
--- a/account_bank_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_bank_statement_import_online/models/online_bank_statement_provider.py
@@ -61,32 +61,8 @@ class OnlineBankStatementProvider(models.Model):
         required=True,
         readonly=True,
     )
-    interval_type = fields.Selection(
-        selection=[
-            ("minutes", "Minute(s)"),
-            ("hours", "Hour(s)"),
-            ("days", "Day(s)"),
-            ("weeks", "Week(s)"),
-        ],
-        default="hours",
-        required=True,
-    )
-    interval_number = fields.Integer(
-        string="Scheduled update interval",
-        default=1,
-        required=True,
-    )
-    update_schedule = fields.Char(
-        string="Update Schedule",
-        compute="_compute_update_schedule",
-    )
     last_successful_run = fields.Datetime(
         string="Last successful pull",
-    )
-    next_run = fields.Datetime(
-        string="Next scheduled pull",
-        default=fields.Datetime.now,
-        required=True,
     )
     statement_creation_mode = fields.Selection(
         selection=[
@@ -115,11 +91,6 @@ class OnlineBankStatementProvider(models.Model):
             "UNIQUE(journal_id)",
             "Only one online banking statement provider per journal!",
         ),
-        (
-            "valid_interval_number",
-            "CHECK(interval_number > 0)",
-            "Scheduled update interval must be greater than zero!",
-        ),
     ]
 
     @api.model
@@ -144,26 +115,7 @@ class OnlineBankStatementProvider(models.Model):
             )[0][1]
 
     @api.multi
-    @api.depends("active", "interval_type", "interval_number")
-    def _compute_update_schedule(self):
-        for provider in self:
-            if not provider.active:
-                provider.update_schedule = _("Inactive")
-                continue
-
-            provider.update_schedule = _("%(number)s %(type)s") % {
-                "number": provider.interval_number,
-                "type": list(
-                    filter(
-                        lambda x: x[0] == provider.interval_type,
-                        self._fields["interval_type"].selection,
-                    )
-                )[0][1],
-            }
-
-    @api.multi
     def _pull(self, date_since, date_until):
-        is_scheduled = self.env.context.get("scheduled")
         for provider in self:
             statement_date_since = provider._get_statement_date_since(date_since)
             while statement_date_since < date_until:
@@ -173,19 +125,11 @@ class OnlineBankStatementProvider(models.Model):
                 statement_date_until = (
                     statement_date_since + provider._get_statement_date_step()
                 )
-                try:
-                    data = provider._retrieve_data(
-                        statement_date_since, statement_date_until
-                    )
-                except:
-                    # For scheduled pull, continue with next provider.
-                    if is_scheduled:
-                        break
-                    raise
+                data = provider._retrieve_data(
+                    statement_date_since, statement_date_until
+                )
                 provider._process_data(data, statement_date_since, statement_date_until)
                 statement_date_since = statement_date_until
-            if is_scheduled:
-                provider._schedule_next_run()
 
     @api.multi
     def _retrieve_data(self, statement_date_since, statement_date_until):
@@ -198,27 +142,11 @@ class OnlineBankStatementProvider(models.Model):
             return data
         except:
             _logger.warning(
-                'Online Bank Statement self "%s" failed to'
-                " obtain statement data since %s until %s"
-                % (
-                    self.name,
-                    statement_date_since,
-                    statement_date_until,
-                ),
+                _("Provider %s failed to obtain statement data since %s until %s"),
+                self.name,
+                statement_date_since,
+                statement_date_until,
                 exc_info=True,
-            )
-            self.message_post(
-                body=_(
-                    "Failed to obtain statement data for period "
-                    "since %s until %s: %s. See server logs for "
-                    "more details."
-                )
-                % (
-                    statement_date_since,
-                    statement_date_until,
-                    escape(str(exc_info()[1])) or _("N/A"),
-                ),
-                subject=_("Issue with Online Bank Statement self"),
             )
             raise
 
@@ -243,7 +171,6 @@ class OnlineBankStatementProvider(models.Model):
         ):
             # Continue with next possible statement.
             return
-        statement = self._get_statement(statement_values, statement_date)
         self._update_statement_balance_values(
             statement_values,
             lines_data,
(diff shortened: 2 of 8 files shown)


account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -161,158 +162,206 @@ def _compute_update_schedule(self):
...
+        statement_values.update(
+            {
+                "line_ids": [[0, False, line] for line in filtered_lines],
+            }
+        )
+        statement.write(statement_values)
+
+    @api.multi
+    def _get_statement(self, statement_values, statement_date):
+        """Search existing or makr new statement."""
```
alexey-pelykh: typo

NL66278: Corrected

NL66278:
@thomaspaulb Please have a look

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -141,184 +118,237 @@ def _compute_name(self):
...
+                    "since %s until %s: %s. See server logs for "
+                    "more details."
+                )
+                % (
+                    statement_date_since,
+                    statement_date_until,
+                    escape(str(exc_info()[1])) or _("N/A"),
+                ),
+                subject=_("Issue with Online Bank Statement self"),
+            )
```
thomaspaulb: This message post will be rolled back, because it is directly followed by an uncaught `raise`. Plus, after the `raise`, it will stop, and not continue with the next statement. 

If you want to do it that way, you should create a new `Environment` for the `message_post`, and then raise.

But shouldn't you do it like the original does: log and continue in case of `self.env.context.get("scheduled")`, and stop and raise to the user in case of a manual action?

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -141,184 +118,237 @@ def _compute_name(self):
...
+        return filtered_lines
 
     @api.multi
-    def _schedule_next_run(self):
+    def _update_line_values(self, lines_data):
+        """Enhance data in lines."""
         self.ensure_one()
-        self.last_successful_run = self.next_run
-        self.next_run += self._get_next_run_period()
+        povider_tz = timezone(self.tz) if self.tz else utc
```
thomaspaulb: This is a typo

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -394,43 +424,38 @@ def _sanitize_bank_account_number(self, bank_account_number):
...
+        _logger.info(_("Scheduled pull of online bank statements..."))
+        today = datetime.today()
+        daystart = datetime(
+            year=today.year,
+            month=today.month,
+            day=today.day,
+            hour=0,
+            second=0
         )
+        providers = self.search([("active", "=", True)])
```
thomaspaulb: Doesn't Odoo do that by default?

account_bank_statement_import_online/models/online_bank_statement_provider.py
thomaspaulb: Ah no I'm wrong, you're catching it in the `_scheduled_pull` wrapper function

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -141,184 +118,237 @@ def _compute_name(self):
...
+        2020-02-02 the balance was 50.00. So the amounts for lesser dates should
+        be added to the balance_start. Similarly amounts from later dates should
+        be subtracted from balance_end_real. Which should be 125.00 for
+        2020-02-02.
+        """
+        for line_values in lines_data:
+            date = self._get_utc_line_date(line_values)
+            date = date.astimezone(utc).replace(tzinfo=None)
+            # Add amounts from dates before statement to balance_start.
+            if date < statement_date_since:
```
thomaspaulb, line 259: Is this correct, timezone-wise? If I have a statement from '2020-02-03 00:01', CEST, then in  UTC it will be yesterday 22:01, and it will be lower than the statement_date_since which is `2020-02-04`, so it will be counted as the day before, but that's not correct.

Shouldn't `get_utc_line_date` be replaced with a function `get_line_date` which always returns a naive date without a time component, as the date was in the Bank's timezone, which can then be used to determine which date the statement should get?

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -127,197 +104,251 @@ def _get_available_services(self):
...
+        statement_values: {
+            "balance_start": 25.00,
+            "balance_end_real": 150.00,
+        }
+        lines_data = [
+            {"date": "2020-02-01", "amount": 25.00},
+            {"date": "2020-02-02", "amount": 25.00},
+            {"date": "2020-02-02", "amount": 50.00},
+            {"date": "2020-02-03", "amount": 25.00},
+        ]
```
thomaspaulb, line 247: This would also work well as a unit test.

It's unclear to me though - are you modifying existing bank statements, or are you always creating new ones for each date? Because I don't see you loading the balances of existing statements anywhere.

account_bank_statement_import_online/models/account_journal.py
```diff
@@ -9,48 +9,44 @@
 
 
 class AccountJournal(models.Model):
-    _inherit = 'account.journal'
+    _inherit = "account.journal"
```
thomaspaulb, line 12: I guess it's a matter of taste, but I see some drawbacks in sneaking in `black` runs into regular PR's. I guess your aim was to be able to cherry-pick the contentful commits easily in 13.0 and 14.0 as well. That's an argument, but I'd rather do `black` only in version-to-version migrations.

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -394,43 +424,38 @@ def _sanitize_bank_account_number(self, bank_account_number):
...
+            for provider in providers:
+                date_since = provider.last_successful_run or daystart
+                try:
+                    provider.with_context(
+                        {"tracking_disable": True}
+                    )._pull(date_since, date_until)
+                    provider.last_successful_run = date_until
+                except:
+                    # Continue with next provider.
+                    _logger.warn(
```
thomaspaulb, line 442: Maybe add a context key of `simulate_failure`, which throws an exception somewhere, and then we can make a test case for this too.

thomaspaulb:
@NL66278 Is this just a refactor, or is it also a fix of some kind? And if yes, what does it fix?

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: Well it will be rollbacked for non scheduled sessions. The log entry will still be made. This is very usefull as the end-user receiving a message will often dismiss it without saving. For support it is still in the log then. I will move the message post so it is only done for scheduled pull's.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: Will correct.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: I tried to keep the timezone logic as it was before.

However correct handling of timezones is quite difficult:,
All datetime fields in the database are assumed to be in UTC format.
The bank can be in one timezone,
The server can be in a second timezone,
The user can be in a third timezone
The company holding the accounts can be in a fourth.

Best to convert all incoming datetimes as soon as possible to UTC.

However when considering which date (as opposed to datetime) to use for the bank statement, we have IMHO to look at the timezone for the company that holds the account. The res.company does not have that atttribute at the moment, so we would need to add it.

Or we can simply use the timezone of the bank, if this differs from the timezone of the company, the meaning of the statement date will be the date in the timezone of the bank. The datetime of the line will still need to be stored as utc.

Will check the code for this.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: Loading the existing balances is part of the standard Odoo code. I had the same amazement.

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: Instead of making guesses which is better, why not just to let (which was already a feature IIRC) user configure which TZ to convert to?

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: I will modify existing statements if there are open ones for the right date. Same as original code, I just added the explanation once I understood what was going on.

account_bank_statement_import_online/models/account_journal.py
NL66278: I did the black in a separate commit, otherwise would be very difficult to forward migrate these improvements. And the functional changes can be studied on their own.

account_bank_statement_import_online/models/online_bank_statement_provider.py
thomaspaulb: @NL66278 Yes we should convert to UTC for datetimes, but bank statements are tied to dates, so we have to make a choice somewhere. Configurable is a good idea. I don't know where that was implemented but since @alexey-pelykh is the original author it must be somewhere :-) My guess would be on the provider record.

NL66278:
The main fix is that the separate scheduling was flawed. The root cause of many of the problems of some of hour customers. So basically statements would be retrieved and created until the next_run date. If the scheduling used steps of an hour, the next_run date would be increased by an hour. This assumed that the code would always run, day and night without exception, no downtime at all. But any missed run would put the next_run an hour back compared to the real time. Of course this could be manually fixed, but studying it I could not think of any usecase where it was needed to duplicate the scheduling options already in ir.cron. You just have to note the time at the start of a run, and use that as your launching time on the next run. Then every moment will be covered.

Then I also solved the problem of creating duplicate empty statements (if an earlier statement was already validated, and newly retrieved data did not contain new tranactions). I added the option (by default off to keep existing functionality) to prevent empty statements altogether.

account_bank_statement_import_online/models/online_bank_statement_provider.py
thomaspaulb: Fair enough. If you have the time, you could make this example into a unit-test and make the module more insightful

account_bank_statement_import_online/models/account_journal.py
thomaspaulb: OK

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: I will add a configuration option then.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: Yes, but was also already there. Will remove it.

account_bank_statement_import_online/models/online_bank_statement_provider.py
thomaspaulb: Isn't it [this one](https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import_online/models/online_bank_statement_provider.py#L39)

thomaspaulb: Or i mean, [this one](https://github.com/OCA/bank-statement-import/blob/12.0/account_bank_statement_import_online/models/online_bank_statement_provider.py#L49), in 12.0

NL66278: The thing is, we need two timezone fields. One for the timezone of the bank. One fr the timezone of the company (to be set on the provider).  Unless the dates in the data from the bank are timezone aware.

thomaspaulb: Ahhh yes, you're right. So I guess it needs `source_tz` and `tz` (the latter being `dest_tz`)

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -291,17 +301,40 @@ def _update_statement_balance_values(
...
+            # Add amounts from dates before statement to balance_start.
             if date < statement_date_since:
                 if "balance_start" in statement_values:
                     statement_values["balance_start"] = Decimal(
                         statement_values["balance_start"]
                     ) + Decimal(line_values["amount"])
                 continue
-            elif date >= statement_date_until:
+            # Subtract amounts from dates after statement from balace_end_real.
+            if date >= statement_date_until:
```
thomaspaulb, line 266: Why change this `elif` into an `if`? That would mean you expect that it could be that `statement_date_until < statement_date_since`, and that's something I would rather add an `assert` for that it does not happen.

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -291,17 +301,40 @@ def _update_statement_balance_values(
...
+            {"date": "2020-02-02", "amount": 25.00},
+            {"date": "2020-02-02", "amount": 50.00},
+            {"date": "2020-02-03", "amount": 25.00},
+        ]
+        If we create a statement for only the day 2020-02-02, the first and
+        the last line should not belong to this statement. But at the start of
+        2020-02-02 the balance was 50.00. So the amounts for lesser dates should
+        be added to the balance_start. Similarly amounts from later dates should
+        be subtracted from balance_end_real. Which should be 125.00 for
+        2020-02-02.
```
thomaspaulb, line 253: Okay, I get this... but:

- Let's say yesterday we had statement A
- Today, we have statement B, but during fetching of transactions for statement B, some transactions for end of yesterday still come through.
- We couple these now to statement B, but we add the items from yesterday to the starting balance

That's all fine and merry, but who then updates the end balance of yesterday? That has also shifted because of the new transactions being there.

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -315,11 +348,24 @@ def _update_statement_balance_values(
...
         AccountBankStatementLine = self.env["account.bank.statement.line"]
         filtered_lines = []
         for line_values in lines_data:
+            # Do not include transactions for dates not in statement period.
+            date = self._get_utc_line_date(line_values)
+            date = date.astimezone(utc).replace(tzinfo=None)
+            if date < statement_date_since:
+                continue
+            if date >= statement_date_until:
+                continue
```
thomaspaulb, line 296: Looking at this again - isn't this dangerous, though? Example:

1. We have loaded stuff from day A on statement A
2. We are now in day B, but we fetch still some transactions from day A also, that are unknown in our system
3. With this new code, we don't load those, because our statement starts in day B and so we skip loading these transactions, and they're not in the system at all

account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -232,41 +240,47 @@ def _process_data(self, data, statement_date_since, statement_date_until):
...
+            statement_date_since,
+            statement_date_until
         )
-        statement.write(statement_values)
+        self._update_line_values(filtered_lines)
+        self._create_statement(statement_values, statement_date, filtered_lines)
 
     @api.multi
-    def _get_statement(self, statement_values, statement_date):
+    def _create_statement(self, statement_values, statement_date, lines_data):
```
thomaspaulb, line 189: Looking at this again - why the refactor?

Couldn we have kept `_get_statement` intact like it was, but just not call it when `lines_data` is empty, and `self.create_empty_statements` is not set?

account_bank_statement_import_online/models/online_bank_statement_provider.py
thomaspaulb: On second thought: is that really needed? Because if we get statements like: 2022-03-01 00:30, and we say that this should be interpreted as the timezone GMT+3 where the bank is in; then we save the statement lines as the UTC equivalent and they will appear at the right date and time in the user's timezone; and for the statement date we use the bank's timezone and truncate the time. Why do we then still need the company timezone?

alexey-pelykh (changes requested):
My general impression is that this PR involves a lot of refactoring including change of logic without clearly expressed rationale for those changes: why old logic needed change, what was the old behavior, what will be the new one. On top of that, definitely separate the reformatting with refactoring for hooks and refactoring for logic into separate PRs. Reformatting would be quickly merged so we can continue of bigger changes in this one when it's rebased.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: An `elif` combines an else with an if. The else part here is superfluous, but forces you to look at the other if to which the else is connected. So a pure if here does not cause any other funtionality, but makes the code more clear and readable.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: No, the extra transactions for statement A (yesterday) are NOT coupled to statement B. They are added to statement A, causing its end balance to be updated. After this statement B (today) will be done.

Because our last run (yesterday) was before the end of the day, each run (scheduled pull ) will retrieve first the transactions from yesterday, start of day. It will discard the transactions already imported with help of the unique_import_id. Then it will retrieve the transactions for today, as far as already available.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: You are mistaken. See above.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: Well having a `_get_statement` is already part of the refactor. I Try to give each function as clear a role as possible, minimizing data passed. All according to `clean coding` principles. The main constraint is to not break inheriting modules so keeping existing function signatures like for `_pull` and `_obtain_statement_data`.

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: Well if we have a transaction for a company that occuring to its utc timestamp happened 2020-02-01 at 23:00, this  means for the company it occured on 2020-02-02 at 01:00, so should go into a statement for 2020-02-02. This problem becomes bigger the further companies / Odoo customers are from UTC.

thomaspaulb: But that's assuming the company's timezone determines the date of the transactions, whereas I would think the timezone of the bank decides that.

I mean, if you have a bank account in hong kong, the transactions will be divided in statements per day according to the hong kong timezone, right?

NL66278: Wrong. The division of transactions within Odoo over statements (daily, weekly or monthly) has no relation at all with the organization of transactions at the online provider / bank. Actually, the provider might not have organized the transactions into statements at all. (For instance Ponto has no concept of statements). The transactions retrieved might have timezone data in their dates. Or not, then we need a default for the provider. For instande Qonto is supposed to be in Paris. We have to translate the transaction timestamps from the bank, into timestamps for the company / Odoo customer, to determine to which day, week or month they belong.

belisac:
Hello @NL66278  y @alexey-pelykh 

Thank you for the changes and to follow and correct this bug. Meanwhile this finished to merge it, I proceed to test in my Local and I'm still having the same problem when I tried to Pull Movements Since dd-mm-aaaa Until dd-mm-aaaa

I haven't seen the change or correction. Do I have to considered something else in order to get the information done?

Please Let me know as soon as you can, I have couple issues from Clients.

Thanks in advance,

Regards,

NL66278:
@belisac What provider are you using?

belisac:
Hello @NL66278 

Thanks for your answer 
MyPonto

Regards,

NL66278:
The empty statement problem was fixed by backporting 14.0 solution. Will provide other PR for the problem of the missing schedules.

PR #501, opened 2022-10-02
[12.0][FIX] *_online_ponto: limit interactive wait for synchronization

NL66278:
Also correct message for succesfull sync.


NL66278:
We will no longer use synchronization

PR #502, opened 2022-10-02
[14.0][IMP] *_online_ponto: forward port 12.0 improvements

NL66278:
Separate retrieval of data from ponto (buffer data) and creation of statements.


account_statement_import_online_ponto/__manifest__.py
```diff
@@ -1,14 +1,22 @@
...
+# Copyright 2022 Therp BV <https://therp.nl>.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Online Bank Statements: MyPonto.com",
-    "version": "14.0.1.1.0",
+    "version": "14.0.2.0.0",
     "category": "Account",
     "website": "https://github.com/OCA/bank-statement-import",
-    "author": "Florent de Labarre, Odoo Community Association (OCA)",
+    "author": "Florent de Labarre" ", Therp BV" ", Odoo Community Association (OCA)",
```
alexey-pelykh: `" "` should be removed

NL66278: Done!

account_statement_import_online_ponto/models/ponto_buffer.py
```diff
@@ -0,0 +1,80 @@
...
+"""Define model to hold data retrieved from Ponto."""
+import json
+import logging
+
+from odoo import _, fields, models
+
+_logger = logging.getLogger(__name__)
+
+
+class PontoBuffer(models.Model):
```
alexey-pelykh, line 12: @pedrobaeza requesting second opinion - it feels like models should've been prefixed, yet I'm not entirely sure what are the guidelines on that

alexey-pelykh (approved):
I'm generally OK with the code, except some naming style concerns that I'd like to hear second opinion. Otherwise LGTM

rafaelbn (approved):
Thank you! 👍

pedrobaeza:
I would like to know the goals that motivate this refactoring to judge about the huge changes.

NL66278:
@pedrobaeza Most online statement providers provide selection of the data to be downloaded per date. That is not the case with Ponto. You can only retrieve the most recent data, or data before that using an identifier retrieved previously, to get the data before that.
The existing code did not deal with this situation correctly in all instances. Correcting this with minimal changes proved to be complex. So I decided for a refactoring that both solved the functional problems and made the code more readable and maintainable and on top more efficient.

pedrobaeza:
But I don't see such a problem, as if you have the last identifier, you use it, and if not, you put the interval. I'm using this module already 2 years with no problem as you describe. Can you specify a real situation that got you to be stuck that can't be solved manually pulling the interval you want?

NL66278:
@pedrobaeza I have regularly seen that even manually pulling would not fetch missing transactions. Most of my customers have configured to pull every hour. Maybe the problem does not manifest itself, or not so easy, if you pull daily. What do your customers do? How often they pull, and do they use daily or weekly statements?

NL66278:
@alexey-pelykh I will rename the buffer models to give them a more generic name, including a prefix. I think the idea of first storing transactions in a buffer might have more general applicability. So when migrating to new versions  I will probably put this in a separate module and make the Ponto import dependent on that. I hesitate to change the dependency chain in this existing version.

pedrobaeza:
I have also the pulling each hour:

Selección_003

I can't accept this huge refactoring that also means to store a lot of useless transactional information unless a good reason for it.

NL66278:
@pedrobaeza And what scheduling is configured on the journal?

The transaction information will only be stored temporarily. It provides enormously helpful information when things go wrong, or when results of import are not what is expected.

pedrobaeza:
There's no scheduling in journals...

Debug information should be put on logs and when requested, not stored on DB IMO.

NL66278:
@pedrobaeza The scheduling, and statement frequency are configured through the provider information via the journals. See: https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import_online/models/online_bank_statement_provider.py#L53

pedrobaeza:
OK, on the provider...

It's also hourly:

Selección_006

alexis-via:
This module needs a cleanup/refactoring of the code and more debugging capabilities. For example, this piece of code needs a full re-write:
https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py#L135
(and I see that this piece of code has been re-written in this PR !)

I have been deploying this module in 2022 on 2 Odoo v14 in production, and, although it works well most of the time, it regularly fails to download some ponto transactions, and it's very difficult to understand the reason and debug it. This kind of problem happen 2 or 3 times per quarter.

Also, I talked about this module with Florent de Labarre (the initial author of the module) two months ago. He told me that he doesn't use it any more and doesn't maintain it.

So I think a refactoring is really needed. I'll test/review this PR.

pedrobaeza:
The eventual errors that happen with Ponto is due to communication problems, so storing what can be communicated twice (on the statement lines and this new model) is redundant.

NL66278:
@pedrobaeza The main point of the buffers is not the communication information. But to detach the retrieval of data from the distribution over statements. This has to to do with the mismatch between the Online modules trying to retrieve data per statement period (day, week or month) and the fact that Ponto does not allow data retrieval with a date (or period) selection. That the buffers provide a convenient way to look at the communication info is only a usefull, at least in my opinion useful, side-effect.

pedrobaeza:
Adding such buffer only complicates things and can provoke awkward situations like relying on local buffer while it's not complicated, among the unneeded increase of space, so I'm against such buffer, sorry.

I don't see either what is the problem with the retrieval of data as it's now.

NL66278:
@pedrobaeza The updated code is much easier to understand then the original code. So it escapes me how this complicates things. Fact is that several clients regularly miss data without the changes and have everything working smoothly with the changes.

pedrobaeza:
As said, it complicates the data model and put more things in the middle to fail. Please don't add the intermediate buffer.

account_statement_import_online_ponto/models/ponto_buffer_line.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+"""Define model to hold transactions retrieved from Ponto."""
+from odoo import fields, models
+
+
+class PontoBuffer(models.Model):
```
alexis-via: Class name should be PontoBufferLine

alexis-via:
BTW, the new model ponto.buffer + ponto.buffer.line has already landed in the 12.0 branch via this (merged) PR for 12.0:
https://github.com/OCA/bank-statement-import/pull/493/files

I don't have a strong opinion on this new model. Maybe it's overkill, but I suppose that @NL66278 had a good reason to add this new model. I am not very familiar with this module and not with account_statement_import_online, so I don't know.

pedrobaeza:
The fact that they pushed that on an old branch that I didn't monitor doesn't make this as something mandatory to be merged now on this branch, and more having the 13.0 version (which is one of the versions I'm using among 14.0) without it.

alexis-via:
I'm working on 2 things at the moment on this PR (my goal is to submit a PR on this PR):
1) make date field configurable. At the moment, we always use effectiveDate as the date for the statement line. But, on some bank, it may be better to use the valueDate as the date for the statement line. So I'm working on adding a configuration parameter on online.bank.statement.provider. But the impact is bigger than I thought, because it would imply to rename 1 field on ponto.butter (effective_date -> date) and ponto.buffer.line (effective_date_time -> date_time). Are you ok with this @NL66278 ?
2) the download of Ponto transactions into Odoo takes quite some time, which is surprising for such simple data that is already available on Ponto servers. I see that, when we download Ponto transactions, we call the Ponto API "/synchronizations", which, if I understand correctly, asks Ponto to re-sync with the bank. I wonder if calling "/synchronizations" is really necessary. Ponto already gets transactions data from banks regularly, I don't think we need to "force" it to resync everytime we want to download Ponto transaction. In fact, this point was my suspect n°1 for the bug of missing transactions between Ponto and Odoo: my idea was the following: when we want to download Ponto transactions, we first call "/synchronizations", which make Ponto resync with all the banks... if one bank takes time to answer (or have API problems at that moment), then the call to "/synchronizations" fails or timeouts and then we don't go further and fetch the Ponto transactions. I would be interested to have your opinion on it. In the short term, I'll propose to add a boolean option "ponto_force_sync" on online.bank.statement.provider and see how it behaves in real life with real bank accounts. What do you think @NL66278 ?

alexis-via:
On the Ponto API specs (https://documentation.myponto.com/1/api/curl#create-synchronization), they say:

**Create Synchronization**
_Important: Synchronizations must always be created in the presence of the Ponto user. For this reason the API call requires the real IP address of the currently present customer. Using a fake IP address or executing this call without the presence of the customer is considered as an infringement of our T&Cs and may lead to the blocking of your integrations/organizations._

In tends to confirm my idea that we should not call /synchronisations when we download ponto transactions.

NL66278:
@alexis-via I have no problem with dropping the synchronization from the download. Neither with changing the date field.

I think the idea behind the synchronization was to make sure that the data at Ponto  would be up to date. In the original code this was not working at all.  I corrected this, but then found the triggered synchronization did not complete in a reasonable amount of time.

pedrobaeza:
I think Alexis is digging on the root cause, but having the buffers is not going to improve it, so the refactoring should go on calling properly the API.

It's also worth noting about a notification I received about some change in the API:
```
Dear Ponto user,

With this email we would like to announce a change in our API.

    The "customerIpAddress" becomes a mandatory parameter when your software creates a synchronization resource in Ponto (https://documentation.myponto.com/api/curl#create-synchronization).

The presence of this parameter will demonstrate that you are actively using the service at the time of the request, while the absence thereof is perceived by the banks as the evidence that the user is not actively requesting the information at the time of access to the bank account , in the sense of article 36.5 (b) of the Regulatory Standards of Strong Customer Authentication (‘RTS SCA’). In that latter case, the access request is considered as unattended. The RTS SCA only authorizes four (4) unattended access requests per period of 24 hours.

If your software makes use of Ponto custom integrations, your developer must adapt your software to cope with the new "customerIpAddress" parameter. If you are not using Ponto custom integrations, you don't need to do anything since we have already contacted your software provider.

We understand that your developer will need time to implement these changes. Therefore, the "customerIpAddress" parameter will remain optional until the 15th of June 2023. During this transition period, your software will have to be adapted and you will be able to choose the date at which you want to switch.

On the other hand, Ponto is now synchronizing your bank account data four (4) times per day in line with the RTS SCA requirements (at 2am, 8am, 2pm and 8pm). Therefore, you may choose to have your software simply stop synchronizing your bank accounts without user intervention.

We have also updated our Terms and Conditions for Ponto service, in order to reflect the provisions of the RTS SCA on bank account data synchronizations, by adding a new clause 3.7. ‘Bank Account Data Synchronizations’. The new provision prohibits the creation of automated synchronization by your software if the user is not actively requesting information at that moment. This means that implementing automated synchronization patterns will no longer be allowed in Ponto.

The amended terms and conditions are available on https://myponto.com/en/terms-and-conditions/v3.0 and will be effective as from the 15th of June 2023.
```

NL66278:
@alexis-via I removed the synchronization.

alexis-via:
@pedrobaeza Thanks for sharing this email that contains very important infos. With this information, I think I'll just remove the call of the API /synchronisations and not make it a configuration option.

About the debate on the "buffer" object, I don't have a strong opinion on this. I understand the arguments of each side. In the short term, my code will be based on this PR with the buffer object.

NL66278:
@pedrobaeza @alexis-via I implemented an alternative approach without buffer here: https://github.com/OCA/bank-statement-import/pull/542

All tests are running fine, but I also tested from a live Ponto environment to retrieve a large amount of data (from mid november 2022) without any problem.

The only disadvantage,, compared with the buffered approach is that it will retrieve the same data several times if the cron job has not run for some time, but this does not lead to actual problems and was the same before the introduction of the buffer.

alexis-via:
Here is my PR to improve this PR: https://github.com/Therp/bank-statement-import/pull/1

I will probably not contribute any more on this, because I eventually decided to re-write  the module "account_statement_import_online" to simplify it (and improve it too). It's a bit early to explain all the ideas I have and they will certainly evolve over time, but I'll just give few of them that should not change:
1) remove the option "statement_creation_mode" on the provider. Working today on this code reminded my that, one of my customer using Ponto noticed that the failures often happen on the last day of the month... and he is configured with statement_creation_mode == 'monthly' and I now think that it is linked. Anyway, when you start processing the bank statement, the next ponto download will create a new one, so this config parameter is not a "must have".
2) a provider should be linked to journals via O2M and not M2O. If you have several bank accounts handled with ponto, you should not have to create several providers!
3) use more date fields vs datetime fields in the code.
4) have a cron per provider

I plan to call my new modules "account_statement_import_api" and "account_statement_import_api_ponto". I also plan to have "account_statement_import_api_qonto". I plan to maintain it in an Akretion github repo. If I succeed in the implementation and it significantly simplifies the code, maybe I'll propose it to OCA for a future version. Otherwise, I'll just continue to maintain it in the Akretion repo.

alexey-pelykh:
@alexis-via I take it that the improvements that you've listed are limited to Ponto experience only? Since I find it confusing if extended to other providers where having a provider per bank accounts is a necessity to handle per-account configuration. Obviously, that could've been managed by single O2M provider and per-account O2O provider configuration yet that would've added more complexity with effectively same result.

NL66278:
@alexis-via I would find it a loss if we would have alternative repositories, instead of pooling our ideas and resources. I highly value your input and contributions.

A few other remarks:
* The decision to group statements per week, per day or per month, based on either transaction date, or based on execution date might be more general then just for the Ponto import. Maybe we could have this even configured in the base module.
* On the missing transactions. Specifically for Ponto we might use the fact that all transaction id's are always in the same order. So we might readback from the last transaction until we come to an id already stored on a bank statement line.
* Question on the idea of removing the statement_creation_mode. Do you then always want to have daily statements? I think it is nice for the customer to have this choice. Even more, I thought it might be useful as an option even for transactions imported through the file import. In the modern era the whole idea of "statements" is a little bit, or rather a lot, outdated. What we have are transactions, that we like to group by day, week or month (or by downloaded file) just for convenience.

alexey-pelykh:
I support what @NL66278 said on grouping transactions - we must have flexibility on how we group statements.

pedrobaeza:
Yes, but I don't know why always Alexis want to fork everything. The features you are commenting can be added as an option on the selection on how to group: "No group", "Daily", "Weekly"...

alexis-via:
I agree with @NL66278 that the concept of account.bank.statement is outdated: we could work without account.bank.statement and use only account.bank.statement.line. In v16, we can have account.bank.statement.line without account.bank.statement. I just discovered that, in v16, when you delete an account.bank.statement, it doesn't delete the account.bank.statement.line (no more ondelete='cascade') !

My main goal is to simplify the code of account_statement_import_online and the api-specific modules to make it easier to understand and debug. I have a bad experience with online import of bank statement in v14, both with ponto and qonto, with too many bugs. If I manage to develop an alternative module that is simpler (and easier to debug), then I'll show it to you and maybe propose it for a future odoo version. Maybe it will be as complex and hard to debug as account_statement_import_online and I will just throw it away ! :)

I need to implement my idea in code to really know if they are good idea or not. But I need to start again from scratch, otherwise my ideas cannot be seriously tested.
I hope to find time in the coming weeks to develop this alternative module, but I'm also busy on porting accounting module to v16...

NL66278:
Closing this in favour of the PR without buffering, that now also includes the configuration proposed by @alexis-via to use either execution date or value date to group transactions.

PR #504, opened 2022-10-04
[14.0][FIX] *_import_online: Just use cron scheduling

NL66278:
Rely on cron scheduling. This solves the problem that when the scheduled pull doest not always run, due to downtime or other caused, the next_run, is more and more behind the current time.

### DIFF
--- a/account_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_statement_import_online/models/online_bank_statement_provider.py
@@ -1,8 +1,10 @@
 # Copyright 2019-2020 Brainbean Apps (https://brainbeanapps.com)
 # Copyright 2019-2020 Dataplug (https://dataplug.io)
+# Copyright 2022 Therp BV <https://therp.nl>.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import logging
+import sys
 from datetime import datetime
 from decimal import Decimal
 from html import escape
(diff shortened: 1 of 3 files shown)


alexey-pelykh:
@NL66278 how I will be able to configure different pull intervals for different providers via cron?

alexey-pelykh (changes requested):
As far as I can understand, this will cause ability to configure different pull intervals for different providers, which is not ok with me. If there's a bug - it needs to be fixed though

NL66278:
@alexey-pelykh Why would you need different intervals for different providers? Nine out of ten customers will have one provider anyway. But less us take for arguments sake the example of a customer that needs all its paypall data very much up to date, so they want to check every one hour. But they also have a ponto account that is not that time critical.  Also checking ponto every hour will not hurt them in the least.

alexey-pelykh:
@NL66278 

> Why would you need different intervals for different providers?

To be able to pull card processing every 10 minutes and regular bank every day, e.g.

> Nine out of ten customers will have one provider anyway.

I assume you've made a representative survey?

> But less us take for arguments sake the example of a customer that needs all its paypall data very much up to date, so they want to check every one hour. But they also have a ponto account that is not that time critical. 

Exactly the point 

> Also checking ponto every hour will not hurt them in the least.

That is an assumption, since some APIs do charge per call.

So what this supposed fix fixes by removing feature?

NL66278:
@alexey-pelykh There is the very real problem that nearly all of the customers that use online import encounter of the next_run falling behind. By the way, maybe not representative, but I do not know of any customer with more then one provider. Maybe your experience is different. What about if we keep the interval per provider, but use it as the minimal elapsed time before retrieving data again. In this way your concern about limiting the calls dependent on provider is taken care off, but when retrieving data in a scheduled run, we are always up to date?

alexey-pelykh:
> There is the very real problem that nearly all of the customers that use online import encounter of the next_run falling behind

I've been using the module for years, haven't noticed such issue. If `next_run` is behind now, it will be executed so eventually it will catch up. So the only possible issue would be too many pulls in sequence that ~can be fixed by ensuring that when setting `next_run` - add interval until it's in future.~ (can not since it will break logic)

Thus, what issue the current behavior is causing for you and how? Servers running Odoo are powered up on demand?

> Maybe your experience is different.

Exactly, my average customer has 2-3 providers with largest having 6 (PayPal + several banks + Wise.com + Braintree)

> What about if we keep the interval per provider, but use it as the minimal elapsed time before retrieving data again. In this way your concern about limiting the calls dependent on provider is taken care off, but when retrieving data in a scheduled run, we are always up to date?

I'm not ready to discuss this part until I understand what the bug is and why it is happening.

NL66278:
I will create an alternative PR that solves the problem while keeping the scheduling per provider as is.

PR #505, opened 2022-10-04
[12.0][IMP] account_statement_import_online: don't import empty statements

NL66278:
Backport from 14.0

### DIFF
--- a/account_bank_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_bank_statement_import_online/models/online_bank_statement_provider.py
@@ -106,6 +106,7 @@ class OnlineBankStatementProvider(models.Model):
     certificate_public_key = fields.Text()
     certificate_private_key = fields.Text()
     certificate_chain = fields.Text()
+    allow_empty_statements = fields.Boolean()
 
     _sql_constraints = [
         (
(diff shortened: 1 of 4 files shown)


account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -106,6 +106,7 @@ class OnlineBankStatementProvider(models.Model):
     certificate_public_key = fields.Text()
     certificate_private_key = fields.Text()
     certificate_chain = fields.Text()
+    allow_empty_statements = fields.Boolean(string="Allow empty statements")
```
alexey-pelykh: `string` is not needed, it will be automatically named like that. Also, you're changing default behavior - the default value should be `True` to keep things unaltered. Also, explaining that in README would be great

alexey-pelykh (changes requested):

account_bank_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: But I stand corrected about the default value due to this https://github.com/OCA/bank-statement-import/pull/408#issuecomment-902509072

NL66278: I deleted the string

alexey-pelykh (approved):

PR #507, opened 2022-10-06
[12.0] schedule improvements

NL66278:
Optionally adjust next_run when it appears scheduled runs have been missed.

### DIFF
--- a/account_bank_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_bank_statement_import_online/models/online_bank_statement_provider.py
@@ -81,6 +81,7 @@ class OnlineBankStatementProvider(models.Model):
     )
     last_successful_run = fields.Datetime(
         string='Last successful pull',
+        readonly=True,
     )
     next_run = fields.Datetime(
         string='Next scheduled pull',
@@ -423,7 +424,8 @@ class OnlineBankStatementProvider(models.Model):
             _logger.info('Pulling online bank statements of: %s' % ', '.join(
                 providers.mapped('journal_id.name')
             ))
-            for provider in providers.with_context({'scheduled': True}):
+            for provider in providers.with_context(scheduled=True):
+                provider._adjust_schedule()
                 date_since = (
                     provider.last_successful_run
                 ) if provider.last_successful_run else (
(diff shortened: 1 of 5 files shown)


account_bank_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -107,6 +108,10 @@ class OnlineBankStatementProvider(models.Model):
     certificate_private_key = fields.Text()
     certificate_chain = fields.Text()
     allow_empty_statements = fields.Boolean()
+    adjust_next_run = fields.Boolean(
```
alexey-pelykh: It's safe to have this always enabled, don't think there's need for an option tbh

account_bank_statement_import_online/tests/test_account_bank_statement_import_online.py
```diff
@@ -267,6 +267,48 @@ def test_interval_type_hours(self):
...
+        provider.interval_type = 'hours'
+        provider.adjust_next_run = True
+        five_minutes_ago = self.now - relativedelta(minutes=5)
+        three_hours_earlier = five_minutes_ago - relativedelta(hours=3)
+        fiftyfive_minutes_later = five_minutes_ago + relativedelta(hours=1)
+        provider.next_run = three_hours_earlier
+        provider.last_successful_run = three_hours_earlier - relativedelta(hours=1)
+        provider._scheduled_pull()
+        self.assertEqual(provider.last_successful_run, five_minutes_ago)
+        self.assertEqual(provider.next_run, fiftyfive_minutes_later)
```
alexey-pelykh, line 288: To be on the super-safe size, please also check that number of pulled statement lines matches the expectation in both cases, whatever that number is

account_bank_statement_import_online/models/online_bank_statement_provider.py
NL66278: I will leave out the boolean

account_bank_statement_import_online/tests/test_account_bank_statement_import_online.py
NL66278: Little bit complicated as I check for hours of delay but statements are created daily, and the number of statements created might depend on the time the tests run (so either 1 or 2 actual dates might be involved).

NL66278:
@alexey-pelykh Can you live with the PR as it is now, or do you think the test for number of statements is absolutely needed?

NL66278:
@alexis-via Could you also have a look at this one please?

alexey-pelykh:
@NL66278 same comment on commit message :)

NL66278:
@alexey-pelykh Also adjusted this commit message.

NL66278:
Closed in favor of more complete update here: https://github.com/OCA/bank-statement-import/pull/590

PR #342, opened 2020-10-16
[13.0][MIG] account_bank_statement_import_qif

Abranes:
- Standard migration to 13.0
- Migration from #215

PR #508, opened 2022-10-17
[13.0][MIG] account_bank_statement_import_qif: Migration to 13.0

victoralmau:
Superseed https://github.com/OCA/bank-statement-import/pull/342

Migration to 13.0

Please @pedrobaeza can you review it?

@Tecnativa TT26069


account_bank_statement_import_qif/wizards/account_bank_statement_import_qif.py
```diff
@@ -82,12 +86,12 @@ def _parse_file(self, data_file):
...
     def _complete_stmts_vals(self, stmt_vals, journal_id, account_number):
         """Match partner_id if hasn't been deducted yet."""
-        res = super(AccountBankStatementImport, self)._complete_stmts_vals(
-            stmt_vals, journal_id, account_number,
-        )
+        res = super()._complete_stmts_vals(stmt_vals, journal_id, account_number)
         # Since QIF doesn't provide account numbers (normal behaviour is to
         # provide 'account_number', which the generic module uses to find
         # the partner), we have to find res.partner through the name
+        if not self.is_qif:
```
pedrobaeza: Please don't create a new field. Use existing method `_check_qif`. Anyway, is this really needed?

victoralmau: This is necessary so that other modules do not execute that code that is only needed in this module. 

Perhaps another possible approach would be to inject it in context here https://github.com/odoo/odoo/blob/13.0/addons/account_bank_statement_import/account_bank_statement_import.py#L50 but I didn't want to do that in order not to repeat code. 
What do you think about this?

pedrobaeza: OK, but it should be a method, not a field.

account_bank_statement_import_qif/wizards/account_bank_statement_import_qif.py
```diff
@@ -18,9 +20,13 @@ class AccountBankStatementImport(models.TransientModel):
     def _check_qif(self, data_file):
         return data_file.strip().startswith(b"!Type:")
 
+    def _check_is_qif(self):
```
pedrobaeza: Why 2 methods? Just use `_check_qif`

pedrobaeza (approved):

pedrobaeza:
I think we can finally merge this one without waiting for a second review, after being all this time queued, and being tested:

/ocabot merge nobump

PR #512, opened 2022-10-21
[14.0][IMP] account_statement_import_txt_xlsx: First line different from 0

ghost:
In the case of the Belgian bank Belfius (and probably others banks), when you export your bank statements in CSV format, they add about ten lines with information about the export (date of export, filters applied,...) so rather than modifying the CSV file each time and removing these lines, I propose that the mapping allows you to determine at which line the file actually starts.


account_statement_import_txt_xlsx/tests/fixtures/sample_statement_en_with_header.csv
```diff
@@ -1,15 +1,15 @@
-Posting date from;;;;;;
```
alexey-pelykh: This file worked before, why you needed to update it?

ghost: I based the one with the headers (sample_statement_en_with_header.csv) on the basic one (sample_statement_en.csv) and also to fit with the sample mapping

PR #517, opened 2022-11-14
[ADD] account_bank_statement_import_txt_xlsx: existing statemnet

zaoral:
Be able to import lines to and existing statement instead of create a new one everytime. This is usefull because if the user group the staments by month they are able to made the importation several times in the same month checking the balance as they have in the bank instead of have different statments on per each import


zaoral:
same change of https://github.com/OCA/bank-statement-import/pull/490 but in version 15.0

cvinh:
@zaoral I'm interested with this PR, do you want me to review ?

PR #519, opened 2022-11-15
[ADD] account_statement_import_txt_xlsx: show mapping error

zaoral:
If there is a problem parsing the file we are not showing the error to the user, instead a generic message telling something is wrong with the type of file is shown which is not helpful (see image bellow), and we only show the error if we are running unit tests.

image

In this case, we think is good to show the error to the user, at least part of it, that way they can check what is wrong with the sheet mapping and fix it.

* We are only showing the title of the error as a HINT, we think that is a good start to understand what needs to change in the sheet mapping. Example   image

* The complete traceback error is shown in the log if we needed for more detailed reviews from the technical teams. Example  image

### DIFF
--- a/account_statement_import_txt_xlsx/models/account_statement_import.py
+++ b/account_statement_import_txt_xlsx/models/account_statement_import.py
@@ -3,7 +3,8 @@
 
 import logging
 
-from odoo import fields, models
+from odoo import _, fields, models
+from odoo.exceptions import UserError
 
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

cvinh (approved):
LGTM +1

AaronHForgeFlow (approved):
Code LG!

PR #522, opened 2022-11-17
[IMP] account_statment_import: Add button for importing on top

etobella:
This PR has two goals:

- [x] Adding a button on the top of the Journal Dashboard to import files
- [x] Block the import of statements from different journals if we use the one from the journal item

The main goal is to simplify the usage (and make it more consistent)

@pedrobaeza as talked on OCA/l10n-spain#2610

### DIFF
--- a/account_statement_import/wizard/account_statement_import.py
+++ b/account_statement_import/wizard/account_statement_import.py
@@ -220,6 +220,15 @@ class AccountStatementImport(models.TransientModel):
                 ],
                 limit=1,
             )
+            journal_id = self.env.context.get("journal_id")
+            if journal_id and journal.id != journal_id:
+                raise UserError(
+                    _(
+                        "The journal found for the file is not consistent with the "
+                        "selected journal. You should use the proper journal or "
+                        "use the generic button on the top of the Accounting Dashboard"
+                    )
+                )
 
             if not journal:
                 bank_accounts = self.env["res.partner.bank"].search(
--- a/account_statement_import/wizard/account_statement_import_view.xml
+++ b/account_statement_import/wizard/account_statement_import_view.xml
@@ -39,4 +39,12 @@
             <field name="target">new</field>
         </record>
 
+
+        <record id="account_statement_import_menu" model="ir.ui.menu">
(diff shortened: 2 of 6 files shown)


etobella:
image

account_statement_import/wizard/account_statement_import.py
```diff
@@ -220,6 +220,16 @@ def _match_journal(self, account_number, currency):
                 ],
                 limit=1,
             )
+            if self.env.context("journal_id") and journal.id != self.env.context.get(
+                "journal_id"
+            ):
```
pedrobaeza:
```suggestion
            journal = self.env.context.get("journal_id")
            if journal and journal.id != journal:
```

account_statement_import/wizard/account_statement_import.py
```diff
@@ -220,6 +220,16 @@ def _match_journal(self, account_number, currency):
...
+                raise UserError(
+                    _(
+                        "The journal found for the file is not consisent with the "
```
pedrobaeza:
```suggestion
                        "The journal found for the file is not consistent with the "
```

account_statement_import/wizard/account_statement_import.py
```diff
@@ -220,6 +220,16 @@ def _match_journal(self, account_number, currency):
...
+                        "selected journal. You should use the proper journal or "
+                        "use the generic button on the top of Invoicing Dashboard"
```
pedrobaeza:
```suggestion
                        "use the generic button on the top of the Accounting Dashboard"
```

account_statement_import/wizard/account_statement_import_view.xml
```diff
@@ -39,4 +39,12 @@
             <field name="target">new</field>
         </record>
 
+
```
pedrobaeza: Extra empty lines not needed

pedrobaeza (changes requested):

pedrobaeza (approved):

LoisRForgeFlow (approved):
LGTM

PR #526, opened 2022-11-20
[15.0]]MIG] account_statement_import_online_paypal

Fatimahairup:
In addition to migration that was done in https://github.com/Nitrokey/odoo-bank-statement-import/tree/15.0-mig-account_statement_import_online_paypal this PR contains fix to address two errors:

1. ValueError: Invalid field 'note' on model **account.bank.statement.line**: There isn't a field named note in **account.bank.statement.line**. However, there is a field named **narration** that has a string value of Note. Based on my understanding, it should be **narration** instead of **note**

2. Validation error that appears stating how a **mandatory** field is not filled up. This required field is **payment_ref**(string value Label) that should be assigned a value when creating line items in the **account.bank.statement.line** model. I have mapped value as I so fit(open to change)


dreispt:
@Fatimahairup Can you please squash your three last commits? Let me know in case you need help.

dreispt (approved):

PR #527, opened 2022-11-28
[FIX] account_statement_import

etobella:
This PR fixes two cases:

- It was leading to some JS test errors because the registry has not been correctly loaded. It might happen if you had this module installed and you tested `web_disable_export_group`
- The button was not showed because the create attribute must be set to 1. It will be modified later by the JS (probably an error from my local installation :cry: )

@pedrobaeza

### DIFF
--- a/account_statement_import/views/account_journal.xml
+++ b/account_statement_import/views/account_journal.xml
@@ -11,6 +11,13 @@
         <field name="model">account.journal</field>
         <field name="inherit_id" ref="account.account_journal_dashboard_kanban_view" />
         <field name="arch" type="xml">
+            <!--
+                We need to add the create tag in order to show buttons.
+                However, we will change them using the JS definition.
+            -->
+            <kanban position="attributes">
+                <attribute name="create">1</attribute>
+            </kanban>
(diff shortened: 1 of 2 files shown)


account_statement_import/views/account_journal.xml
```diff
@@ -11,6 +11,9 @@
         <field name="model">account.journal</field>
         <field name="inherit_id" ref="account.account_journal_dashboard_kanban_view" />
         <field name="arch" type="xml">
+            <kanban position="attributes">
+                <attribute name="create">1</attribute>
```
pedrobaeza, line 19: This one is weird, as creating a journal from the dashboard is something outside the module scope.

etobella: The create is used to show the buttons on top. Otherwise it is not rendered and we cannot show the import button

pedrobaeza (approved):
OK, please put the comment in XML clarifying that this makes the button to appear

pedrobaeza:
Let's fix the problem:

/ocabot merge minor

PR #530, opened 2022-12-05
[13.0][FIX] account_bank_statement_import_online_qonto: Avoid errors if year is different

victoralmau:
Avoid errors if year is different.

If we define as monthly frequency, the dates that were defined before were: 2022-12-01 and 2023-01-01. This seems correct, but the API returns error:

`The datetime in settled_at_to param isn't well formatted (ISO 8601) or year is out of range`

The error is actually related to the fact that the year of both dates is different.

Please @pedrobaeza can you review it?

@Tecnativa TT40742

### DIFF
--- a/account_bank_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
+++ b/account_bank_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
@@ -6,7 +6,7 @@ from datetime import datetime
 import pytz
 import requests
 
-from odoo import _, api, models
+from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
 from odoo.addons.base.models.res_bank import sanitize_account_number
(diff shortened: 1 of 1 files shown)


account_bank_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
```diff
@@ -62,6 +62,11 @@ def _qonto_obtain_transactions(self, slug, date_since, date_until):
         self.ensure_one()
         url = QONTO_ENDPOINT + "/transactions"
         params = {"slug": slug, "iban": self.account_number}
+        # Avoid errors if year is different (set last day of the year)
+        if date_since and date_until and date_since.year != date_until.year:
+            date_until = fields.Datetime.from_string(
+                "%s-12-31 23:59:59" % date_since.year
+            )
```
pedrobaeza: Remove this empty line now that you are here.

account_bank_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
```diff
@@ -62,6 +62,11 @@ def _qonto_obtain_transactions(self, slug, date_since, date_until):
         self.ensure_one()
         url = QONTO_ENDPOINT + "/transactions"
         params = {"slug": slug, "iban": self.account_number}
+        # Avoid errors if year is different (set last day of the year)
```
pedrobaeza: Include the error you have if not:
```suggestion
        # Avoid this error if the end year is different: "The datetime in 
        # settled_at_to param isn't well formatted (ISO 8601) or year is out
        # of range". We set the last day of the year in such case.
```

victoralmau:
Changes done.

pedrobaeza (approved):

victoralmau:
Ping @CarlosRoca13

CarlosRoca13 (approved):

PR #533, opened 2022-12-10
[15.0]account_statement_import - set correct company_id on ir_attachment

luc-demeyer:
Small fix to set correct company_id on bank statement attachments.

At this point in time, setting the incorrect company_id (the first from the 'allowed_company_ids' is set) has no operational effect on a standard Odoo setup since there are no record rules on ir.attachments, hence a wrong company_id has no operational impact.


MiquelRForgeFlow (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #536, opened 2022-12-16
[IMP] account_bank_statement_import_txt_xlsx: Allow mapping with reference to column numbers, and concatenation

JordiBForgeFlow:
A new field 'File does not contain header line' is added in the Statement Sheet Mappings. If you set to True, then you can map the columns by indicating in each field of the 'Columns' section the column number in the file.

We also allow to concatenate multiple columns in the file to a single column to a single field of the statement line. You have to indicate the names of the columns separated by comma.

image

### DIFF
--- a/account_bank_statement_import_txt_xlsx/__manifest__.py
+++ b/account_bank_statement_import_txt_xlsx/__manifest__.py
@@ -17,7 +17,7 @@
         "multi_step_wizard",
         "web_widget_dropdown_dynamic",
     ],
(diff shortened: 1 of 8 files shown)


AaronHForgeFlow (approved):
Code review. LG both improvements, the reference column numbers, and the encoding detection part.

MiquelRForgeFlow:
Forward ported in https://github.com/OCA/bank-statement-import/pull/568.

PR #513, opened 2022-10-22
[14.0][IMP] account_statement_import_txt_xlsx - improve columns/fields inheratibility 

ghost:
If you want to add new fields in the mapping, it's a bit hard since everything is done inside loops. So I split the parsing in several functions to be able to do a super in the middle of the process and be able to add new fields/columns

Example (partial one) with the improvements : 

```
def _parse_columns(self, mapping, header, columns):
        res = super()._parse_columns(mapping, header, columns)
        columns["transaction_date_column"] = (
            header.index(mapping.transaction_date_column) if mapping.transaction_date_column else None
        )
        columns["value_date_column"] = (
            header.index(mapping.value_date_column) if mapping.value_date_column else None
        )
        return res
```

PR #538, opened 2022-12-26
[14.0][REF] account_statement_import_txt_xlsx: refactor for easy to hook

ps-tubtim:
This pull request is used to refactor for an easy hook.


ps-tubtim:
Move to https://github.com/OCA/bank-statement-import/pull/513.

PR #541, opened 2023-01-06
[12.0] [FIX] Remove synchronization from Ponto

NL66278:


NL66278:
The pre-commit failed with a message that it could not find python 3.6. Does anybody have any idea why that can be? It is in the list of supported versions for github actions.

pedrobaeza:
Pre-commit fixed in #544

NL66278:
Closed in favor of this one: https://github.com/OCA/bank-statement-import/pull/552

PR #542, opened 2023-01-06
[14.0] [FIX] ponto import

NL66278:
Alternative attempt to fix Ponto import (problem of missing data), without buffering.


pedrobaeza:
Yes, please let's go with this one instead of the buffer one. cc @alexis-via @alexey-pelykh

pedrobaeza:
Can you simplify the commit history a bit for not adding and then removing things?

NL66278:
@pedrobaeza There are now just two commits left.

pedrobaeza (approved):
@alexis-via is it OK for you?

njeudy (approved):
Just tested on a copy of production database, and check code . Ok for me.

pedrobaeza:
Let's merge it then:

/ocabot merge nobump

Can you do the fw-port?

pedrobaeza:
Seeing that the PR already contains the version bump:

/ocabot merge nobump

@NL66278 it's not needed to do it, as the bot does it and there's no migration scripts.

PR #543, opened 2023-01-06
[14.0][IMP] *_import_online: always update next_run to cover current period

NL66278:


alexey-pelykh (approved):
Looks good, optimization! Just adjust the commit message to be more specific

NL66278:
@alexey-pelykh I changed the commit message. Thanks for your review!

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -69,7 +69,10 @@ class OnlineBankStatementProvider(models.Model):
         string="Update Schedule",
         compute="_compute_update_schedule",
     )
-    last_successful_run = fields.Datetime(string="Last successful pull")
+    last_successful_run = fields.Datetime(
+        string="Last successful pull",
+        readonly=True,
```
pedrobaeza, line 74: Why doing this readonly? What if I want to change it manually?

pedrobaeza (commented):
I have problems understanding what this is fixing.

account_statement_import_online/models/online_bank_statement_provider.py
NL66278: @pedrobaeza Why would you change this manually? Is is literally the last time the job last run.

pedrobaeza: OK, if it's not used for any sync related thing.

NL66278: @pedrobaeza As for what it is fixing: suppose the server running Odoo is not always active (for upgrade or other maintenance etc.). It means that the next_run which is supposed to be updated at regular intervals (example every 4 hours), is not always updated and will get further and further back in time. As the `_scheduled_pull` method only retrieves data until the next_run date, you will be getting data back with ever greater delays.  This makes sure that at the end of each run, when the next_run is increased by the run interval, the next_run will be in the future.

pedrobaeza: And why not simply change the fetch to get from last_run to now?

NL66278: Simply running until now would also solve the problem I outlined. Adjusting the next_run has the same effect with the added advantage of correctly showing when the next run will be.

pedrobaeza: Well, I always prefer the simplest formula, but as you prefer.

PR #535, opened 2022-12-15
[16.0][MIG] account_statement_import: Migrate to version 16.0

sonhd91:
- `view_bank_statement_line_form` was [removed](https://github.com/odoo/odoo/commit/4fb0933986534e31337aa1dce3478a67deea72de#diff-f3049e00919d62bbf6b11a37517e834396157646af188640bf7763122fe92957) from odoo 16
- Change module name and it's content from `account_statement_import` to `account_statement_import_file`
- Port-forward
* https://github.com/OCA/bank-statement-import/pull/482
* https://github.com/OCA/bank-statement-import/pull/522
* https://github.com/OCA/bank-statement-import/pull/527

PR #548, opened 2023-01-17
[16.0][MIG] account_statement_import_online: Migration to 16.0

NL66278:
This builds on the migration of the base modules here: https://github.com/OCA/bank-statement-import/pull/535

Update: PR 575 was closed in favour of this that has been merged: https://github.com/OCA/bank-statement-import/pull/575


requirements.txt
```diff
@@ -0,0 +1,2 @@
+# generated from manifests external_dependencies
+odoo_test_helper
```
MiquelRForgeFlow: Could you please cherry-pick https://github.com/OCA/bank-statement-import/commit/9df67995671b2f47b91d591e47ebf38af3115f1c? :)

NL66278: Done!

NL66278:
@MiquelRForgeFlow Everything green now after updating test and rebasing. Could you have a fresh look?

MiquelRForgeFlow (approved):

requirements.txt
```diff
@@ -1,2 +1,3 @@
 # generated from manifests external_dependencies
 ofxparse
+odoo-test-helper
```
MiquelRForgeFlow: see https://github.com/OCA/bank-statement-import/blob/15.0/test-requirements.txt, it should go in test-requirements.txt

NL66278: @MiquelRForgeFlow I changed this now. Thanks!

MiquelRForgeFlow (approved):

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -177,12 +177,14 @@ def _log_provider_exception(
         """Both log error, and post a message on the provider record."""
         self.ensure_one()
         _logger.warning(
-            'Online Bank Statement provider "%s" failed to'
-            " obtain statement data since %s until %s"
-            % (
-                self.name,
-                statement_date_since,
-                statement_date_until,
+            _(
```
StefanRijnhart, line 221: Can't seem to find anything in the guidelines, but isn't it the best practice *not* to translate logging messages (as opposed to Exception messages, which are shown to the user)?

NL66278: The guidelines say messages should be translated. I like to see these messages in English too, as it helps in googling if there are any problems. But that can be done by running these procedures under a profile that has English as a language. On the other hand there maybe admins in many countries that prefer their own language or do not command the English language at all.

StefanRijnhart: OK, thanks for clarifying!

StefanRijnhart (approved):

account_statement_import_online/models/online_bank_statement_provider.py
pedrobaeza: This is not only a question of preferences, but logging tools are not very well handled with other non-English characters, so this message shouldn't be translated. Only UI messages should. If you think the guidelines are not very clear in this, we can improve it through a PR on them.

NL66278: @pedrobaeza What logging tools are not able to handle non English characters? Definitely the standard logging fully supports internationalization.

pedrobaeza: Well, you would be surprised, specially pipelining, or remote logging. But anyway, it's not the point, logging messages always in English, please.

inafarrate-busman (approved):

len-foss (approved):
~~@pedrobaeza this one is ready to be merged when you pass by :-)~~
(clearly went too fast with this one)

len-foss (changes requested):
Actually, why is 32d155eedfb5b8c5492a852152981407c4ff0149 removing a lot of features without any explanation?
I would expect at least a message explaining what should be used instead of the removed fields. 
It would also be good to put a warning somewhere about it (like in the Migration issue, or even the repo readme as long as every depending module is not migrated).

Found this while testing the migration of _online_paypal.

NL66278:
I put this in WIP state to address later the remarks made.

mariadforgeflow:
@NL66278 hi! Will you continue with this one? Or can our team open a new one and continue with it?

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
```diff
@@ -0,0 +1,33 @@
...
+        required=True,
+        default=fields.Datetime.now,
+    )
+
+    def action_pull(self):
+        """Pull statements from provider and then show list of statements."""
+        self.ensure_one()
+        provider_model = self.env["online.bank.statement.provider"]
+        active_id = self.env.context.get("active_id")
+        provider = provider_model.browse(active_id)
```
artem1205: Button does not work

```suggestion
        active_id = self.env.context.get("active_id")
        journal_id = self.env[self.env.context.get('active_model')].browse(active_id)
        provider = journal_id.online_bank_statement_provider_id
```

account_statement_import_online/models/account_journal.py
```diff
@@ -0,0 +1,101 @@
...
+            ) or OnlineBankStatementProvider.create(
+                {
+                    "journal_id": journal.id,
+                    "service": service,
+                }
+            )
+            journal.online_bank_statement_provider_id = provider
+            _logger.info("Journal %s now linked to service %s", journal.name, service)
+
+    @api.model
```
artem1205: Please add batch support for this method to be align with parent method `@api.model_create_multi`

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
astirpe: @artem1205 your solution fixes the button on the journal but breaks the button on the payment provider.

NL66278: @artem1205 Thanks for your comment and suggestion. With new commit wizard should work both from journal and from provider record.

account_statement_import_online/models/account_journal.py
NL66278: @artem1205 Done! This also led me to do some more streamlining / refactoring.

NL66278:
@mariadforgeflow would you have another look at this as well?

victoralmau (changes requested):
Can you unify all your changes into 1 or 2 commits? Now there are too many extra commits and I think almost everything could go in 1 explaining in description commit.

The changes in the setup folder would be part of pre-commit (no such commit before the migration commit),

NL66278:
@victoralmau I consolidated all my migration commits into 1, only kept separate the commit contributed by Miquel Raïch

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
```diff
@@ -0,0 +1,37 @@
...
+        self.ensure_one()
+        active_model = self.env.context.get("active_model")
+        active_id = self.env.context.get("active_id")
+        active_record = self.env[active_model].browse(active_id)
+        if active_model == "account.journal":
+            provider = active_record.online_bank_statement_provider_id
+        else:
+            provider = active_record
+        provider._pull(self.date_since, self.date_until)
+        action = self.env.ref("account.action_bank_statement_tree").sudo().read([])[0]
```
victoralmau, line 35:
```suggestion
        action = self.env["ir.actions.act_window"]._for_xml_id(
            "account.action_bank_statement_tree"
        )
```

victoralmau (approved):
Just a small comment, if `odoo_test_helper` is not needed, can you remove it from `test-requirements.txt`?

MiquelRForgeFlow:
> Just a small comment, if odoo_test_helper is not needed, can you remove it from test-requirements.txt?

Is needed, but just in the tests. That's why is inside test-requirements.txt.

patel1008b:
@NL66278 @StefanRijnhart Any idea when it will be merge into 16.0 ?

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -93,7 +92,6 @@ class OnlineBankStatementProvider(models.Model):
     certificate_public_key = fields.Text()
     certificate_private_key = fields.Text()
     certificate_chain = fields.Text()
-    allow_empty_statements = fields.Boolean()
```
stefan-tecnativa: I was reading the comments and story of this PR, and I can't figure out why this field is being deleted. 
@victoralmau  do you remember this PR? Could you explain to me why this change is made?

NL66278: @stefan-tecnativa Empty statements where a bit strange always, but if you wanted to have a statement for each day, they could be used to check for completeness: have we retrieved all the statements or are there gaps. In 16.0 statement date has become a computed field, dependent on the statement lines. So this does not work anymore. In a lot of ways in 16.0 a statement is a kind of summary of the lines it contains. So empty statements do not make sense and should no longer be created or even allowed.

stefan-tecnativa: @NL66278 thank you very much for the explanation!

PR #552, opened 2023-01-22
[12.0][IMP] *_online_ponto improvements

NL66278:
This is a more comprehensive change then just removing the synchronization, similar to the same changes proposed for 14.0.


NL66278:
Now based on https://github.com/OCA/bank-statement-import/pull/590

Apart from better structure, back-ported from 14.0, if also removes the buffer models that where never accepted for merge in 14.0, replaced by buffering in memory.

PR #125, opened 2017-09-12
[MIG][10.0] account_bank_statement_import_mt940_base: migration to 10.0

astirpe:
This is the porting of module `account_bank_statement_import_mt940_base` to V10.

Based on the job made in https://github.com/OCA/bank-statement-import/pull/98 by @pedrocasi

PR #553, opened 2023-01-25
[10.0][FIX] bank_statement_import_mt940_base: Allow importing files with special characters

rousseldenis:


account_bank_statement_import_mt940_base/test_files/test-sns-utf8.940
```diff
@@ -0,0 +1,78 @@
+{1:F01SNSBNL2AXXXX0000000000}{2:O940SNSBNL2AXXXXN}{3:}{4:
```
NL66278, line 1: Is this real customer data? In that case maybe better anonymize it...

rousseldenis: @NL66278 I've done nothing more than a copy paste from existing ones in tests. I've added a special character 'à' in it. That's all

rousseldenis: Maybe we can ask @astirpe who did the first work

astirpe: I got those files from here: https://github.com/OCA/bank-statement-import/pull/125#discussion_r159695587 
Not sure if they are already anonymized.

rousseldenis: @danger89

melroy89: My bank account was already anonymized. It could be that the other side (which is always a business) is not always anonymized.

PR #575, opened 2023-03-06
[16.0][MIG] account_statement_import_file

alexis-via:
As already discussed, the module was renamed from account_statement_import to account_statement_import_file.

The migration has been done from v14 to v16, to take into account the big changes that took place in v14 with the introduction of the module account_statement_import_base. The changes from v15 have been forward-ported to v16.

I re-designed the notification system to avoid having specific code in account_reconcile_oca: I just use the notification pop-up system.

PR #577, opened 2023-03-06
[16.0][MIG] account_statement_import_ofx

alexis-via:


alexis-via:
Requires https://github.com/OCA/bank-statement-import/pull/575

PR #580, opened 2023-03-29
[13.0][IMP] account_bank_statement_import_camt_oca: extend the use of imported CAMT054.

JordiBForgeFlow:
The CAMT054 file is often used in combination with the CAMT053. While the CAMT053 contains the statement of an account, the CAMT054 provides additional details on the partner and the transaction.

CAMT053 and CAMT054 are two XML-based message types used in financial messaging
between banks and their customers.

CAMT053 is used for reporting of account statements, while CAMT054 is used for reporting of
individual transactions.

In combination, CAMT053 and CAMT054 can be used to provide a complete picture of a bank account's activity.
CAMT053 can be used to provide a statement of account balances and transactions for a specific period,
while CAMT054 can be used to provide detailed information about each individual transaction within that period.

For example, a bank may provide a CAMT053 message to their customer at the end of the month,
which includes the balance and summary of all transactions for the month. If the customer wants more detailed
information on a specific transaction, they can request a CAMT054 message for that transaction, which provides
details such as the date, amount, currency, and reference information.

Using CAMT053 and CAMT054 together can provide a more comprehensive view of a bank account's activity,
allowing customers to track their finances more effectively and make informed decisions.

Usage
=====

#. Go to the Accounting Dashboard, to the specific bank journal and press Import.
#. Select the SEPA CAMT053 file and import it. Press "UPLOAD".
#. A new bank statement will be created, that will include all the transactions that have been imported.

If you have a SEPA CAMT054 file that provides additional information to the CAMT053, you can follow the same
steps as above, and the application will update the previously imported transactions, updating the fields 'Label'
and the partner name.

Sometimes companies will only import the CAMT054 as a standalone file (not related with a CAMT053 file).
In this case a new statement will be created.

### DIFF
--- a/account_bank_statement_import_camt_oca/models/account_bank_statement_import.py
+++ b/account_bank_statement_import_camt_oca/models/account_bank_statement_import.py
@@ -4,7 +4,7 @@ import logging
 import zipfile
 from io import BytesIO
 
-from odoo import models
+from odoo import _, models
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


JordiBForgeFlow:
@luc-demeyer @gaikaz @astirpe @i-vyshnevska what do you think about this PR?

JordiBForgeFlow:
Closing the PR for now. As I understand it now, CAMT054 files sould be imported into a separate journal.

PR #573, opened 2023-02-25
[15.0][FIX] account_statement_import_txt_xlsx: Don't try to parse if no mapping

pedrobaeza:
@Tecnativa

PR #582, opened 2023-04-13
[14.0][FIX] account_statement_import_txt_xlsx: only consider parsing when a mapping sheet is specified (forward port)

MiquelRForgeFlow:
Forward port of:
- https://github.com/OCA/bank-statement-import/pull/581


pedrobaeza (approved):
I did it for v15 in #573 

/ocabot merge patch

Please check the Paypal too that has the same problem

PR #583, opened 2023-04-13
[13.0][FIX] account_bank_statement_import_paypal: parse only if mapping if specified

MiquelRForgeFlow:
Similar to:

- https://github.com/OCA/bank-statement-import/pull/581


pedrobaeza (approved):
/ocabot merge patch

PR #584, opened 2023-04-13
[14.0][FIX] account_statement_import_paypal: parse only if mapping if specified (forward port)

MiquelRForgeFlow:
Forward port of:

- https://github.com/OCA/bank-statement-import/pull/583


pedrobaeza (approved):
/ocabot merge patch

PR #588, opened 2023-05-01
[FIX] account_statement_import_online_paypal: empty statements create…

kwoychesko:
…d regardless of configuration setting

Empty statements are created upon each download from PayPal regardless of the setting of the "Allow Empty Statements" field (allow_empty_statements) because this field is checked within an _if statement_ that can never be true. This fix checks allow_empty_statements and aborts the statement creation as appropriate.


kwoychesko:
Forgot to update Tests — will create tests and recreate a PR.

PR #496, opened 2022-09-15
[15.0][add] account_bank_statement_import_online_nordigen

JordiBForgeFlow:
Integration with Nordigen

TODO:

* [ ] Test scripts

PR #592, opened 2023-05-11
[15.0] [ADD] account_bank_statement_import_online_nordigen

Entrepreneur-AJ:
Commit the hard work already created by Jordi with PR https://github.com/OCA/bank-statement-import/pull/496

Remove the sandbox institution id that prevented real open banking operation.


ayushin:
Awesome! Let get this one in, I've done testing on my side with no findings.

MiquelRForgeFlow:
Hello @Entrepreneur-AJ , thanks for your work, but it would be nicer if you could respect commit history 🤔

Rad0van:
> Hello @Entrepreneur-AJ , thanks for your work, but it would be nicer if you could respect commit history thinking

That's what I was going to write. This way it seems you (@Entrepreneur-AJ) have created this while the only thing you did is remove one line and maybe some other cosmetic changes?  Hard to tell as commit history is lost here.

This is definitely not the way to go...

Entrepreneur-AJ:
@MiquelRForgeFlow Sorry, wasn't 100% on best way to take the work that had already be done and add on thats why I referenced PR #496 as first priority on this thread

Entrepreneur-AJ:
@MiquelRForgeFlow @Rad0van Need me to start again figuring out retaining commit history?

Rad0van:
> @MiquelRForgeFlow @Rad0van Need me to start again figuring out retaining commit history?

This way there is no chance of being accepted. But I do not think that by creating a copy with one line removed it is going to pass (even if Jordi suggested something like that). Jordi can remove that line perfectly by himself. If you tested the thing properly you would see there are still some issues (mostly coming from the way API is working) that need to be addressed. At least we have found some and will propose changes into original PR.

If so, try to focus on adding some added value to it. If you need to use it now, nothing is blocking you from doing so. Download it, deploy it and use.

JordiBForgeFlow:
Closing in favor of https://github.com/OCA/bank-statement-import/pull/496

PR #593, opened 2023-05-15
[16.0][imp] - sanitized bank account number instead of account number

badbole:
Use sanitized_acc_number instead of acc_number - so the wizard from module import_bank_statement_file gets correct values for comparison, as it is also using sinitized_acc_number


badbole:
@alexis-via could you check/validate  this one please

badbole:
@alexis-via did you have the time to check this? not sure why runboat/build is pending or how to start it...

bosd:
> why runboat/build is pending or how to start it...

You could rebase and (force) push to create a runboat.

PR #595, opened 2023-05-16
[13.0][ADD] account_bank_statement_import_file_hook

JaumeBforgeFlow:


GuillemCForgeFlow (approved):
LGTM! 👍🏽

mariadforgeflow (approved):
LVGTM!

MiquelRForgeFlow (approved):

JordiMForgeFlow (approved):

AaronHForgeFlow (approved):

pedrobaeza (changes requested):
Wait, aborting. Why do you need to do such hooks? This is a very bad practice. What do you want to get?

JaumeBforgeFlow:
@pedrobaeza, Odoo does the processing of the file in such a way that you are not able to tap into the logic, to modify it while having access to the ir.attachment record being processed. None of the functions called inside the for loop in the import_file() function are passed the attachment as a record.

Having this hook might be interesting in cases such as the one we are working with, where we want to link the resulting statement with the attachment.

pedrobaeza:
I think you can override `_parse_file` method, that has as argument the decoded file content, and include in the returned dictionary these contents in a special key. Then override the `create` method of `account.bank.statement`, and if you find such key, pop it, call super, and then create the attachment.

AaronHForgeFlow:
Hi, just reopening this

AaronHForgeFlow:
sorry wrong PR

PR #602, opened 2023-06-08
[16.0] account_statement_import_file: fix string

alexis-via:


alexis-via:
The string "Get you bank statements" is wrong ("r" is missing), so it needs to be fixed. I decided to re-write the full string.

PR #604, opened 2023-06-12
[16.0] [MIG] account_statement_import

marielejeune:
In version 16 `journal_id` field on `account.bank.statement` became a computed (stored) field. It is set by default as the journal_id of the account bank statement lines. Hence when creating the bank statements when importing bank files, we must add the journal_id not only on the bank statement, but also on the bank statement lines.


marielejeune:
The module was already migrated and renamed `account_statement_import_file`

PR #605, opened 2023-06-13
[16.0][mig] account_bank_statement_clear_partner: Migration to 16.0

dsolanki-initos:


alihashaam:
@dsolanki-initos review is done

PR #610, opened 2023-07-13
[16.0][FIX] account_statement_import_file: Rename file_import selection value to file_import_oca

ao-landoo:
Both this module and enterprise's `account_bank_statement_import` module are adding the same key (`file_import`) to the list. This was not thowing errors until V15, but in V16 OWL is throwing a traceback because duplicated keys are not allowed.

This PR makes sure the key does not exist before adding it.

@ae-landoo @alexis-via @pedrobaeza

### DIFF
--- a/account_statement_import_file/migrations/16.0.1.0.1/post-migration.py
+++ b/account_statement_import_file/migrations/16.0.1.0.1/post-migration.py
@@ -1,0 +1,16 @@
+# Copyright 2023 Landoo Sistemas de Informacion SL
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_journal
+        SET bank_statements_source = 'file_import_oca'
+        WHERE bank_statements_source = 'file_import'
+        """,
+    )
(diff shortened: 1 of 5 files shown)


pedrobaeza:
But why is duplicated?

ao-landoo:
Because enterprise also adds it in the module I mentioned above

This screenshot is taken on V15

Captura de pantalla de 2023-07-13 09-24-39

pedrobaeza:
Then this one should have other selection value IMO.

ao-landoo:
Does it look good this way for you @alexis-via @pedrobaeza ?

account_statement_import_file/migrations/16.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,15 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not version:
+        return
+    journals = env["account.journal"].search(
+        [("bank_statements_source", "=", "file_import")]
+    )
+    for journal in journals:
+        journal.bank_statements_source = "file_import_oca"
```
pedrobaeza: - No need of the version check (it's done by the decorator)
- Do things always on SQL (when possible), for performance and for avoiding triggering business logic that may have side effects.

```suggestion
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE account_journal
        SET bank_statements_source = 'file_import_oca'
        WHERE bank_statements_source = 'file_import'
        """
    )
```

ao-landoo: Changed

rp-landoo (approved):
Functional approval

ao-landoo:
@pedrobaeza If you see it OK can you merge it?

pedrobaeza (approved):
/ocabot merge nobump

PR #612, opened 2023-07-17
[FIX] account_statement_import_file: Compatibility module account_bank_statement_import from enterprise.

pablohmontenegro:
Ticket: 63584


pedrobaeza:
Isn't this solved already in https://github.com/OCA/bank-statement-import/pull/610?

Issue #603, opened 2023-06-09
[V15] Error when trying to import CSV (or XLSX) bank statement


FreezeS:
Hi guys,

Running Odoo 15.0-20230517 (Community Edition).
I'm trying to import a bank statement and keep getting a very unhelpful error.

Module installed:
![image](https://github.com/OCA/bank-statement-import/assets/2807003/a60869c4-4270-4fc0-ad82-c5c0ff6c6fc6)

The first line of the file is this:
    01.05.2023; Detalii: INREGISTRARE DOBANDA OVERDRAFT;-550,71;-113324,64

Created manual mapping:
![image](https://github.com/OCA/bank-statement-import/assets/2807003/d96cc718-9bff-45b7-a1a8-1a1d825f2401)

Created automatic mapping (Action - Import Mapping):
![image](https://github.com/OCA/bank-statement-import/assets/2807003/ffa20082-6477-476d-ac63-f13867c45293)
![image](https://github.com/OCA/bank-statement-import/assets/2807003/d21ff5cb-ae40-4acc-a142-485c7b06ccde)

I tried to import the CSV using both mappings:
![image](https://github.com/OCA/bank-statement-import/assets/2807003/c5bb57c6-34d5-4e96-9bdf-3e67c3202d35)
![image](https://github.com/OCA/bank-statement-import/assets/2807003/636fe201-1aa7-45d4-9be1-6fcc24c3d471)

And got the same error:
This bank statement file format is not supported.
Did you install the Odoo module to support this format?

![image](https://github.com/OCA/bank-statement-import/assets/2807003/a0d7c89c-c97b-4596-b504-19ba2e655790)

I tried for CSV and XLSX
When I run Import Mapping it sees all the fields and the mapping is generated properly but the import itself fails.
A more detailed error would help so at least I know where to look.

denilson020898:
You may want to check your csv or excel file.
For my case, decimal number must not be null and the date format must match the setting template.

Try make your csv as the following, from the code sample

"Date","Label","Debit","Credit","Balance","Partner Name","Bank Account"
"12/15/2018","Credit 20.00","0.00","20.00","-10.00","John Doe","123456789"
"12/15/2018","Credit 13.50","0.00","-13.50","-23.50","John Doe","123456789"
"12/15/2018","Debit 33.50","-33.50","0.00","10.00","Azure Interior",""
"12/15/2018","Debit 1500","1,500.00","0.00","1,510.00","Azure Interior",""

[...]

PR #613, opened 2023-07-20
14.0_txt_xlsx_better_error_handling

saravanabalagi:
Currently any problem during parsing is reported as "File format not supported, did you install the Odoo module". Related issue: #603 

- Handle empty str* in amount as 0, prevents Decimal Conversion error
- Show more common file content problems: Key and Value error
- For all other errors, report as parsing error

*some banks provide CSVs with only debit or credit filled while leaving the other empty

### DIFF
--- a/account_statement_import_txt_xlsx/models/account_statement_import.py
+++ b/account_statement_import_txt_xlsx/models/account_statement_import.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from odoo import fields, models
+from odoo import _, fields, models
 from odoo.exceptions import UserError
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


saravanabalagi:
Catching decimal exception in the inner function is necessary, as catching it in parent similar to ValueError will not provide any details of the original value that was attempted to be parsed. Args for decimal exceptions have only the signal flags  [[1](https://docs.python.org/3.11/library/decimal.html#quick-start-tutorial), [2](https://stackoverflow.com/a/23322692/3211422), [3](https://stackoverflow.com/a/72697436/3211422)].

cvinh:
Hello @saravanabalagi did you check this PR : https://github.com/OCA/bank-statement-import/pull/519
It looks like it's simpler

PR #621, opened 2023-08-31
[15.0][ADD] account_statement_import_txt_xlsx: Strip the currency symbol from numeric fields

azubieta:
Some bank statements include the currency symbol in the numeric fields, this adds a field to list which currencies are being used in the statement and remove their symbol when parsing decimal lines.


bosd:
Did this get merged in any way?

azubieta:
@bosd It seems that the maintainer have been a bit busy.

alexey-pelykh:
I think it's better to just strip all non-digit-comma-dot characters then prior to parsing rather than selectively replace currency symbols.

bosd:
> I think it's better to just strip all non-digit-comma-dot characters then prior to parsing rather than selectively replace currency symbols.

I agree on this one :+1:

azubieta:
@alexey-pelykh sounds good will do the change.

PR #624, opened 2023-09-17
Fix journal matching for CAMT files from e.g. Commerzbank

luc-demeyer:


BT-cjimeno (approved):
LGTM, however I think it's needed to also check if result["account_number"] and result["currency"] are there before the last character checks.

[Or this is what I see them doing at the bottom of this file](https://github.com/OCA/bank-statement-import/pull/624/files#diff-19266ce0b77e7996d1ea147cee54328a046626866715eaa8964dfc0fef3af934L453-L456)

Tested successfully with files from Commerzbank that contain the currency at the end of the bank_account, e.g. "EUR" while the matching journals in Odoo don't include the currency code at the end.

luc-demeyer:
Can this PR be reopened ?

PR #625, opened 2023-09-19
Fix translation

JonathanStein:
Button text was not translated with the _t function, but Odoo seems to handle the translation by native methods


pedrobaeza:
Please use a proper commit message: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

and check why it was done that way through git blame.

PR #626, opened 2023-09-25
[16.0][MIG] account_statement_import_online_adyen

NL66278:
This PR includes some changes to other modules to prevent interference with the workings of this module.

Note that the migrated import of statements has been tested and has extensive user tests, but that the bare minimum has been done for the technical migration of the account_statement_clearing_account module, as I have no extensive knowledge of reconciliation in 16.0.


ntsirintanis (approved):

lfreeke (commented):
:+1: (functional review)

Works well for our customer

NL66278:
@StefanRijnhart Could you have a look at this one and reopen it? Maybe even review and merge it?

lfreeke (approved):
👍 Approved

thomaspaulb:
@NL66278 In a case like this the correct ping is to the maintainer team of the repo which can be found [here](https://oca.github.io/repo-maintainer-conf/bank.html)

For example:

@OCA/banking-maintainers this is approved and ready for merge

account_statement_import_adyen/wizards/account_statement_import_adyen_parser.py
```diff
@@ -0,0 +1,283 @@
...
+        seen_company_account = False
+        for num, header in enumerate(row):
+            if not header.strip():
+                continue  # Ignore empty columns.
+            if header == "Company Account":
+                seen_company_account = True
+            if header not in COLUMNS:
+                _logger.debug(_("Unknown header %s in Adyen statement headers"), header)
+            else:
+                COLUMNS[header] = num  # Set the right number for the column.
```
alexey-pelykh: `_set_columns()` mutates the module-level `COLUMNS` dict. This state persists for the lifetime of the Odoo worker process, so importing a file with non-standard column order permanently affects all subsequent imports on that worker.

Safest fix: copy the dict at the start of `parse_rows()` and use the instance copy throughout:
```python
def parse_rows(self, rows):
    self._columns = dict(COLUMNS)  # fresh copy per parse
    ...

def _set_columns(self, row):
    ...
    self._columns[header] = num

def _get_value(self, row, column):
    return row[self._columns[column]]
```

account_statement_import_adyen/wizards/account_statement_import.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def _parse_file(self, data_file):
+        """Parse an Adyen xlsx file and map merchant account strings to journals."""
+        try:
+            return self._parse_adyen_file(data_file)
+        except Exception as exc:  # pylint: disable=broad-except
+            message = (
+                _("Statement file %s was not a Adyen settlement details file.")
+                % self.statement_filename
+                or "* no filename *"
```
alexey-pelykh: Operator precedence: `%` binds tighter than `or`, so this evaluates as `(formatted_string) or fallback`. The formatted string is always truthy, making `or "* no filename *"` dead code.

If you need the filename fallback:
```python
filename = self.statement_filename or "* no filename *"
message = _("Statement file %s was not a Adyen settlement details file.") % filename
```

alexey-pelykh (commented):
Thanks for the migration work, Ronald -- getting three modules ported together with their interdependencies is no small task, and the CAMT recursive call fix is a nice catch.

I reviewed the code and found a couple of things worth flagging.

**Mutable module-level `COLUMNS` dict (concurrency risk)**

In the Adyen parser, `_set_columns()` mutates the module-level `COLUMNS` dictionary. Since module-level state persists across requests in a long-running Odoo worker, importing a file whose columns are in a different order would permanently alter the column mapping for all subsequent imports until the worker restarts. This is a real risk with Adyen files since different report versions can have different column layouts.

Safest fix would be to copy the dict at the start of each `parse_rows()` call, store it on `self`, and use that instance copy throughout.

**Operator precedence in error message**

In the import wizard, `%` binds tighter than `or`, making the `or "* no filename *"` fallback dead code (the formatted string is always truthy). See inline comment for the fix.

**Clearing account module -- incomplete migration acknowledged**

As noted in the PR description, the clearing account migration is minimal. The test has TODO comments about the statement model changes and doesn't fully exercise the reconciliation logic. Since CI passes, the basic structure works. The functional review from @lfreeke is reassuring here -- just want to confirm the actual confirm/reconcile flow was tested in a 16.0 instance.

**README badges reference 13.0**

The generated `README.rst` files in all three new modules have badge URLs and links pointing to `13.0` branches (GitHub, Weblate, Runbot). These should be regenerated with `oca-gen-addon-readme` for 16.0.

**Minor: `attrs` dict syntax in view**

In `online_bank_statement_provider.xml`, the `attrs="{'invisible':...}"` dict syntax works in 16.0 but is deprecated. The newer inline syntax would be `invisible="service != 'adyen'"`.

The CAMT fix (`self._parse_file()` -> `parser.parse()`) is correct -- the recursive call was clearly a bug since it would have re-entered the full parser chain instead of just the CAMT parser.

Overall: well-structured migration with good test coverage for the Adyen import path. The module-level COLUMNS mutation is the most concerning issue -- worth addressing before merge.

account_statement_import_adyen/wizards/account_statement_import.py
NL66278: This has been corrected

account_statement_import_adyen/wizards/account_statement_import_adyen_parser.py
NL66278: I went for a little bit different solution to prevent modification of the COLUMNS dict. Basically determine the position of all valid/known columns, and using that to turn every row into a dict. This also makes the remainder of the code a lot less verbose in accessing values in a row (or in a cell actually).

NL66278:
@alexey-pelykh I took up your points, and added some additional tests as well. Could you have another look and if satisfied merge?

NL66278:
@OCA/banking-maintainers this is approved and ready for merge

etobella:
Hi Ronald,

There is several changes on other modules, not just account_statement_import_online_adyen

NL66278:
> Hi Ronald,
> 
> There is several changes on other modules, not just account_statement_import_online_adyen

@etobella I know, as I stated in the description of this PR. Actually this is a migration of three modules that work together. To get everything working I had to change exactly one line in a fourth module.

lfreeke:
@etobella  Are you satisfied with @NL66278  answer, or do you still have any concerns?

etobella:
My comment was non blocking, Usually it is easier when each module has it's own PR, so it is easier to get reviews

NL66278:
@OCA/banking-maintainers Can somebody merge this one?

account_statement_clearing_account/models/account_bank_statement.py
```diff
@@ -0,0 +1,86 @@
...
+    def get_reconcile_clearing_account_lines(self):
+        """If this statement qualifies for clearing account reconciliation,
+        return the relevant lines to (un)reconcile. This is the case if the
+        default journal account is reconcilable, each statement line has a
+        counterpart line on this account for the full amount and the sum of
+        the counterpart lines is zero.
+        """
+        self.ensure_one()
+        if (
+            self.journal_id.default_debit_account_id
```
alexey-pelykh, line 19: `default_debit_account_id` / `default_credit_account_id` don't exist on 16.0 `account.journal` — they were consolidated into a single `default_account_id` (`account/models/account_journal.py` on 16.0 has `default_account_id` + `suspense_account_id`). This condition, and `account = …default_debit_account_id` below, will `AttributeError` whenever this method runs.

account_statement_clearing_account/models/account_bank_statement.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def button_reopen(self):
+        """When setting the statement back to draft, unreconcile the
+        reconciliation on the clearing account"""
+        res = super().button_reopen()
+        for statement in self:
+            statement.unreconcile_clearing_account()
+        return res
+
+    def button_confirm_bank(self):
```
alexey-pelykh, line 80: `button_confirm_bank` (and `button_reopen` above) no longer exist on 16.0 `account.bank.statement` — the confirm/state workflow was removed in the 16.0 accounting rewrite. `super().button_confirm_bank()` has no parent to call, and nothing in core invokes these overrides, so `reconcile_clearing_account()` is never triggered on 16.0. This is the confirm/reconcile flow I asked about in March; it needs rehooking to the 16.0 reconciliation model.

account_statement_clearing_account/tests/test_clearing_account.py
```diff
@@ -0,0 +1,92 @@
...
+            }
+        )
+        cls.partner_customer = cls.env["res.partner"].create({"name": "customer"})
+        cls.partner_provider = cls.env["res.partner"].create({"name": "provider"})
+
+    def test_reconcile_unreconcile(self):
+        """Test that a bank statement that satisfies the conditions, can be
+        automatically reconciled.
+        """
+        # TODO bank statement no longer has a state, nor a confirm button.
```
alexey-pelykh, line 40: These TODOs are the crux — the test only asserts the statement balances to zero, never the reconcile itself, so the broken paths above go uncovered, which is what `codecov/patch` is flagging. A test that drives the actual 16.0 reconcile would both prove the feature and clear the coverage gate.

account_statement_import_online_adyen/models/online_bank_statement_provider.py
```diff
@@ -0,0 +1,144 @@
...
+            lambda r: r.service not in ("adyen", "dummy_adyen")
+        )
+        if other_providers:
+            result = super(OnlineBankStatementProvider, other_providers)._pull(
+                date_since, date_until
+            )
+        for provider in adyen_providers:
+            is_scheduled = self.env.context.get("scheduled")
+            try:
+                self._import_adyen_file()
```
alexey-pelykh, line 46: Should this be `provider._import_adyen_file()`? `_import_adyen_file` calls `self.ensure_one()`, and the base `_pull` is called on multi-record sets (`for provider in self`), so `self._import_adyen_file()` raises once more than one provider is pulled together.

alexey-pelykh (commented):
Thanks for pushing this forward, Ronald. I re-checked against head `d9fc11a` and the points I raised in March are addressed: the `COLUMNS` dict is no longer mutated, the filename-fallback precedence is fixed, and the README badges point to 16.0. The `attrs` in the online-provider view is fine to leave — the inline `invisible="…"` form is 17.0+, so `attrs` is correct on 16.0.

I can't recommend merge yet, though. My earlier ask to confirm the clearing-account reconcile flow on 16.0 turns out to be a real problem, not just a coverage gap — `account_statement_clearing_account` still targets the pre-16.0 accounting model:

- it reads `journal_id.default_debit_account_id` / `default_credit_account_id`, which no longer exist on 16.0 `account.journal` (consolidated into `default_account_id`);
- it overrides `button_confirm_bank` / `button_reopen`, which no longer exist on 16.0 `account.bank.statement`, so the reconcile logic never actually runs.

So the module installs but its core feature is inert on 16.0 — which is also why `codecov/patch` is red. Two clean ways forward: (a) rewire the reconcile to the 16.0 model and add a test that exercises it (also clears codecov/patch); or (b) if the clearing reconcile is out of scope for this migration, split it out and get PSC agreement to merge the working Adyen import path on its own. Happy to help either way.

One small thing in the online provider regardless (inline).

Co-Reviewed-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #631, opened 2023-10-25
add informations from additionnal node

davidwul:
A Danish (Sydbank) bank is setting the additional information in <NtryDtls><TxDtls><BkTxCd><Prtry><Cd> node


ecino (approved):
Nothing to complain about in my opinion. Could we merge?

ecino:
@davidwul Could you also propose the change on newer versions?

ecino:
@davidwul I think you can close this one, #632 is the same.

PR #635, opened 2023-11-03
[15.0][ADD] account_statement_import_online_gocardless

pedrobaeza:
This module provides online bank statements from GoCardless Bank Account Data, which provides a free API for connecting and getting transactions for bank accounts.

@Tecnativa TT45760


pedrobaeza:
@ChrisOForgeFlow @JordiBForgeFlow this one is the spiritual successor to #496, due to the Nordigen absortion by GoCardless. It includes:

- The new endpoint.
- Simplified code.
- Better token renewal.
- Proper documentation.
- Test (although mocked).

I haven't preserved the commit history, as all the renamings of variables will make a huge diff. I have put you instead as co-authors. Is it OK for you?

JordiBForgeFlow:
Works for us!

Rad0van:
Nice one! Thanx a lot @pedrobaeza & @JordiBForgeFlow !!! Thank you very much. As soon as it's merged we'll port to 16 (we'll start now to be ready). Wanna try it in our Odoo...

achulii:
@pedrobaeza Hello! We are porting this module to version 16, as @Rad0van mentioned earlier. In version 16 the situation is like this, ondelete="cascade" is no longer there for statement_id in `account.bank.statement.line` (was there up until 15 - removed with this commit https://github.com/odoo/odoo/commit/4fb0933986534e31337aa1dce3478a67deea72de#diff-96c5c4fc954f9deb57474d052d9c59b68cc41b610205d342949727ece08ac404 ).  When I manually delete `account.bank.statement`, related `line_ids` will remain untouched, and at every next import, they will be ignored (because of `unique_import_id`). Can you maybe suggest any ideas on how to treat this? Or should I just manually delete appropriate `line_ids` as well?

pedrobaeza:
IMO, this should be handled at `account_statement_base` (https://github.com/OCA/account-reconcile/tree/16.0/account_statement_base), re-adding such constraint or adding an extra statement lines menu entry/shortcut to it.

Rad0van:
> IMO, this should be handled at `account_statement_base` (https://github.com/OCA/account-reconcile/tree/16.0/account_statement_base), re-adding such constraint or adding an extra statement lines menu entry/shortcut to it.

I was opposed to re-adding the constraint (if you mean `ondelete="cascade"`) because I thought it was removed for some reason - i.e. they wanted to be able to retain statement lines even without the statement itself. Was afraid it could break something where they rely on the lines being present even without statement itself. Cannot really imagine what the scenario could be though.

If you're talking about extra menu entry/shortcut, you mean something like a filter for "orphaned" statement lines?

pedrobaeza:
I mean having a view to manually see/delete statement lines.

carlosdauden (approved):

victoralmau (approved):

PR #648, opened 2023-12-18
[16.0][FIX] models/online_bank_statement_provider_ponto.py: prevent a…

vaab:
…ttempt to modify of posted account.move

We have cases where the setattr on `line.partner_id` triggers modification on the associated `move_id` that is already in `"posted"` `state`. And this is refused. 

Please note that actually in our cases, even thou the `line._retrieve_partner()` value is still empty, the simple act of triggering the `line.partner_id` setattr is enough to get the exception.

We know this is not the proper fix, and we are not asking to integrate this PR, but wanted to have a discussion on this point. I'm not fully knowledgeable on the functional aspects that should be considered.

Many thanks for your feedback on this one and we'll be happy to contribute if you have any direction to suggest a better fix.

### DIFF
--- a/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
+++ b/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
@@ -169,11 +169,9 @@ class OnlineBankStatementProvider(models.Model):
         # For each period, create or update statement lines
         for period, statement_lines in grouped_periods.items():
             (date_since, date_until) = period
-            statement = self._create_or_update_statement(
+            self._create_or_update_statement(
                 (statement_lines, {}), date_since, date_until
             )
-            for line in statement.line_ids.filtered(lambda l: not l.partner_id):
-                line.partner_id = line._retrieve_partner()
 
     def _ponto_get_transaction_vals(self, transaction):
         """Translate information from Ponto to statement line vals."""


StefanRijnhart:
I ran into the same issue when I configured the module on a high traffic account. The issue occurs when newly imported transactions are grouped into a statement that contains transactions that are already reconciled.
```
  /home/odoo/user/oca-bank-statement-import/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py(56)_pull()
-> provider._ponto_pull(date_since, date_until)
  /home/odoo/user/oca-bank-statement-import/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py(86)_ponto_pull()
-> self._ponto_store_lines(lines)
  /home/odoo/user/oca-bank-statement-import/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py(168)_ponto_store_lines()
-> line.partner_id = line._retrieve_partner()
  /home/odoo/odoo/odoo/fields.py(1325)__set__()
-> records.write({self.name: write_value})
  /home/odoo/odoo/addons/account/models/account_bank_statement_line.py(380)write()
-> self._synchronize_to_moves(set(vals.keys()))
  /home/odoo/odoo/addons/account/models/account_bank_statement_line.py(832)_synchronize_to_moves()
-> st_line.move_id.write(st_line_vals)
  /home/odoo/odoo/addons/base_automation/models/base_automation.py(411)write()
-> write.origin(self.with_env(actions.env), vals, **kw)
  /home/odoo/enterprise/documents_account/models/account_move.py(29)write()
-> res = super().write(vals)
  /home/odoo/odoo/addons/purchase/models/account_invoice.py(148)write()
-> res = super(AccountMove, self).write(vals)
  /home/odoo/odoo/addons/account/models/account_move.py(2345)write()
-> )).write(vals)
  /home/odoo/odoo/addons/mail/models/mail_thread.py(309)write()
-> return super(MailThread, self).write(values)
  /home/odoo/odoo/addons/mail/models/mail_activity_mixin.py(241)write()
-> return super(MailActivityMixin, self).write(vals)
  /home/odoo/odoo/odoo/models.py(3776)write()
-> field.write(self, value)
  /home/odoo/odoo/odoo/fields.py(4265)write()
-> return self.write_batch([(records, value)])
  /home/odoo/odoo/odoo/fields.py(4286)write_batch()
-> return self.write_real(records_commands_list, create)
  /home/odoo/odoo/odoo/fields.py(4485)write_real()
-> flush()
  /home/odoo/odoo/odoo/fields.py(4437)flush()
-> comodel.browse(to_delete).unlink()
  /home/odoo/odoo/addons/sale_timesheet/models/account_move.py(131)unlink()
-> return super().unlink()
  /home/odoo/odoo/addons/account/models/account_move_line.py(1578)unlink()
-> res = super().unlink()
  /home/odoo/odoo/odoo/models.py(3560)unlink()
-> func(self)
> /home/odoo/odoo/addons/account/models/account_move_line.py(1546)_unlink_except_posted()
-> raise UserError(_('You cannot delete an item linked to a posted entry.'))
```

Removing the `_retrieve_partner` calls is the right solution I think. The OCA reconciliation module, as well as the enterprise reconciliation calls this method just-in-time anyway, which is only appropriate as the counterpart entry might not yet be present in Odoo at the moment of banks statement import.

StefanRijnhart:
@vaab Can you rebase and simply delete the lines?

njeudy:
@StefanRijnhart @vaab  code rebased and precommit Ok now :)

StefanRijnhart (approved):
@astirpe and/or @NL66278 can you live with this?

account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
```diff
@@ -169,11 +169,9 @@ def _ponto_store_lines(self, lines):
         # For each period, create or update statement lines
         for period, statement_lines in grouped_periods.items():
             (date_since, date_until) = period
-            statement = self._create_or_update_statement(
+            self._create_or_update_statement(
                 (statement_lines, {}), date_since, date_until
             )
-            for line in statement.line_ids.filtered(lambda l: not l.partner_id):
```
NL66278, line 175: IMHO getting the partner for a line is an important functionality that should not be discarded when not needed. I understand the problem with already reconciled / posted entries, but I would suggest to enhance the filter here to leave the lines out for where that is the case.

StefanRijnhart: As I tried to explain above, assigning a partner before the entry can be reconciled can be harmful because it sets the partner_id on the statement line, and [when the partner_id is set on the statement line it takes precedence in `_retrieve_partner`](https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_bank_statement_line.py#L638-L643) and is never updated by it.

It may be the case that the matching entry is not present in the system at the moment of bank statement import. In that case, another partner can take precedence and if that partner is hard linked to the transaction as is currently the case it will be very difficult to make the right (automatic) reconciliation.

It is therefore more appropriate to try and retrieve the partner at the moment of reconciliation and only store it on the statement line when the reconciliation succeeds. This is what Odoo Enterprise does.

NL66278: @StefanRijnhart OK, I am more or less convinced. Mainly because the _retrieve_partner method also does an ilike match on name, that to me seems quite unreliable. I only do not understand the reference to the lines 638-643, as this will not be triggered by the calling ponto import code, because that filters out all lines where partner already has been set.

NL66278 (approved):
:+1: LGTM

account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
pedrobaeza: If the partner is clear, why postpone it to the reconciliation phase. And is there any automatic partner matching on that case? In my experience, it isn't. Anyway, I have deprecated the use of MyPonto in favor of GoCardless (free up to 50 accounts), so it's just a general opinion.

PR #650, opened 2023-12-19
[15.0][FIX] account_statement_import_online_gocardless: Unique ID fallback

pedrobaeza:
Not all banks provide an entry reference or transaction ID, so we need to fallback to the Gocardless internal ID for getting the unique ID in such cases.

@Tecnativa TT46640

### DIFF
--- a/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
+++ b/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
@@ -312,8 +312,11 @@ class OnlineBankStatementProvider(models.Model):
                     "payment_ref": tr.get(
                         "remittanceInformationUnstructured", partner_name
                     ),
-                    "unique_import_id": tr.get("entryReference", False)
-                    or tr.get("transactionId", False),
+                    "unique_import_id": (
+                        tr.get("entryReference")
+                        or tr.get("transactionId")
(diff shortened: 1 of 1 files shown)


victoralmau (approved):

account_statement_import_online_gocardless/models/online_bank_statement_provider.py
```diff
@@ -312,8 +312,11 @@ def _gocardless_obtain_statement_data(self, date_since, date_until):
                     "payment_ref": tr.get(
                         "remittanceInformationUnstructured", partner_name
                     ),
-                    "unique_import_id": tr.get("entryReference", False)
-                    or tr.get("transactionId", False),
+                    "unique_import_id": (
+                        tr.get("entryReference")
+                        or tr.get("transactionId")
+                        or tr.get("internalTransactionId")
```
chienandalu, line 318: Is it ok to fallback to `None`?

pedrobaeza: Yes, `None` is the same as `False` at ORM level. But any of the 3 should have a value. That's why we add this third one, as if no ID, each download, the transactions will be duplicated.

chienandalu (approved):

PR #651, opened 2023-12-20
[13.0][IMP] *_online_paypal: make sure unique_import_id within download is unique...

NL66278:


alexey-pelykh:
@NL66278 Please elaborate a little bit more - have you checked with PayPal how their ID is not unique?

NL66278:
@alexey-pelykh This happens when a refund is processed. We get three lines on the update of the "general hold" lines, having the same transaction idea and - as it is one transaction - also the last transaction update date/time. @lfreeke Maybe you can elaborate on this?

alexey-pelykh:
@NL66278 Thanks for the explanation! In this case I don't think that hashing is a proper approach. I've encountered same thing with Wise.com, ended up suffixing with type of operation like `123412341234` and `123412341234-FEE` in that case. Is that something possible with Paypal data?

NL66278:
@alexey-pelykh In the past I have found hashing to be a very reliable way to really produce unique id' s, that still prevent duplicate downloads of the same data (used this as an alternative for downloaded camt files that for that customer resulted in overlapping data because of multiple downloads per day). I am however open to any other way to solve this issue. I will ask for the complete set of data to see whether I can find something else to make the id really unique.

alexey-pelykh:
@NL66278 I guess we totally on a fact that hashing is great for fingerprinting data for deduplication, yet in this case I appeal to attempt to use API-sources unique ID - it's just a bummer that PayPal does not treat transaction ID as truly unique. Thanks in advance for looking into that!

NL66278:
@alexey-pelykh Instead of hashing I add event_code to make unique_import_id indeed unique. Customer confirmed that this works.

alexey-pelykh (approved):

alexey-pelykh:
@NL66278 one thing that would be an issue - this would be a breaking change and would duplicate all already imported transactions. Not sure how to handle that though

NL66278:
@alexey-pelykh I went for another solution, that minimizes the changes and will not give a possibility to duplicate imports. Basically the problem is that within a download one transaction might have multiple lines (with different event code). Instead of always adding the event code, I will leave the first occurrence of a key in the imported lines unchanged. This will ensure no duplicate download happens. Subsequent duplicate keys **within the import**, get a number affixed to make them unique.

dreispt (approved):

PR #652, opened 2023-12-26
[15.0][FIX] account_statement_import_online_gocardless: Fallback to journal company

pedrobaeza:
If the bank account is shared across multiple companies (no assigned company_id), we can't launch the "Select Bank Account Identifier" wizard, so we fallback to the journal company for getting the country.

@Tecnativa TT43849


chienandalu (approved):

pedrobaeza:
Let's fast-track this easy one:

/ocabot merge patch

PR #656, opened 2024-01-04
[16.0] bank_statement_import_camt: Alternatively support datetime (DtTm) as booking date

ayushin:
Certain institutions (wise.com) use DtTm instead of Dt for transaction date, this fix supports either original Dt or DtTm


hbrunn:
@ayushin I've recovered your code in https://github.com/OCA/bank-statement-import/pull/842, squashed your commits and slightly changed the commit message. If you can review that, we're already halfway there to get this merged

PR #660, opened 2024-01-17
[15.0][IMP] account_statement_import_online: Add debug wizard

pedrobaeza:
It's useful to debug RAW data received from your online sources. With this improvement, you can do it easily from the manual pull wizard. It's only available for users in debug mode (or with technical features).

@Tecnativa TT47238


account_statement_import_online/wizards/online_bank_statement_pull_wizard.xml
```diff
@@ -28,6 +28,12 @@
                         default_focus="1"
                         class="btn-primary"
                     />
+                    <button
+                        name="action_debug"
+                        string="Debug"
```
StefanRijnhart: I think the label on this one can be a little bit more descriptive (accountants also use debug mode to see more features, not just developers). Maybe `Show transaction data` plus a more explanatory help text on this one and then replace `Pull` on the other button by something like `Import transactions`?

StefanRijnhart (commented):
Very useful! I'm porting this to 16.

carlosdauden (approved):

account_statement_import_online/wizards/online_bank_statement_pull_wizard.xml
StefanRijnhart: My suggestions here: https://github.com/stefanrijnhart/bank-statement-import/commit/6d104bd597bbde65a3542c81e365acb5b4b4e665

PR #665, opened 2024-02-02
[15.0][FIX] account_statement_import_online_gocardless: Payment reference extraction

pedrobaeza:
Since one week ago, banks like Sabadell is sending the information in the JSON list `remittanceInformationUnstructuredArray` instead of `remittanceInformationUnstructured`.

Thus, we need to parse both for finding the proper element. For the rest that are using the previous one, there's no change.

@Tecnativa TT47491


pedrobaeza:
Fast-tracking this for solving the problem for all affected, as this is harmless for the rest:

/ocabot merge patch

PR #666, opened 2024-02-05
[IMP] account_statement_import_camt: multi account support

gurneyalex:
The module account_statement_import_file supports handling files containing informations related to multiple accounts, but this was not taken advantage of by account_statement_import_camt.

We update the structure returned by parser.parse() to use the expected format in case of multiple accounts encountered in CAMT file.

This has the side effect of fixing a bug in the previous implementation, when all the transactions in a CAMT file would be parsed as being on the same account (the one on the last statement found in the file), and possibly even more wrong, in the currency of the last statement).


StefanRijnhart (approved):
Nice one!

StephaneMangin (approved):

StefanRijnhart:
@gurneyalex any chance of a rebase to resolve conflicts?

PR #641, opened 2023-12-14
[16.0][MIG][WIP] account_statement_import_online_gocardless: Migration to 16.0

AlexPForgeFlow:

PR #667, opened 2024-02-05
[16.0][MIG] account_statement_import_online_gocardless

njeudy:
Using initial forgeflow migration PR, and add some update


njeudy:
continue the work for https://github.com/OCA/bank-statement-import/pull/641

njeudy:
@pedrobaeza do i need to add a timeout to request ?

pedrobaeza:
Yes, please. There are missing commits that has landed in 15.0 branch later than the original PR. Please cherry-pick them. You need to reorder the commits into a logical sequence (the migration commits to be the last), and please remember other times to put correctly the PR title and don't let the default proposed one.

njeudy:
@pedrobaeza I push all, and decide to set timeout to 10 min because of the first import that could or for some big bank statement file .. what do you think ? I can add a resconfig parameter to if needed ..

pedrobaeza:
Thanks. That huge timeout is not really needed. Take into account that the timeout is just for the HTTP request, not for all process. This takes at maximum seconds, and although you ask for whole years, the requests are split by the statement interval set in the journal. And I think this timeout is only for getting in contact, not for finishing whole request.

I think 5 seconds is the ideal timeout.

njeudy:
ok I change to 5 sec and will see if any issue with this .. Thanks for your help ..

njeudy:
@pedrobaeza I finish the migration, rebase all commits and test it with real sync (gocardlesse <-> qonto) and all is ok for me.
Now, some test need to be done as I see, but don't know how to test it because of api call ..

account_statement_import_online_gocardless/tests/test_account_statement_import_online_gocardless.py
```diff
@@ -80,7 +80,6 @@ def setUpClass(cls):
 
     def test_mocked_gocardless(self):
         vals = {
-            "provider_ids": self.provider.ids,
```
pedrobaeza: Why removing this?

pedrobaeza (commented):
I don't see the timeouts applied. Please squash latest commit into the migration one.

njeudy:
I squash all commit. Timeout are set in request params dans defined by default on top of py file ...

njeudy:
@alexis-via can you check this, I think it can replace myponto :) I need help to add test on this

account_statement_import_online_gocardless/tests/test_account_statement_import_online_gocardless.py
pedrobaeza: Seeing the tests, now I see why you removed it. I was just doing the question for knowing it. Anyway, I see that you were the one removing that field:

https://github.com/OCA/bank-statement-import/commit/bd10c2f59e4c37cc9b9c400bd3eab3a192b257c0#diff-724a1bdf113d58c503e61bbc4e54e4416dc30cc2246214f19bbe2e521539eadeL24

Please remove it again.

pedrobaeza:
@njeudy I'm not able to push to your branch. Please remove again the line `"provider_ids": self.provider.ids,` and this is good to be merged.

njeudy:
@pedrobaeza done, thanks for your time !

pedrobaeza (approved):
/ocabot merge nobump

PR #668, opened 2024-02-06
[16.0][IMP] account_statement_import_online: only update journal when online service is set

StefanRijnhart:
This prevents a flood of logs in case your journals are written very often.

### DIFF
--- a/account_statement_import_online/models/account_journal.py
+++ b/account_statement_import_online/models/account_journal.py
@@ -77,7 +77,8 @@ class AccountJournal(models.Model):
     def write(self, vals):
         self._update_vals(vals)
         res = super().write(vals)
-        self._update_providers()
+        if vals.get("online_bank_statement_provider"):
+            self._update_providers()
         return res
 
     def _update_vals(self, vals):


alexey-pelykh (approved):

pedrobaeza (approved):
/ocabot merge patch

Does this apply to lower versions?

StefanRijnhart:
@pedrobaeza No, it was a glitch during the migration to 16: https://github.com/OCA/bank-statement-import/commit/bd10c2f59e4c37cc9b9c400bd3eab3a192b257c0#diff-90c6d4b206ebf24081d715a75e8f7f950f10cd4e2f6cfb95829e47b3727da06fR79-L84

The `bank_statements_source` part is now handled by `_update_vals` which is harmless to call on every write, and with this PR the `online_bank_statement_provider` check is restored.

pedrobaeza:
Then shouldn't we add a similar check in the `create`: https://github.com/OCA/bank-statement-import/commit/bd10c2f59e4c37cc9b9c400bd3eab3a192b257c0#diff-90c6d4b206ebf24081d715a75e8f7f950f10cd4e2f6cfb95829e47b3727da06fR76 ?

StefanRijnhart:
@pedrobaeza It's not worth the effort. Most likely the online import is not configured immediately on create (in which case the journal is skipped in `_update_providers`), and if it is, then either the action in `_update_providers` is appropriate, or it will emit a single INFO log saying that the journal is already configured properly.

pedrobaeza:
Ok then

PR #673, opened 2024-03-06
[16.0][FIX] account_statement_import_camt: fix label issue

ByteMeAsap:
Add more info on payment reference instead of '/' when more info available

### DIFF
--- a/account_statement_import_camt/models/parser.py
+++ b/account_statement_import_camt/models/parser.py
@@ -61,7 +61,6 @@ class CamtParser(models.AbstractModel):
             node,
             [
                 "./ns:RmtInf/ns:Ustrd|./ns:RtrInf/ns:AddtlInf",
-                "./ns:AddtlNtryInf",
                 "./ns:Refs/ns:InstrId",
             ],
             transaction,
@@ -329,6 +328,7 @@ class CamtParser(models.AbstractModel):
 
         details_nodes = node.xpath("./ns:NtryDtls/ns:TxDtls", namespaces={"ns": ns})
         if len(details_nodes) == 0:
+            self.amend_transaction(transaction)
             self.generate_narration(transaction)
             yield transaction
             return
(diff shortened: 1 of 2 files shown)


bosd (approved):
LGTM

account_statement_import_camt/models/parser.py
```diff
@@ -294,6 +293,14 @@ def parse_entry(self, ns, node):
...
             "%s (AddtlNtryInf)" % _("Additional Entry Information"),
         )
+        self.add_value_from_node(
+            ns,
+            node,
+            "./ns:AddtlNtryInf",
+            transaction,
+            "payment_ref",
+            join_str="\n",
+        )
```
hbrunn: this is parsed into the `narration` field just above your change, so I don't think we should duplicate this data by default.

You can better add a function `amend_transaction` or such which is called at the end of this function before `generate_narration`, where the `payment_ref` field is set from some field(s) in the narration if it's unset (/). This will also be an easy hook for custom code to change this behavior to a customer's sepcs, which currently people do in `generate_narration` I guess.

hbrunn (changes requested):

account_statement_import_camt/models/parser.py
ByteMeAsap: @hbrunn ,thanks for your inputs!!. I have made changes as you suggested

account_statement_import_camt/models/parser.py
```diff
@@ -330,12 +329,14 @@ def parse_entry(self, ns, node):
         details_nodes = node.xpath("./ns:NtryDtls/ns:TxDtls", namespaces={"ns": ns})
         if len(details_nodes) == 0:
             self.generate_narration(transaction)
+            self.amend_transaction(transaction)
```
hbrunn, line 331: please move this before the call to `generate_narration`, at this point the field is already flattened to a string

hbrunn (changes requested):
thanks for the quick followup!

account_statement_import_camt/models/parser.py
ByteMeAsap: @hbrunn , done!

hbrunn (approved):

account_statement_import_camt/models/parser.py
hbrunn: thanks

ByteMeAsap:
@pedrobaeza , can you please get this merged?

PR #676, opened 2024-03-15
[16.0][IMP] account_statement_import_camt: refactor tests for inheritability

SilvioC2C:
This refactoring will allow using `TestParser._do_parse_test()` method for subclasses defined in inheriting modules.

This way, if the new module defines new files in other paths than `account_statement_import_camt/test_files/`, subclasses of `TestParser` will still be able to use the `_do_parse_test()` method for imports and checks.

Moreover, a custom value can now be defined for the acceptable differing lines number between inputfile and goldenfile.

It will also maintain backward compatibility with the current usage of the method (ie: using only a filename with the assumption of finding it in `account_statement_import_camt/test_files/`).

NB: methods are now defined outside of `TestParser` to allow importing them directly, instead of inheriting from `TestParser`.

### DIFF
--- a/account_statement_import_camt/tests/test_import_bank_statement.py
+++ b/account_statement_import_camt/tests/test_import_bank_statement.py
@@ -7,42 +7,80 @@ import difflib
 import pprint
 import tempfile
 from datetime import date
+from pathlib import Path
 
 from odoo.modules.module import get_module_resource
 from odoo.tests.common import TransactionCase
 
 
-class TestParser(TransactionCase):
-    """Tests for the camt parser itself."""
-
+class TestParserCommon(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         cls.parser = cls.env["account.statement.import.camt.parser"]
 
-    def _do_parse_test(self, inputfile, goldenfile):
-        testfile = get_module_resource(
-            "account_statement_import_camt", "test_files", inputfile
+    def _do_parse_test(self, inputfile, goldenfile, max_diff_count=None):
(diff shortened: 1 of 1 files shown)


simahawk (commented):
this method seems not tied at all to the class: I would make it a simple function to call from everywhere w/o having to inherit this class (which btw has tests and will be inherited as well).

simahawk (changes requested):
the other options is to make it a class method but is pointless IMO

SilvioC2C:
>this method seems not tied at all to the class: I would make it a simple function to call from everywhere w/o having to inherit this class (which btw has tests and will be inherited as well).

I'll create an external function that can be safely imported by other modules, and then the class's method will simply redirect to it

account_statement_import_camt/tests/test_import_bank_statement.py
```diff
@@ -20,29 +61,34 @@ def setUpClass(cls):
         super().setUpClass()
         cls.parser = cls.env["account.statement.import.camt.parser"]
 
-    def _do_parse_test(self, inputfile, goldenfile):
-        testfile = get_module_resource(
-            "account_statement_import_camt", "test_files", inputfile
+    def _do_parse_test(self, inputfile, goldenfile, max_diff_count=None):
```
simahawk, line 22: if your intent is to still use the class you should move this to a mixin or common setup class otherwise your inheriting class will run the same tests

SilvioC2C:
@simahawk mind reviewing?

simahawk (approved):

PR #677, opened 2024-03-17
[16.0][FIX] account_statement_import_online_gocardless: Increase timeout

pedrobaeza:
It turns out that the timeout is for receiving whole answer, so in real tests, the previous timeout was not enough for getting the answer when there are some transactions load (like a month).

Let's increase this timeout then to a reasonable amount.

@Tecnativa


pedrobaeza:
Merging directly as very straight-forward:

/ocabot merge patch

PR #679, opened 2024-04-03
[16.0][IMP] account_bank_statement_import_online: create not override in batch

daiduongnguyen-odoo:


pedrobaeza:
Tag should be `[IMP]`, not `[MISC]`.

daiduongnguyen-odoo:
> Tag should be `[IMP]`, not `[MISC]`.
Longtime no see
ok, actually i intend to make a pr for my fork, but it's ok anyway,

pedrobaeza:
It's good to have it here in OCA and don't need to maintain a fork for these things. And even if it's something private, that you stick to the same conventions.

You have renamed the title of the PR, but not the commit message.

daiduongnguyen-odoo:
> It's good to have it here in OCA and don't need to maintain a fork for these things. And even if it's something private, that you stick to the same conventions.
> 
> You have renamed the title of the PR, but not the commit message.

Your guy have some greate addons, i really appreciate that, but some need further customize for our need so we did it on the fork, But thanks anyway

pedrobaeza:
I think modifying existing modules in a fork is not a good strategy, as you need to maintain them continuously updating from upstream. Instead, what you can do is to add extra custom modules on top of these ones, reducing the maintenance, and adding here hooks if you need to. Obviously, this is just an advice.

pedrobaeza (approved):
Fast-tracking it, as obvious patch:

/ocabot merge patch

daiduongnguyen-odoo:
> I think modifying existing modules in a fork is not a good strategy, as you need to maintain them continuously updating from upstream. Instead, what you can do is to add extra custom modules on top of these ones, reducing the maintenance, and adding here hooks if you need to. Obviously, this is just an advice.

Indeed, many times my boss keep merge from upstream and remove all of my customize and i had to re add them :)

PR #683, opened 2024-04-22
[15.0][IMP] account_statement_import_online_gocardless: Wizard for choosing if reusing or not credentials

pedrobaeza:
Some banks only allow one credentials, while others work better with separate ones, so it's a matter of trying. This commit introduces a wizard for choosing one method or the other.

imagen

@Tecnativa TT48828


carolinafernandez-tecnativa:
1) When i go to Select account bank identifier, i got this error message:

image


image

image

2) When i press "Pull online bank statement" i got this error message, when selecting Show transaction data

image

image

3) If i select multiple journals and press Action

image

I got this error message:

image

pedrobaeza:
You can't do anything without the GoCardless credentials. It's true that the button may be hidden if no credentials, so I have added an extra commit for that.

carolinafernandez-tecnativa (approved):
LGTM, Thanks :) Please fw to v16

pilarvargas-tecnativa (approved):
:+1:

PR #685, opened 2024-04-22
[16.0][IMP] account_statement_import_online_gocardless: Wizard for choosing if reusing or not credentials + hide button

pedrobaeza:
Forward-port to v16 of #683 

Some banks only allow one credentials, while others work better with separate ones, so it's a matter of trying. This commit introduces a wizard for choosing one method or the other.

imagen

It includes also a patch for button conditional visibility.

@Tecnativa TT48828


carolinafernandez-tecnativa (approved):
LGTM :) Thanks

pilarvargas-tecnativa (approved):
thank you! :smiley:

PR #688, opened 2024-04-30
[14.0][IMP]account_statement_import_txt_xlsx: additional mapping controls

PicchiSeba:
These changes allow to skip a specified amount of rows/columns during the file parsing phase.
There is also an option that allows to skip completely empy lines.

Please let me know what you think about it and if you like these changes to be made differently.


account_statement_import_txt_xlsx/models/account_statement_import_sheet_mapping.py
```diff
@@ -61,6 +62,23 @@ class AccountStatementImportSheetMapping(models.Model):
...
     )
+    skip_empty_lines = fields.Boolean(
+        "Skip Empty Lines",
+        default=False,
+        help="Allows to skip empty lines",
+    )
+    offset_column = fields.Integer(
+        "Offset Column",
+        default=0,
+        required=True,
```
alexey-pelykh: Why do you need `required`? Especially if the field is optionally hidden

PicchiSeba: @alexey-pelykh 
Thanks for pointing that out. I forgot about it from an earlier iteration

alexey-pelykh (changes requested):
Please add tests to cover this use-case and I'll proceed with approving

PicchiSeba:
@alexey-pelykh I added a couple of test that cover the new features.
Are they good enough or do we require more in-depth test scenarios?

alexey-pelykh (approved):

alexey-pelykh:
Please take some time to back- and forward- port the improvement.

PR #689, opened 2024-05-01
[16.0][IMP] account_statement_import_online_gocardless: prefer valueDate over bookingDate

Rad0van:
This fixes behavior when there are debit card transactions with bookingDate behind the actual valueDate. In current situation this leads to a situation where these are later discarded in here:
https://github.com/OCA/bank-statement-import/blob/e91e28117d5ae3fbb5e1fb168b734f6690689433/account_statement_import_online/models/online_bank_statement_provider.py#L345-L351

Here's a screenshot of such situation.
gocardless_issue

### DIFF
--- a/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
+++ b/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
@@ -316,7 +316,7 @@ class OnlineBankStatementProvider(models.Model):
         currencies_cache = {}
         for tr in transactions.get("transactions", {}).get("booked", []):
             # Reference: https://developer.gocardless.com/bank-account-data/transactions
-            string_date = tr.get("bookingDate") or tr.get("valueDate")
+            string_date = tr.get("valueDate") or tr.get("bookingDate")
             # CHECK ME: if there's not date string, is transaction still valid?
             if not string_date:
                 continue


Rad0van:
Aby comments? GitHub is telling me it's going to be closed due to inactivity..m

droidosergioh:
Hello,

Any news about it? This fixes an error we are facing.

Kindest Regards

pedrobaeza:
Sorry for not being aware of this, and this explains why I lost some transactions in some of our customers without explanation. Resuming this in #782 with the proper commit message.

pedrobaeza:
@Rad0van I'm having troubles with this change, as my customers ask for having the previous date in the transaction, not the new one. Have you got troubles also with this?

Rad0van:
Hi,I'm mostly using myself and it works fine for me. I suggest we add configuration option that would allow to choose the behavior. What is your opinion? Best regards,   RadovanOn Aug 8, 2025 16:42, "Pedro M. Baeza" ***@***.***> wrote:pedrobaeza left a comment (OCA/bank-statement-import#689)
@Rad0van I'm having troubles with this change, as my customers ask for having the previous date in the transaction, not the new one. Have you got troubles also with this?

pedrobaeza:
@Rad0van I have just tried to obtain transactions in both dates, booking date or value date, and GoCardless return the transaction in both of them, so the problem you depicted is not really there.

Rad0van:
> @Rad0van I have just tried to obtain transactions in both dates, booking date or value date, and GoCardless return the transaction in both of them, so the problem you depicted is not really there.

I am not sure I follow. If there is the same date this change should not affect anything. Or am I missing something?

On the other hand I have faced the situation as described in the PR and it was confirmed by Sergio here: https://github.com/OCA/bank-statement-import/pull/689#issuecomment-2382739514 In the original state it would just discard operations like these - have a look at the screenshot I provided. As mentioned - this was in my case specific to debit card transactions. bookingDate in such case is when the transaction was done and valueDate holds later date when it was processed (or something).

pedrobaeza:
I think I know the problem you are depicting: the booking date in your case is previous to the value date, so your transaction wasn't available to download until value date. In my case, it's the contrary: the booking date is after the value date, and thus the transaction is ready, but we want to book it on the booking date. Real example: booking date is 2025-08-06, and value date 2025-07-30.

I'm adding the date type selector.

Rad0van:
It also may be that your / my / both banks use these fields incorrectly. I have just checked the documentation and it says:

> bookingDate = The date when an entry is posted to an account on the financial institutions books.
> valueDate = The Date at which assets become available to the account owner in case of a credit

In my case it wasn't a credit. I just paid for petrol at a petrol station with debit card.

pedrobaeza:
Check #820

PR #693, opened 2024-05-10
[16.0][FIX] account_statement_import_sheet_file: Remove empty lines with spaces

mileo:
Some banks send a file with a space in the empty lines.

image


alexey-pelykh (changes requested):
Thanks for the PR! Please add a test case with sample file and I would consider stripping all whitespace-alike symbols with regex

alexey-pelykh (approved):
Nicely done!

pedrobaeza:
Several things: 

- The commit message should include the module name. 
- Don't add empty lines inside a method. 
- Although previous test uses demo data, it's not a good practice, so please create the template for empty lines inside the test (although you can leave the demo one if you want).
- Should this really be an extra option or directly included for all the templates?

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
```diff
@@ -538,3 +538,51 @@ def test_amount_inverse_sign(self):
...
+                    "header_lines_skip_count": 1,
+                    "quotechar": '"',
+                    "float_thousands_sep": "dot",
+                    "reference_column": "MOVIMIENTO",
+                    "description_column": "DESCRIP",
+                    "amount_credit_column": "DEBE",
+                    "amount_debit_column": "HABER",
+                    "balance_column": "SALDO",
+                    "timestamp_format": "%d/%m/%Y",
+                    "timestamp_column": "FECHAMOVI",
```
rvalyi: not so important: but probably better to put the values of the previous 7 dict items in English.

rvalyi (changes requested):

mileo:
Hi Guys, thank you for the review. I've added some unit tests to ensure that the conversion is working correctly for all cases.

mileo:
> Several things:
> 
> * The commit message should include the module name.
> * Don't add empty lines inside a method.
> * Although previous test uses demo data, it's not a good practice, so please create the template for empty lines inside the test (although you can leave the demo one if you want).
> * Should this really be an extra option or directly included for all the templates?

All done... With these new tests, we can ensure that it will work for all cases, without any extra configuration.

rvalyi (approved):

pedrobaeza:
Please don't use Decimal on tests. Just use regular floats.

mileo:
> Please don't use Decimal on tests. Just use regular floats.

@pedrobaeza the code is base on decimal, the change to remove the Decimal is bigger.... Could you please review the second commit?

pedrobaeza (approved):
/ocabot merge patch

PR #696, opened 2024-05-22
[16.0][ADD] account_statement_import_online_plaid

adasatorres:
This addon allow get your transactions using plaid.com platform.


christian-ramos-tecnativa (approved):
Great solution for bank statements import in USA 👍🏻

lauradiaz22 (approved):
Functional test OK. LGTM!

account_statement_import_online_plaid/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": ["account_statement_import_online"],
+    "data": [
+        "views/online_bank_statement_provider.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
+            "/account_statement_import_online_plaid/static/src/**/*.js",
+            "https://cdn.plaid.com/link/v2/stable/link-initialize.js",
```
etobella: It is prefered to download the library locally in order to avoid external connections

etobella (commented):
Technically seems correct, however, I would make a small change

account_statement_import_online_plaid/__manifest__.py
pedrobaeza: Yes, please put a copy. Systems may be isolated and don't access any external resource. Also for security/privacy policies.

adasatorres: Hi @pedrobaeza @etobella , I understand, but on Plaid documentation indicate to use the CDN link because they haven't any local library. https://plaid.com/docs/link/web/.
I download the file with wget but when add the file route on backend assets odoo stop working.
image

pedrobaeza: Maybe there are extra files that are imported in that main file? @etobella can you give here a hand?

etobella: They ask for cdn because it is easier, but we can download the code from https://cdn.plaid.com/link/v2/stable/link-initialize.js and copy it directly to a JS, I don't see the issue there.

where you using debug=assets? Sometimes it gives better errors on JS

carolinafernandez-tecnativa:
Hi, 

1- I have created new journal with this configuration:

image

When i press Pull online i got this error message:

image

But the record exists:

image

adasatorres:
> Hi,
> 
> 1- I have created new journal with this configuration:
> 
> image
> 
> When i press Pull online i got this error message:
> 
> image
> But the record exists:
> 
> image

Hello @carolinafernandez-tecnativa , sorry for the delay. I have checked the mentioned issue, and it seems to be unrelated to this addon, as the problem persists even after uninstalling this addon in the runboat.

jelenapoblet:
@pedrobaeza @etobella anything else we need here to push this through ?

rrebollo:
@adasatorres @pedrobaeza @etobella I'm starting the migration to v17 from this PR. As this is not merged yet, is there any guidelines to follow?

account_statement_import_online_plaid/models/plaid_interface.py
```diff
@@ -0,0 +1,95 @@
...
+from plaid.model.products import Products
+from plaid.model.transactions_get_request import TransactionsGetRequest
+from plaid.model.transactions_get_request_options import TransactionsGetRequestOptions
+
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class PlaidInterface(models.AbstractModel):
+    _name = "plaid.interface"
```
rrebollo, line 21: Migrating to 17.0: Shouldn't this class has security access model records?

rrebollo (approved):
Code Review LGTM

account_statement_import_online_plaid/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "data": [
+        "views/online_bank_statement_provider.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
+            "/account_statement_import_online_plaid/static/src/**/*.js",
+        ],
+    },
+    "external_dependencies": {
+        "python": ["plaid-python"],
```
rrebollo, line 21: Doing the migration to v17 some tests failed because it looks like some things have changed in latest releases
```suggestion
        "python": ["plaid-python == 23.0.0"],
```

rrebollo (changes requested):
I miss the plaid-python version before. Please accept my suggestion.

account_statement_import_online_plaid/models/plaid_interface.py
adasatorres: Good morning, if I remember correctly, abstract models cannot generate a record by themselves; they are only used to create a common structure for other addons.

etobella (approved):

pedrobaeza:
Checking these days. Please wait a bit.

ayushin:
For international companies using banks in different country `action_sync_with_plaid` would work better with:
```
        country_code = self.journal_id.bank_id.country.code or self.env.user.company_id.country_id.code
```

pedrobaeza (changes requested):
This doesn't serve without Plaid giving the prod API keys, which is not easy and very pricey, so it shouldn't be merged for not creating false expectations.

ayushin:
That's not correct, Plaid prod API keys for transactions are possible on entry tier - my company pays less than a dollar a month for 10 accounts

pedrobaeza:
Well, I want to know how can that possible nowadays. I wasn't able. Maybe you have some instructions to follow?

ayushin:
@pedrobaeza You just choose "Pay As You Go" plan and that's it.

pedrobaeza:
I requested that, but still no access:

imagen

And same things seem to happen for others: https://www.reddit.com/r/fintech/comments/bxhvs3/plaid_pay_as_you_go_pricing/?rdt=52816

jelenapoblet:
Hi, i managed to get approved by following the instructions:
image

ayushin:
from my side, I've tested and everything works. the only issue i had is that the bank is in the different country than my company and i suggested a fix above, other than that it all seem to work

lauradiaz22 (approved):
Good morning,

I have tested this PR repeatedly and it's performance has always been correct. This LGTM👍

account_statement_import_online_plaid/models/online_bank_statement_provider_plaid.py
```diff
@@ -0,0 +1,95 @@
...
+        args = [self.username, self.password, self.plaid_host]
+        client = plaid_interface._client(*args)
+        lang = (
+            self.env["res.lang"]._lang_get(self.env.lang or self.env.user.lang).iso_code
+        )
+        company_name = self.env.user.company_id.name
+        country_code = country_code = (
+            self.journal_id.bank_id.country.code
+            or self.env.user.company_id.country_id.code
+        )
```
pedrobaeza:
```suggestion
        country_code = country_code = (
            self.journal_id.company_id.country_id.code
            or self.env.company.country_id.code
        )
```

adasatorres:
> For international companies using banks in different country `action_sync_with_plaid` would work better with:
> 
> ```
>         country_code = self.journal_id.bank_id.country.code or self.env.user.company_id.country_id.code
> ```
 
 @ayushin Good morning, and thank you for the suggestion, I just added it.
 
 @rrebollo Can you add this suggestion to the following PR? #742

account_statement_import_online_plaid/models/online_bank_statement_provider_plaid.py
adasatorres: Good afternoon @pedrobaeza , why not use the country code configured in the bank specified in the journal? In my opinion, I think it's a good idea since you could have banks in different countries than your company.  Additionally, I have been researching, and it is recommended to use the country or region code where the bank you are connecting to operates. For example, to determine the available financial institutions, configure the environment and ensure regulatory compliance, select the language and currency, define the structure of banking credentials, and so on.

pedrobaeza: You may not have the bank defined. And is `country` a field in standard Odoo or is it something of `l10n_es_partner`? At minimum, you need to put that fallback to company's country.

adasatorres: The country field in the bank_id comes from the base module. I understand the possibility that the journal might not have an associated bank, but in that case, the company's country can be used as a default. 
image

pedrobaeza: OK about the first one, but the fallback is incorrect. `self.env.user.company_id` is not the current company, but the default company of the user. You have to use `self.env.company.country_id.code` instead, and I insist that the first fallback must be the company's journal, as you may have selected several companies at the same time, and being linking a journal not from the main company, so `self.env.company` won't be correct as well.

adasatorres: Yes, you're right. It doesn't make sense to use the country code of either the user or the current company. The company associated with the journal should be used.

adasatorres: Would it be a good idea to configure a system parameter with a default country code?  There will always be a possibility that the country field is False.

pedrobaeza: The country is mandatory in the company, so I don't think so. You shouldn't over-complicate this case. If you want, you can issue a `UserError` if no country code as a way to stop the process.

rrebollo:
@adasatorres can you pin me when the requested changes are done, to forward-port them to v17 PR.

adasatorres:
Good morning @rrebollo ,

The changes have already been uploaded; I just need to add a test for the new method I created, which is called _check_country. Additionally, I need to modify the translation file for the exception string. I’ll let you know as soon as I have those changes ready.

jelenapoblet:
Hi @adasatorres , while testing in a live environment, I noticed that when the user has the English language set, the following error occurs:
image

adasatorres:
Hi @jelenapoblet , when I get a chance, I will review it. There are languages that Plaid does not support, but US English should work . I need to create a method to verify those languages.

ayushin:
@adasatorres I discovered a bug that when I try to pull historical transactions for two months when interval is set to day - i end up reaching API limits. I suppose it should be possible to do historical pull in a single API request

pedrobaeza:
I should be very unlucky, as I have never been approved to have production keys...

adasatorres:
Hello @ayushin , thanks for the comment. I was already aware of that issue. I will try to review it again to see if there is a possibility of optimizing the requests to the API.

jelenapoblet:
> I should be very unlucky, as I have never been approved to have production keys...

@pedrobaeza , have you tried and american or an spanish company ? I'm currently working with this, 200 live API calls per product are free and each statement import i think it uses 2-3 API calls. So far, i'm happy with the result

ayushin:
@adasatorres I still get too many requests limit from plaid if i try to download 2 months of transactions (not even that many)

ayushin:
@adasatorres the issue is when you try to use pull wizard and you have daily statements - it generates a lot of unnecessary API calls and runs out of limit immediately - fails. This is a critical issue, any idea when you can fix that?

edescalona (approved):
LGTM

jelenapoblet:
@pedrobaeza do you know what's holding this back ?

pedrobaeza:
The fact that this is not a valid solution, as it's barely impossible to get the authorization from Plaid, so having something merged that technically works, but that can't be used is not OK.

jelenapoblet:
We have clients using it, and we use it ourselves. @ayushin what's your experience with this ?

pedrobaeza:
Yes, you are the only one, and it got you 1 year to get the authorization (and if I'm not wrong, you are still on "developer mode"). That's because Plaid expects you to implement a middle platform to manage all your customers, and use their API with just a single key, which is not the approach of this set of modules.

jelenapoblet:
To use the statement import feature, you don’t need developer mode. You can simply create a free pay-as-you-go account, and that will give you access to statement import. The feature you’re mentioning belongs to ‘bill pay,’ which is separate.

jelenapoblet:
This is the one you’re referring to: https://github.com/OCA/account-payment/pull/809

pedrobaeza:
Not really, Jorge, I haven't been able to download a single statement with the account I created. Please refer to your colleagues to ask about it.

ayushin:
Just because it does not work for you personally @pedrobaeza is not a good reason to stall it. It's the only open source bank import solution for Odoo in the USA. We use it and it seems more people do judging by comments in the thread. 

It also needs to be ported to 18.0 asap.

That's why I was saying maybe this module should be published outside of the OCA - as it works great for us @jelenapoblet

pedrobaeza:
No, it's not that it's not working for me. It's not working for anyone trying to use it in the way it's intended (putting on each customer the API keys, not a centralized way as it's expected by Plaid). As it's open source, you can host this in your own repository if you want, but I don't see this module as a valid one for OCA. Anyway, this can have the opinion of any other PS of @OCA/banking-maintainers.

crrodrigueztrujillo:
Just to add some production feedback here: we are using this addon with Plaid in several live Odoo 16.0 customer databases in the US, and the online statement import is working well for our use cases.

@jelenapoblet @ayushin @pedrobaeza

jelenapoblet:
Rather than pushing it outside OCA, I believe the right direction is to keep it and continue improving. The concerns (per-customer vs. centralized API keys) are valid, but this is the best solution we currently have, it’s fully open source, OCA-compliant, and it works well. We can continue improving it over time.
Maybe moving it under the l10n_us repository could further strengthen its place within the localization effort.....

pedrobaeza:
As said, for me it's a no, but if other PSCs consider adequated, I won't oppose.

ayushin:
> Rather than pushing it outside OCA, I believe the right direction is to keep it and continue improving. The concerns (per-customer vs. centralized API keys) are valid, but this is the best solution we currently have, it’s fully open source, OCA-compliant, and it works well. We can continue improving it over time.
> 
> Maybe moving it under the l10n_us repository could further strengthen its place within the localization effort.....

Well the way plaid works you need to have your own API key otherwise you would need to have another paid service with all security implications 

So I wouldn't want it any other way

account_statement_import_online_plaid/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,27 @@
...
+#. Switch to *Bank Account* tab
+#. Set *Bank Feeds* to *Online*
+#. Select *Plaid.com* as online bank statements provider in
+   *Online Bank Statements (OCA)* section
+#. Save the bank account
+#. Click on provider and configure provider-specific settings.
+
+To obtain *Login* and *Key*:
+
+#. Open `Plaid.com <https://plaid.com/>`_.
```
etobella: It would be nice to have a better explanation of how to create your account in plaid to make it work with this App.

etobella (commented):
IMO, I think it might have sense to have the app. However, i would like to see a better explanation of how to create my account. For example, I started doing this, but I was not sure which was the right approach.

account_statement_import_online_plaid/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "data": [
+        "views/online_bank_statement_provider.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
+            "/account_statement_import_online_plaid/static/src/**/*.js",
+        ],
+    },
+    "external_dependencies": {
+        "python": ["plaid-python == 28.0.0"],
```
sbidoul: Please do not place exact pins on external dependencies. See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#181manifest for more information about this.

edit: added missing **not** :facepalm:

sbidoul:
I'm fine with merging this with some more explanations in the readme about the keys and caveat about the need to obtain developer keys, as suggested by Enric.

Can you also remove the icon (it's the old one and the latest will be set by the bot after merge).

I also recommend to declare maintainers in the manifest.

account_statement_import_online_plaid/__manifest__.py
rrebollo: > Please do place exact pins on external dependencies. See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#181manifest for more information about this.

reading the resource you pointed out, I think you mean "don't place", right?

sbidoul: Oh, yes :facepalm:

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution!

jelenapoblet:
Thanks @antoniodavid and @rrebollo for your contribution, great job!! @etobella @sbidoul , the comments have been addressed and ready to merge, and we’re very happy about this important contribution for the US community!

etobella (approved):
Thanks for the clarification.

pedrobaeza (approved):
With the text added about the approval process (and it seems US based companies are lucky in this), let's merge it having been warned if a user wants to use it.

/ocabot merge nobump

ayushin:
Hallelujah!

PR #698, opened 2024-05-23
16.0 mig account statement import txt xlsx dummy draft

rov-adhoc:


pedrobaeza:
Why are you proposing again this? As said in the previous PR, it has been migrated and renamed to the more generic one.

PR #699, opened 2024-05-24
16.0 account statement import txt xlsx dummy only for adhoc

rov-adhoc:
Please do not close


pedrobaeza:
Why it shouldn't be closed? Please expose the reasons. If it's just for Adhoc (but I don't understand the reasons, as the merged module is exactly this one), you should do it on your fork, not here. You can activate the CI in your fork.

jjscarafia:
Hi @pedrobaeza 
The reason of having this draft PR is that we're using gitaggregate with referrals to PR's numbers when building our images.
We've a pending task so that we can improve this and avoid the need of the PR, but we don't have that implemented yet.

If the PR is closed and we push changes to the branch, somewhow those changes are note being considered.

IMHO, having a draft PR shouldn't bother anyone. I'm wrong?

Thanks for your active work :)

pedrobaeza:
I insist: this module is already migrated in 16.0. Why do you need another different migration?

jjscarafia:
Hi @pedrobaeza 
Our situation is the following:
1. we've started a PR migrating this module
2. before being merge we start using this module on 16 customers
3. on the meantime, a new module with a new name was merge
4. we decide to keep using our pr instead of moving to new module to avoid issues with customers (and decide to move to new module on 17)

As said, it is more an internal need and we used to think that draft prs were not harmful nor checked by anyone. So now we will work on changing our approach and avoid these kind of prs and close this one

Thanks!

pedrobaeza:
Thanks for answering. Now I understand a bit more your problem. You should consider adding in your deployment toolset the possibility to use your branches without the need of a pull request. https://github.com/acsone/git-aggregator for example allows that, so you don't depend on these things. Take into account that the bot will also close this PR in the future.

But I think you should switch already to the new module. That can be done on existing DBs without too much pain:

- Have `openupgradelib` library installed.
- Open a Odoo shell.
- Execute this code:

```python
from openupgradelib import openupgrade
openupgrade.update_module_names(env.cr, [("account_statement_import_txt_xlsx", "account_statement_import_sheet_file"], merge_modules=True)
env.cr.commit()
```
- Replace one module by the other.
- And your Odoo instance will switch from one module to another without any loss.

jjscarafia:
Great @pedrobaeza !
Thanks for the tips. We'll close the PR

cc @rov-adhoc

PR #700, opened 2024-05-28
[14.0][FIX] account_statement_import: Do not overwrite journal active_id

grindtildeath:
The call to parse_file must not overwrite the active_id value in the context as this active_id refers to the journal from which the wizard was called.

It was especially wrong because the existing active_model key is still set to account.journal, with an active_id that was referencing account.statement.import wizard ID.

If the wizard ID is needed, it can still be found on self, and if this ID must be propagated through the context, it must use a dedicated key instead of hijacking the active_id.


SilvioC2C (approved):
LGTM

santostelmo (approved):

cyrilmanuel (approved):

PR #703, opened 2024-06-19
[14.0][FIX] account_statement_import_txt_xlsx:  empty balance_column field

Tiago370:
This change addresses the error obtained when the column cell `amount_column` is empty.

### DIFF
--- a/account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
@@ -4,6 +4,7 @@
 
 import itertools
 import logging
+import re
 from datetime import datetime
 from decimal import Decimal
 from io import StringIO
@@ -61,7 +62,6 @@ class AccountStatementImportSheetParser(models.TransientModel):
         lines = self._parse_lines(mapping, data_file, currency_code)
         if not lines:
             return currency_code, account_number, [{"transactions": []}]
-
         lines = list(sorted(lines, key=lambda line: line["timestamp"]))
         first_line = lines[0]
         last_line = lines[-1]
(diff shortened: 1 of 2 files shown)


mileo:
Backport of https://github.com/OCA/bank-statement-import/pull/693

account_statement_import_txt_xlsx/models/account_statement_import_sheet_parser.py
```diff
@@ -61,7 +62,9 @@ def parse(self, data_file, mapping, filename):
         lines = self._parse_lines(mapping, data_file, currency_code)
         if not lines:
             return currency_code, account_number, [{"transactions": []}]
-
+        if not isinstance(lines[0]["timestamp"], datetime):
```
alexey-pelykh: How this is related to empty balance column?

Tiago370: Hello @alexey-pelykh , this line was necessary, because some banks' statements put the first line without a date, containing only the description and the previous balance, so the solution I found was to copy the date of the next line. But I'm open to suggestions, thanks for the review.

alexey-pelykh: This should be done as a separate PR. Backport is a backport, sneaky changes aren't good :)

Tiago370: I understand. I will provide a new PR and fix this one.

mileo (approved):

mileo (approved):

pedrobaeza:
This is replacing the attribution of @mileo . Why aren't you doing a cherry-pick?

Tiago370:
Good morning @pedrobaeza , thanks for the review. I did this because the module name is different in version 16.0. Is there any way I can do a cherry-pick in this situation? I am available for suggestions.

pedrobaeza:
Well, you can cherry-pick and when giving the error, then manually insert the changes and remove the offending files, and do `git add -A; git cherry-pick --continue`. Another option is to simply do `git commit --all --amend --author ...` with the name and mail of Luis for preserving at least the attribution.

Tiago370:
@pedrobaeza Perfect, I'll try that.

PR #396, opened 2021-06-16
[14.0][MIG] account_statement_import_online_qonto

alexis-via:
Ready for review/merge

PR #718, opened 2024-09-02
[16][FIX] account_statement_import_online_qonto : fix import

florian-dacosta:
I think there are some big issues on the v16 module.
Mainly, the `payment_ref` of the statement line is empty (although it is the main field that give us information about the transaction) and the number (field `name`) of the accounting entry is not the one of the journal, it is the label of the transaction!
Probably a confusion due to the fact that `account.bank.statement.line` inherits `account.move`

This PR also fix the foreign currency management which is not functional at the moment.

Also remove the verify=False on the calls to qonto.
There are no comment whatsoever indicating why it was done, and this is working fine without it.

The changes comes from https://github.com/OCA/bank-statement-import/pull/396/commits/21712f53406e3f1e0381c064c9f9e118b8615e42 

I don't know if the module is used already in v16, I guess/hope not, but it would be nice to review/merge this fast as the issues are importants.
@victoralmau @pedrobaeza @carolinafernandez-tecnativa

### DIFF
--- a/account_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
+++ b/account_statement_import_online_qonto/models/online_bank_statement_provider_qonto.py
@@ -54,9 +54,7 @@ class OnlineBankStatementProviderQonto(models.Model):
     def _qonto_get_slug(self):
         self.ensure_one()
         url = QONTO_ENDPOINT + "/organizations/%7Bid%7D"
-        response = requests.get(
-            url, verify=False, headers=self._qonto_header(), timeout=10
-        )
+        response = requests.get(url, headers=self._qonto_header(), timeout=10)
         if response.status_code == 200:
             data = json.loads(response.text)
             res = {}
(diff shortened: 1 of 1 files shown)


flotho (approved):
Hi,
I confirm that this PR solve some problems in production.
For example, without this PR when importing, Odoo raise an issue about duplicate move name, mainly because of the absence of name in transactions.
This PR works fine.
Code review OK,

flotho (commented):
hi @florian-dacosta ,

could you change the PR nme and the commits to follow guidelines ?

Regarsd

florian-dacosta:
> hi @florian-dacosta ,
> 
> could you change the PR nme and the commits to follow guidelines ?
> 
> Regarsd

It should be ok now @flotho

flotho (approved):
Thanks @florian-dacosta 
LGTM

pedrobaeza (approved):
/ocabot merge patch

PR #719, opened 2024-09-05
[14.0] [ADD] account_statement_import_txt_xlsx: offset header

matthieusaison:


matthieusaison:
Hello @alexey-pelykh, is it possible to take a look at my PR?

alexey-pelykh (changes requested):
While the feature is straightforward, having a test covering that would be great

PR #729, opened 2024-10-18
[16.0][FIX] account_statement_import_base: sanitized acc number

remytms:
The field acc_number in the database is not sanitized by default.

This leads to account_bank_statement_line with an empty partner_id because the check done in _statement_line_import_update_hook() uses the sanitized_acc_number instead of the acc_number.

For a right comparison the sanitized version of the acc_number should be put in the speeddict.

### DIFF
--- a/account_statement_import_base/models/account_journal.py
+++ b/account_statement_import_base/models/account_journal.py
@@ -24,7 +24,10 @@ class AccountJournal(models.Model):
             ["acc_number", "partner_id"],
         )
         for partner_bank in partner_banks:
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
OK about the content. Please avoid the abbreviation in the module name for the commit message.

remytms:
Thanks @pedrobaeza Commit has been fixed.

pedrobaeza (approved):

victor-champonnois (approved):
Functional review OK

houssine78 (approved):

Issue #726, opened 2024-10-08
account_statement_import_online_paypal Paypal Statement Import Failure. Error message: Failed to resolve transaction


maxgrassi:
Describe the bug:
Since 02/10/2024, the PayPal movements are no longer being imported into Odoo through the account_statement_import_online_paypal module. When attempting to manually pull transactions, the following error message appears:
Failed to resolve transaction "transaction number (date and time)". Please see enclosed screenshot.

This issue is not related to duplicated transactions, but seems to be related to communication permissions between Odoo and PayPal. No changes were made to the system prior to the issue, and we are using the latest version of the module.

To Reproduce
Affected versions: Odoo 16 Community Edition

Go to Accounting > Configuration > Journals > Paypal > Pull online bank statement 
Set from / to date
Attempt to import PayPal transactions manually
The error appears, and no transactions are imported
Expected behavior
The PayPal transactions should be imported into Odoo without any error.

Additional context
Odoo 16 Community Edition
Module version: 16.0.1.0.1
Issue started on 02/10/2024
It seems to be related to communication or permission issues between Odoo and PayPal.

![paypal-statement-import-failure](https://github.com/user-attachments/assets/34dd3290-4783-4c3e-8843-b738c1012cae)

Roodin:
It is related to method  **def _paypal_get_transaction. ** Now it seems not to be able to retrive the transaction with the same **start_date** and **end_date** in the url  (transaction_date). If you change that parmeters in 1 second it will work.

maxgrassi:
> It is related to method **def _paypal_get_transaction. ** Now it seems not to be able to retrive the transaction with the same **start_date** and **end_date** in the url (transaction_date). If you change that parmeters in 1 second it will work.

Hi Roodin,

Thank you for your response. I understand that the issue might be related to the method def _paypal_get_transaction, specifically with how it retrieves transactions when the start_date and end_date are the same in the URL (transaction_date).

However, I have already tried adjusting the start and end dates (including changing them by 1 second) in the manual retrieval dialog box, but the issue persists. Even with different date ranges or times, I am unable to retrieve PayPal transactions either automatically or manually.

Previously, the transactions were imported automatically without needing to manually pull them. Now, even when attempting the manual pull with adjusted parameters, it still fails to retrieve the transactions.

Could you clarify if there is another place in the code or configuration where I should modify the date parameters (perhaps directly in the method _paypal_get_transaction)? If so, could you provide guidance on how and where to make these changes?

[...]

PR #730, opened 2024-10-21
[16][FIX] account_statement_import_online_paypal : search transaction with a 1 second margin to ensure it is found

florian-dacosta:
Workaround for https://github.com/OCA/bank-statement-import/issues/726 as proposed by @Roodin 
I propose it here so it is usable for people having this issue, but it may not be the best way.
I do not really know the module so I do not try something else but I do not understand why this is important : https://github.com/OCA/bank-statement-import/blob/16.0/account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py#L235

Don't we have already the transaction in `first_transaction` variable?, why do we need to fetch it once again ?
Maybe @alexey-pelykh could enlighten me ?


chandni299 (approved):
Looks good :+1:

ntsirintanis (approved):

houssine78 (approved):

PR #732, opened 2024-10-25
[17.0][MIG] account_statement_import_move_line: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa TT51242

@pedrobaeza @victoralmau please review


account_statement_import_move_line/tests/test_account_bank_statement_import_move_line.py
```diff
@@ -2,59 +2,48 @@
...
-        super(TestAccountBankStatementImportMoveLine, cls).setUpClass()
-
-        cls.account_type = cls.env["account.account.type"].create(
-            {"name": "Test Account Type", "type": "other", "internal_group": "asset"}
-        )
-
-        cls.a_receivable = cls.env["account.account"].create(
+        super().setUpClass()
+        cls.env.user.groups_id = [(4, cls.env.ref("account.group_account_user").id)]
+        cls.a_income = cls.env["account.account"].create(
```
pedrobaeza: Don't create accounts and just convert the test to post_install. Then, you should create the invoice without a form, as it should work without issues. Check for example this method that creates invoices without using Form: https://github.com/OCA/commission/blob/d9e3e8e944e2fd2ec8eba6e0efea4ddb01e9eb5b/account_commission/models/commission_settlement.py#L129

account_statement_import_move_line/models/account_move_line.py
```diff
@@ -32,6 +32,13 @@ def _prepare_statement_line_vals(self, statement):
         return vals
 
     def create_statement_line_from_move_line(self, statement):
+        # Check if the statement has a journal assigned to it and set a
+        # default journal. This can be modified in the form lines
+        if not statement.journal_id:
+            bank_journal = self.env["account.journal"].search(
+                [("type", "=", "bank")], limit=1
+            )
+            statement.journal_id = bank_journal
```
victoralmau: I think this is not necessary because it will be defined automatically when a line is created; if you should specify in the `_prepare_statement_line_vals()` method the `journal_id` that corresponds.

victoralmau (changes requested):
Code and functional review OK

account_statement_import_move_line/models/account_move_line.py
pilarvargas-tecnativa: An statement assigns journals of type "bank" or "cash" to its lines when they are created. In version 15, statement lines did not have an associated journal and the journal of the statement was set. The wizard imports invoice lines that may have a journal other than ‘bank’ or ‘cash’ and when creating them, it must be done with the journal associated with the extract if it already contains previous lines that have established it or, as is done when creating them from the form, establish by default a journal of type ‘bank’ in this case, as they are created from that journal.

pedrobaeza: Being a computed field, you need to infer the journal other way. Include it in the wizard, and fill it depending on the context (I think it's included in the context when coming from the dashboard).

pilarvargas-tecnativa: I have decided to keep the current logic for cases where a statement with added lines already exists, in case more lines need to be added, limiting the visibility of the current button to the condition "not journal_id."

Additionally, I have added the option to import lines from the dashboard by creating a new statement.

If at some point you permanently lose the form view of the statement, you can continue with the new dashboard option.

account_statement_import_move_line/models/account_move_line.py
```diff
@@ -32,6 +32,13 @@ def _prepare_statement_line_vals(self, statement):
         return vals
 
     def create_statement_line_from_move_line(self, statement):
+        # Check if the statement has a journal assigned to it and set a
+        # default journal. This can be modified in the form lines
+        if not statement.journal_id:
+            bank_journal = self.env["account.journal"].search(
```
pedrobaeza: This search shouldn't be done, as randomly assign one journal is not OK.

pedrobaeza:
Maybe it's better to reorder the commits to put the fix before migration commits.

account_statement_import_move_line/models/account_move_line.py
pilarvargas-tecnativa: It was dead code that I had not removed. It is now as it should be

account_statement_import_move_line/readme/USAGE.md
```diff
@@ -1,8 +1,6 @@
...
-2.  Create a new bank statement from a bank journal.
-3.  Click button "Import Journal Items".
-4.  Select filtering options.
-5.  Click on button "Add All Move Lines" to auto-select the move lines
-    matching the selected criteria or click on Add an item to manually
-    select the move lines filtered by the criteria.
-6.  Click on button "Create Statement Lines".
+Journal entries can be imported in two ways, from a new bank statement or from an existing bank statement.
+1. Go to Invoicing > Dashboard.
+2.  To import to a new statement, in the relevant bank journal click on ‘Import journal items’. To import from an existing bank statement, open the statement form and click on ‘Import journal items’.
```
pedrobaeza: Why the extra spaces?

pedrobaeza (approved):

account_statement_import_move_line/readme/USAGE.md
pilarvargas-tecnativa: Corrected

victoralmau (approved):

PR #733, opened 2024-10-29
[15.0][FIX] account_statement_import_move_line: Set reference in bank statment lines

pilarvargas-tecnativa:
The ‘ref’ field is not visible on invoices of type ‘out_invoice’. In the tests, it is adding value to this field and therefore does not fail to create a bank statement line where the ‘payment_ref’ field is required. To solve this, the value of the ‘payment_ref’ field of the bank statement lines is set as the reference of a supplier invoice (‘ref’ in invoice lines) or as the reference of the payment ‘payment_reference’ (‘name’ in invoice lines).

cc @Tecnativa


pedrobaeza (approved):
Yes, we can avoid problems this way. Let's fast-track it:

/ocabot merge patch

PR #735, opened 2024-11-07
[18.0][MIG] account_statement_import_base: Migration to 18.0

xaviedoanhduy:

PR #736, opened 2024-11-07
[18.0][MIG] account_statement_import_file: Migration to 18.0

xaviedoanhduy:
### depends on:

- https://github.com/OCA/account-reconcile/pull/739
- https://github.com/OCA/bank-statement-import/pull/735

### change in v18

- menu `menu_finance_entries_actions` (Accounting/Actions) was removed in commit: https://github.com/odoo/odoo/commit/70a2618


account_statement_import_file/tests/test_account_statement_import_file.py
```diff
@@ -38,7 +36,9 @@ def setUpClass(cls):
                 "bank_account_id": cls.bank_account.id,
             }
         )
-        f_path = file_path("account_statement_import_file/tests/test_file.txt")
+        f_path = file_path(
+            "account_statement_import_file/test_files/test_statement_import.txt"
+        )
```
mmequignon: Why ? This file is only used in unittests no? Why extracting this in another folder?

xaviedoanhduy: hi @mmequignon,
you're right, it should be in the `/tests/samples` folder because there is a convention for that type of file in https://github.com/OCA/bank-statement-import/blob/a919f9bab7187b8f89d66074c10535a61a0f4977/.pre-commit-config.yaml#L20
so on this occasion, i have moved it to the `/tests/samples` folder

henrybackman (approved):

rov-adhoc:
@pedrobaeza Hey! I would like to ask you if its something missing in order to be approve. Thanks in advance!

grindtildeath (approved):

rov-adhoc (approved):

PR #691, opened 2024-05-07
[15.0][IMP]account_statement_import_txt_xlsx: additional mapping controls

PicchiSeba:
fw-port of #688

PR #740, opened 2024-11-15
[16.0] [FW-PORT] account_statement_import_sheet_file: additional mapping controls

renda-dev:
FW-PORT of https://github.com/OCA/bank-statement-import/pull/691 to v16.0


account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -179,7 +179,7 @@ def _parse_lines(self, mapping, data_file, currency_code):
             csv_or_xlsx = reader(StringIO(decoded_file), **csv_options)
         header = False
         if not mapping.no_header:
```
HekkiMelody: **question**: shouldn't the `if mapping.offset_column` (line 191) be intended under `if not mapping.no_header` (line 181)?

HekkiMelody (approved):
Code review, LGTM.

My question relates to a change that was already merged in previous versions, so not blocking.

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
renda-dev: I think it's the right time to fix it! 

For the future:
We moved the `offset_column` logic under `if not mapping.no_header:` because otherwise it would try to index slice on a `False` value, which would raise an error.

AinohaBH (approved):
Functional review and this LGTM

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -246,136 +375,11 @@ def _parse_rows(self, mapping, currency_code, data, columns):  # noqa: C901
                 if index >= footer_line:
                     continue
                 values = list(row)
-
-            timestamp = self._get_values_from_column(
```
pedrobaeza, line 250: Why all these remove code?

pedrobaeza (changes requested):

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
alexey-pelykh: It all went into `def _parse_row`

pedrobaeza:
Well, I don't see too much value added there, and it creates a lot of noise, specially not being the focus of this PR.

HekkiMelody:
> Well, I don't see too much value added there, and it creates a lot of noise, specially not being the focus of this PR.

@pedrobaeza Thank you for reviewing this. We have removed the refactor in order to make the diff smaller and reduce noise.

pedrobaeza:
Thanks. The CI is not red.

renda-dev:
Green light @pedrobaeza :)

Thanks for checking the PR

pedrobaeza (approved):
/ocabot merge minor

pedrobaeza:
Please fw-port it to 17

PR #743, opened 2024-12-02
[17.0] [FW-PORT] account_statement_import_sheet_file: additional mapping controls

renda-dev:
FW-Port of https://github.com/OCA/bank-statement-import/pull/740


pedrobaeza:
Please check CI

renda-dev:
> Please check CI

Everything ready! @pedrobaeza

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -174,6 +174,14 @@ class AccountStatementImportSheetMapping(models.Model):
         help="Set the Header lines number.",
         default="0",
     )
+    skip_empty_lines = fields.Boolean(
+        default=False,
+        help="Allows to skip empty lines",
+    )
+    offset_column = fields.Integer(
+        default=0,
+        help="Horizontal spaces to ignore before starting to parse",
```
pedrobaeza: Horizontal? Shouldn't it be vertical?

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -174,6 +174,14 @@ class AccountStatementImportSheetMapping(models.Model):
         help="Set the Header lines number.",
         default="0",
     )
+    skip_empty_lines = fields.Boolean(
+        default=False,
```
pedrobaeza: Maybe `True` is a better default.

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
renda-dev: At this point, to be more direct, I changed it to "Columns to ignore before starting to parse"

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
renda-dev: It was meant to keep the old behavior as the standard, but I really don't have any preference on that, so I changed it

pedrobaeza: Yeah, I supposed that, and it's good that you think on that "retro-compatibility", but now on this version that is not so old and established, I was thinking on being more "adventurous" to set a sane default.

renda-dev: I like it! Everything should be ready and good to go :)

HekkiMelody (approved):
Code review, LGTM

rov-adhoc (approved):

pedrobaeza (approved):
/ocabot merge minor

feg-adhoc:
This PR is causing errors in the tests because, when running the tests test_offsets and test_skip_empty_lines, in addition to evaluating a UserError, a ValueError occurs when the mapping continues to execute. Furthermore, these tests are receiving the context account_statement_import_txt_xlsx_test = True. What is its purpose? @renda-dev

renda-dev:
> This PR is causing errors in the tests because, when running the tests test_offsets and test_skip_empty_lines, in addition to evaluating a UserError, a ValueError occurs when the mapping continues to execute. Furthermore, these tests are receiving the context account_statement_import_txt_xlsx_test = True. What is its purpose? @renda-dev

I see that the context [was used in 14.0](https://github.com/OCA/bank-statement-import/blob/14.0/account_statement_import_txt_xlsx/models/account_statement_import.py#L35-L37) to raise an error and stop the execution, but I can't seem to find it anywhere in v17...

You could try to bring back that behavior and see if it fixes @feg-adhoc

feg-adhoc:
Hello! @renda-dev  We made some changes so you can run the tests in version 18. [Here's the link for reference](https://github.com/OCA/bank-statement-import/pull/747/files#diff-9f9a46949b082c3b937202d10d20e903dc52f4b8bfd8959e44060ff17ec9d584R38). We made this modification because we noticed that the "Raise" was not stopping the function, which is why it ended up causing the ValueError. This error also occurred in version 17, so it needs to be reviewed for future changes.

PR #748, opened 2024-12-14
[16.0][FIX] account_statement_import_online_gocardless : token/refresh endpoint

arpol-dev:
Correction of the endpoint token/refresh request datas that were not correctly set up, this request expects the refresh token but not the secret_id and secret_key.

Plus the behavior of the _gocardless_get_token method was calling the endpoint "token/new" instead of "token/refresh" in case of a non expired refresh token.

### DIFF
--- a/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
+++ b/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
@@ -84,17 +84,17 @@ class OnlineBankStatementProvider(models.Model):
             # Refresh token
             if (
                 self.gocardless_refresh_token
-                and now > self.gocardless_refresh_expiration
+                and now < self.gocardless_refresh_expiration
             ):
                 endpoint = "token/refresh"
+                request_data = {"refresh": self.gocardless_refresh_token}
             else:
                 endpoint = "token/new"
+                request_data = {"secret_id": self.username, "secret_key": self.password}
             _response, data = self._gocardless_request(
                 endpoint,
                 request_type="post",
-                data=json.dumps(
-                    {"secret_id": self.username, "secret_key": self.password}
-                ),
+                data=json.dumps(request_data),
                 basic_auth=True,
             )
             expiration_date = now + relativedelta(seconds=data.get("access_expires", 0))
--- a/account_statement_import_online_gocardless/README.rst
+++ b/account_statement_import_online_gocardless/README.rst
@@ -129,6 +129,9 @@ Contributors
 * `Alusage <https://nicolas.alusage.fr>`__:
 
   * Nicolas JEUDY <https://github.com/njeudy>
+* `ArPol <https://arpol.fr>`__:
+
+  * Armand POLMARD <https://github.com/arpol-dev>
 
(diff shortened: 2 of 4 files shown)


arpol-dev:
@pedrobaeza @njeudy
Could one of you review this PR ? 
I don't know what to do with the codecov fails :/
Happy new year and thank you !

njeudy (approved):
Ok for me !

account_statement_import_online_gocardless/models/online_bank_statement_provider.py
```diff
@@ -84,17 +84,17 @@ def _gocardless_get_token(self):
             # Refresh token
             if (
                 self.gocardless_refresh_token
-                and now > self.gocardless_refresh_expiration
+                and now < self.gocardless_refresh_expiration
```
pedrobaeza, line 87: This change is not correct.

account_statement_import_online_gocardless/models/online_bank_statement_provider.py
```diff
@@ -84,17 +84,17 @@ def _gocardless_get_token(self):
...
             ):
                 endpoint = "token/refresh"
+                request_data = {"refresh": self.gocardless_refresh_token}
```
pedrobaeza, line 90: Why is this needed?

account_statement_import_online_gocardless/models/online_bank_statement_provider.py
njeudy: @pedrobaeza refresh_token can be used if it is still valid, else, you should use "create new token".

account_statement_import_online_gocardless/models/online_bank_statement_provider.py
njeudy: if you use "refresh" and not 'create new' you can update openbanking authorization vith the valid refresh token, else, create new.

arpol-dev: Hello @pedrobaeza 
I had the issue a second time recently. If the bank connection expires and is not renewed before the refresh token expires too, the current code is not working and the connection cannot be renewed.
I believe you never tested that case where the refresh token is expired because the endpoint "token/refresh" requires a different dataset than the endpoint "token/new"

account_statement_import_online_gocardless/models/online_bank_statement_provider.py
arpol-dev: @pedrobaeza 
The endpoint "token/refresh" is meant to refresh the bank connection token using the refresh token. But currently, this endpoint is only used when that refresh token is expired.
We should use : 
- the endpoint "token/refresh" when the refresh token is valid
- the endpoint "token/new" when the refresh token is expired or not defined

pedrobaeza (approved):
Sorry for the late review, and yes, you're right. I was thinking in the token invalidity, not when it's invalid, what to do.

Let's finally integrate this.

/ocabot merge patch

pedrobaeza:
Forward-port in #808

PR #747, opened 2024-12-13
[18.0][MIG] account_statement_import_sheet_file: Migration to 18.0

feg-adhoc:

PR #753, opened 2024-12-27
[FIX] account_statement_import_sheet_file: Fixed new tests errors

feg-adhoc:


rov-adhoc (approved):

houssine78 (approved):

account_statement_import_sheet_file/models/account_statement_import.py
```diff
@@ -35,7 +35,7 @@ def _parse_file(self, data_file):
                 )
             except BaseException as exc:
                 if self.env.context.get("account_statement_import_sheet_file_test"):
-                    raise
+                    return
```
pedrobaeza, line 38: Why do you need to change this?

pedrobaeza: I see I did the same question in https://github.com/OCA/bank-statement-import/pull/747#discussion_r2034584275, but the answer is not very explaining. Maybe you can show the problem in a CI log?

PR #754, opened 2024-12-30
[16.0][IMP] account_statement_import_online: add option to enable or disable statement creation.

carlos-lopez-tecnativa:
Starting from Odoo 16, bank statements are optional. This commit introduces the option to create statements automatically or skip their creation.

TT52481
@Tecnativa @pedrobaeza Could you please review this

### DIFF
--- a/account_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_statement_import_online/models/online_bank_statement_provider.py
@@ -26,6 +26,10 @@ class OnlineBankStatementProvider(models.Model):
 
     company_id = fields.Many2one(related="journal_id.company_id", store=True)
     active = fields.Boolean(default=True)
+    create_statement = fields.Boolean(
+        default=True,
+        help="Create statements for the downloaded transactions automatically or not.",
+    )
     name = fields.Char(compute="_compute_name", store=True)
     journal_id = fields.Many2one(
         comodel_name="account.journal",
@@ -76,11 +80,12 @@ class OnlineBankStatementProvider(models.Model):
     )
     statement_creation_mode = fields.Selection(
         selection=[
-            ("daily", "Daily statements"),
-            ("weekly", "Weekly statements"),
-            ("monthly", "Monthly statements"),
+            ("daily", "Day"),
+            ("weekly", "Week"),
+            ("monthly", "Month"),
         ],
         default="daily",
+        string="Transactions interval to obtain",
         required=True,
     )
     api_base = fields.Char()
@@ -298,6 +303,8 @@ class OnlineBankStatementProvider(models.Model):
         """Final creation of statement if new, else write."""
         AccountBankStatement = self.env["account.bank.statement"]
         is_scheduled = self.env.context.get("scheduled")
+        if not self.create_statement:
+            return self._online_create_statement_lines(statement_values)
         if is_scheduled:
             AccountBankStatement = AccountBankStatement.with_context(
                 tracking_disable=True,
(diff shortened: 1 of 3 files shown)


pedrobaeza (commented):
Thanks for the changes. For avoiding too much diff being the same, what you can do is:

```python
if not self.create_statement:
    return self._online_create_statement_lines(statement_values)
```

extracting the specific code of lines there, and leave the rest untouched.

carlos-lopez-tecnativa:
> Thanks for the changes. For avoiding too much diff being the same, what you can do is:
> 
> ```python
> if not self.create_statement:
>     return self._online_create_statement_lines(statement_values)
> ```
> 
> extracting the specific code of lines there, and leave the rest untouched.

Thanks for your suggestion! I have applied it as you recommended. Could you please update your review?

pedrobaeza:
I think you should return the empty statement in the new method.

carlos-lopez-tecnativa:
> I think you should return the empty statement in the new method.

Done

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -81,6 +85,7 @@ class OnlineBankStatementProvider(models.Model):
             ("monthly", "Monthly statements"),
         ],
         default="daily",
+        string="Transaction Syncronization",
```
pedrobaeza:
```suggestion
        string="Transactions synchronization",
```

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -319,6 +326,12 @@ def _statement_create_or_write(self, statement_values):
             statement.write(statement_values)
         return statement
 
+    def _online_create_statement_lines(self, statement_values):
+        AccountBankStatementLine = self.env["account.bank.statement.line"]
```
pedrobaeza, line 330:
```suggestion
        AccountBankStatementLine = self.env["account.bank.statement.line"]
        if is_scheduled:
            AccountBankStatementLine = AccountBankStatementLine.with_context(
                tracking_disable=True,
            )
```

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -26,6 +26,10 @@ class OnlineBankStatementProvider(models.Model):
 
     company_id = fields.Many2one(related="journal_id.company_id", store=True)
     active = fields.Boolean(default=True)
+    create_statement = fields.Boolean(
+        default=True,
+        help="Create statements automatically or not. Since V16, statements are optional.",
```
pedrobaeza:
```suggestion
        help="Create statements for the downloaded transactions automatically or not.",
```

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -81,6 +85,7 @@ class OnlineBankStatementProvider(models.Model):
             ("monthly", "Monthly statements"),
         ],
         default="daily",
+        string="Transactions synchronization",
```
pedrobaeza: Sorry again to change the text, but thinking twice, this is not the frequency of downloading, so this label will confuse users. It's more the packing option when downloading. If you put monthly, on next synchronization, it will download the whole current month. Which label do you think it should put?

carlos-lopez-tecnativa: "Statement Grouping" sounds like a good label. What do you think? ;)

pedrobaeza: Uhm, not exactly, as there's no statement with the new option, and it's more a question of "what to download". Also the labels of each option are not correct (but we can keep the values for not doing migration scripts). Thinking in something...

pedrobaeza: It's a bit long, but "Transactions interval to obtain" is defining exactly what is doing, and the labels should be changed to "Day", "Week", "Month" instead of the adverbs.

carlos-lopez-tecnativa: @pedrobaeza I’ve updated the code. Is it Ok for you?

pedrobaeza (approved):
Thanks for the changes!

chienandalu (approved):
:+1:

PR #757, opened 2025-01-02
[FIX] account_statement_import_sheet_file: Reset fields based on the selected 'amount_type' value when its change

rov-adhoc:
When 'amount_type' is changed, related fields (such as amount_debit_column,amount_credit_column,debit_credit_column,amount_column) are not cleared, leading to potential inconsistencies in the values.

This commit fixes the issue by ensuring that when 'amount_type' is set to a new value, all the related fields are reset accordingly to avoid any residual values.

Steps to reproduce:
1. Start with the amount_type field set to "Simple Value", and ensure the "Amount Column" field is correctly configured, for example, with the value "Amount".
2. Change the amount_type field to "Distinct Credit/Debit Column", and set values for the "Debit Amount Column" and "Credit Amount Column"  fields, for example, "Debit" and "Credit", respectively.
3. Observe that the previously configured "Amount Column" field is not cleared, leading to potential inconsistencies in the data.

With this fix, related fields such as 'amount_column','debit_credit_column', 'amount_debit_column', and 'amount_credit_column' are correctly reset when 'amount_type' changes.

Video with the issue 
[Video](https://drive.google.com/file/d/1WJRmkW5Sdd6fqWio7VEpWEfjxox-kP6q/view)

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
@@ -254,3 +254,17 @@ class AccountStatementImportSheetMapping(models.Model):
 
     def _get_column_delimiter_character(self):
         return self._decode_column_delimiter_character(self.delimiter)
+
+    @api.onchange("amount_type")
+    def _onchange_amount_type(self):
+        if self.amount_type == "simple_value":
+            self.debit_credit_column = False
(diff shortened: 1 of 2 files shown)


cav-adhoc (approved):

gonzalolenzi:
LGTM

PR #758, opened 2025-01-03
[16.0][IMP] account_statement_import_online_gocardless: use a hashing strategy to produce the unique_transaction_id

JordiBForgeFlow:
When using account_statement_import_online_gocardless with Wise, on a given payment Wise creates two transactions. One for the payment and one for the fee. 

Each payment using Wise generates 2 transactions:

1) Payment transaction: contains a unique transactionID
2) Fee transaction: contains a unique transactionID, and adds to the field entryReference the transactionID associated with the payment.

So we cannot rely on the field 'entryReference' to construct the 'unique_import_id'.

In general one realizes that it can be very hard to decide which field will contain the unique transaction identifier. So, why not use all possible fields and produce a hash of them all?

This is what this PR does.

### DIFF
--- a/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
+++ b/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
@@ -1,6 +1,7 @@
 # Copyright 2022 ForgeFlow S.L.
 # Copyright 2023-2024 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import hashlib
 import json
 from datetime import datetime
 from uuid import uuid4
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
@pedrobaeza @etobella What do you think about this?

etobella:
I understand the approach, but wouldn't be this an issue on the migration?

pedrobaeza:
Yes, you can't replace the current strategy. Possible options:

- You put the hook, and do the change on one of your custom modules.
- You put the hook, and have a configurable system to choose the hashing strategy, having at least the old and new one, and you leave the old one as default, or though migration scripts, you set the old one for all coming from previous revision.

JordiBForgeFlow:
Closing in favour of https://github.com/OCA/bank-statement-import/pull/766

PR #760, opened 2025-01-13
[16.0][FIX] account_statement_import_sheet_file: Prevent negative index in parser header

carlos-lopez-tecnativa:
When importing an Excel file with a header and the field `header_lines_skip_count` is set to its default value (0), the header is retrieved using the -1 index (the last row).
complementary to commit https://github.com/OCA/bank-statement-import/commit/13285abe74d326153c8d0e74c21ddbdf3cdf17a0

UI Before
image
UI After
image

TT52618
@Tecnativa @pedrobaeza @victoralmau @zaoral could you please review this

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
@@ -39,22 +39,22 @@ class AccountStatementImportSheetParser(models.TransientModel):
     _description = "Bank Statement Import Sheet Parser"
 
     @api.model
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Let's fast-track this bugfix, as the module is now broken with the prior change.

/ocabot merge patch

cvinh:
Hello
It looks like this [commit](https://github.com/OCA/bank-statement-import/commit/6deab7c481d5f363d9306ba1b5f73a2487da2126) has not been forwarded to v17 and PR to v18

pedrobaeza:
@cvinh indeed. Check #821

PR #763, opened 2025-01-30
[IMP] account_statement_import_sheet_file: XLS[X] file can actually be HTML

Daemo00:
I recently downloaded an XLS file that was actually an HTML file (they exist, check the file in tests/fixtures!) and the module **account_statement_import_sheet_file** wasn't able to import it.
With this PR, it is possible to import such files.

Most of the README edits happened automatically running pre-commit.

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
@@ -1,5 +1,6 @@
 # Copyright 2019 ForgeFlow, S.L.
 # Copyright 2020 CorporateHub (https://corporatehub.eu)
+# Copyright 2025 Simone Rubino
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import itertools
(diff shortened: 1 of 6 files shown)


account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -176,6 +174,25 @@ def _parse_lines(self, mapping, data_file, currency_code):
                         _("No valid encoding was found for the attached file")
                     ) from None
                 decoded_file = data_file.decode(detected_encoding)
+            is_HTML = decoded_file.lower().lstrip().startswith("<html>")
```
alexey-pelykh, line 177: Nit: `is_HTML` → `is_html` per PEP 8 (`snake_case` for local variables).

alexey-pelykh (approved):
Thanks for tackling this — banks exporting XLS files that are actually HTML is a common pain point. The detection via `lstrip().startswith("<html>")` and conversion through lxml is a pragmatic solution.

One minor thing: `is_HTML` doesn't follow Python naming — should be `is_html`. Not blocking.

Code review LGTM.

Daemo00:
I'm no more working on this one because I have moved to `18.0` (https://github.com/OCA/bank-statement-import/pull/871), who wants can supersede.

PR #764, opened 2025-02-04
[14.0][IMP] ensure currency name comparison is case-insensitive

matteotrubini:
Avoid skipping lines just because 'eur' != 'EUR'


alexey-pelykh (approved):
LGTM

StefanRijnhart (approved):
Thanks! Will you propose a port to the latest version?

/ocabot merge patch

PR #765, opened 2025-02-21
[17.0][MIG] account_statement_import_paypal

manuelregidor:
Standard migration

@HaraldPanten @ValentinVinagre @luis-ron 

T-7540


john-herholz-dt:
@HaraldPanten the ocabot command did nothing. Are you a maintainer of this Repo?

ValentinVinagre:
> @HaraldPanten the ocabot command did nothing. Are you a maintainer of this Repo?

Why do you indicate that the bot command has not done anything? You can see that the PR with the module to be migrated has been indicated in the migration issue.

https://github.com/OCA/bank-statement-import/issues/636

Jesarregui (approved):
Functional Review: LGTM

john-herholz-dt:
> > @HaraldPanten the ocabot command did nothing. Are you a maintainer of this Repo?
> 
> Why do you indicate that the bot command has not done anything? You can see that the PR with the module to be migrated has been indicated in the migration issue.
> 
> #636

Ah that is what it did, thanks!

account_statement_import_paypal/models/account_statement_import_paypal_parser.py
```diff
@@ -207,9 +206,9 @@ def _convert_line_to_transactions(self, line):
             note += " (%s)" % payer_email
 
         unique_import_id = f"{transaction_id}-{int(timestamp.timestamp())}"
-        name = (invoice or details or description or "",)
+        name = f"invoice or details or description or ''-{transaction_id}"
```
Tisho99: What is this? 

I think the correct code is:

name = invoice or details or description or "

Tisho99 (changes requested):

HaraldPanten:
> > > @HaraldPanten the ocabot command did nothing. Are you a maintainer of this Repo?
> > 
> > 
> > Why do you indicate that the bot command has not done anything? You can see that the PR with the module to be migrated has been indicated in the migration issue.
> > #636
> 
> Ah that is what it did, thanks!

No problem. Yes, I am one of the PSC team in accounting and banking.

account_statement_import_paypal/migrations/14.0.1.0.0/pre-migrate.py
```diff
@@ -1,40 +0,0 @@
-# Copyright 2021 Tecnativa - Carlos Roca
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
-from openupgradelib import openupgrade
```
StefanRijnhart: @pedrobaeza Can I ask you about the policy to remove migration scripts upon migration to a new version? If someone migrates from 13.0 to 17.0 in one go using the Odoo migration service, they will not pass through intermediate version of the OCA module, and this migration script will still be required. Maybe we should not delete the scripts.

ValentinVinagre: @StefanRijnhart it is an interesting point to discuss.

pedrobaeza: There are several reasons for deleting them:

- You may perform some changes `[OU-FIX]` or `[OU-IMP]` on target branch scripts, and these ones won't contain that changes (and ask to forward-port them are out of scope).
- pre and post migration scripts are designed to be executed in the exact pipeline point, and it's expected to have the DB schema that is present in that moment. If you have skipped the upgrade of OCA modules in between each version, you may be losing DB schema changes performed in the module update phase from which you depend to execute that scripts, having crashes. So removing them is also to remove false expectations.

That's also why I don't recommend to use enterprise upgrade services if you are jumping more than one version and you have OCA modules. It requires you to develop ad-hoc migration scripts for these modules in case they are incompatible, but at least to check all of them, being more time-consuming than performing it with OpenUpgrade passing each version, and adding the available OCA modules on each branch.

Do you agree according this reasoning?

StefanRijnhart: I would prefer to keep the migration scripts for the scenario until it is clear that compatibility issues arise. Regardless of your recommendation, the Odoo enterprise migration is a good proposition for enterprise customers and in my experience, most migrations will skip one or more versions (maybe in this case the jump is quite large and the deletion justified tbh, but I wanted to take the opportunity to discuss). Indeed, in the case of incremental data model changes issues can arise, in which case we may expect PRs to fix an Odoo 14 migration script to keep on working when first run on Odoo 17 for example. I don't think it's a reason to drop migration scripts preemptively.

In this particular example, the data model changes are strictly limited to the data model provided by the module itself and will remain valid indefinitely. I think this is true in the majority of cases.

pedrobaeza: This can't be something selective, as developers doing the migration don't know if they are going to be compatible or not. Instead, you should perform an script that collects all the migration scripts for each branch and put them together in a folder.

And as said, if tomorrow something is amended in the scripts of the previous branch, this branch won't contain the correct migration scripts.

Please don't try to change the status quo in this that it's perfectly reasoned.

StefanRijnhart: Mmm, I understand that there are pros and cons to both approaches but the current approach is not 'perfectly reasoned'.

pedrobaeza: Both reasons I stated are strong enough for not going that way, Stefan. In fact, this guideline was added after some troubles with the 2 things I explained, which costed a lot to debug in that days.

StefanRijnhart: Can we agree that if the migration script only concerns the data model provided by the module we can ask in reviews to keep the migration script for some versions to come?

manuelregidor:
@Tisho99 Changes applied. Could you review again? Thank you

Tisho99 (approved):
LGTM

account_statement_import_paypal/migrations/14.0.1.0.0/pre-migrate.py
pedrobaeza: No, I still keep my position to not preserve it. First, because there may be divergences in the contents if later patches are applied. Second, explaining why to keep them or not will confuse developers migrating modules that don't have nothing to do with the module migration itself. And third, at the end, you are not getting any advantage if you are going to follow this route, as you may need the same to prepare a tool for gathering previous migration scripts. It's the same that this tool gathers some scripts or all of them. Being convenient to keep it your current migration project doesn't mean that it's the general convenience.

StefanRijnhart: I think this deserves to be discussed in a wider audience to determine what is the general convenience. I don't like that you dismiss my point of view as something that is only valid for 'my current migration project'. I do not actually have a concrete case at the moment, and if I did I could easily assemble the required scripts myself as I am quite well versed in the development history of the OCA modules I use. It is with less seasoned users in mind that I am suggesting to reconsider this (c.q. the 'general convenience').

StefanRijnhart (approved):
Comment about the migration script not blocking, because it concerns a very old version.

account_statement_import_paypal/migrations/14.0.1.0.0/pre-migrate.py
pedrobaeza: This concern was raised already in the past in OCA (don't remember where), but you can do it again. I don't get though why you think that is good for the general audience to have things that may or may not work, that may or may be outdated, etc. As said, this reduces false expectations. And you or someone seasoned are not present on most of the migrations to say "this scripts is OK to be kept and this one not", and even if you are there, depending on the combination of initial and target version, the results may vary. Anyway, specially these module scripts are touching DB structure, so they are not harmless.

HaraldPanten (approved):

PR #692, opened 2024-05-08
[17.0][MIG] account_statement_import_file: migration to 17.0

manuelregidor:
Supersedes https://github.com/OCA/bank-statement-import/pull/649
MIgration + Tests added

PR #767, opened 2025-02-24
[16.0][FIX] account_statement_import_file: fix attachment creation

alexis-via:
BEFORE :

bug_import

AFTER : 

fixed_st_import

Cause : in v14/15, account.bank.statement inherited mail.thread and we were using the chatter to display the attachments. cf https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_bank_statement.py#L119

starting from v16, account.bank.statement didn't inherit mail.thread any more and a new M2M field attachment_ids was added, cf https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_bank_statement.py#L100. But the code in account_statement_import_file was not updated accordingly.

If this PR is approved, I'll forward port to v17 and v18.

### DIFF
--- a/account_statement_import_file/wizard/account_statement_import.py
+++ b/account_statement_import_file/wizard/account_statement_import.py
@@ -4,7 +4,7 @@
 import base64
 import logging
 
-from odoo import _, api, fields, models
+from odoo import Command, _, api, fields, models
 from odoo.exceptions import UserError
 
 from odoo.addons.base.models.res_bank import sanitize_account_number
@@ -39,7 +39,6 @@ class AccountStatementImport(models.TransientModel):
                     "only contains already imported transactions."
                 )
             )
-        self.env["ir.attachment"].create(self._prepare_create_attachment(result))
         return result
 
     def import_file_button(self):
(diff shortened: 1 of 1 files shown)


account_statement_import_file/wizard/account_statement_import.py
```diff
@@ -290,6 +284,9 @@ def _complete_stmts_vals(self, stmts_vals, journal, account_number):
         speeddict = journal._statement_line_import_speeddict()
         for st_vals in stmts_vals:
             st_vals["journal_id"] = journal.id
+            attach_vals = self._prepare_create_attachment(journal)
+            if attach_vals:
+                st_vals["attachment_ids"] = [Command.create(attach_vals)]
```
Daemo00, line 289: This creates a new `ir.attachment` for each statement but we only need one `ir.attachment` and it only has to be linked do the statements.
Sure we can trust Odoo that it won't duplicate the file in the filesystem, but still, I think it should be clear that we only want to create 1 file and link it to the statements.

I have found that this issue has already been fixed in `17.0` with https://github.com/OCA/bank-statement-import/pull/692/commits/db8980dabe885f026e9aebcacc0fcded2eafaf9c, maybe you can cherry pick that commit? It would also keep the versions aligned.

Daemo00 (changes requested):
Thanks for thhe PR!
I have tried in `16.0` and the issue exists, and it is fixed from this PR.

I had a look at the code and I think there's a better way to do it, please have a look at the below comment.

Also: could you please add a test? Thanks!

PR #779, opened 2025-03-05
[16.0][IMP] Support Multi-Currency Bank Accounts in `account_statement_import_file`

fd-oncodna:
**`account_statement_import_file`: Support multiple journals (one per currency) for the same bank account number.**  

Our bank accounts operate in a *multi-currency* setup, meaning a single account number corresponds to distinct *sub-accounts*, each handling a different currency. We receive separate bank statement files for each sub-account, but the current module cannot handle this correctly—it selects only the first journal matching the account number.  

This PR addresses the issue by selecting *all* matching journals and, if necessary, further filtering them by currency.

### DIFF
--- a/account_statement_import_file/wizard/account_statement_import.py
+++ b/account_statement_import_file/wizard/account_statement_import.py
@@ -223,8 +223,11 @@ class AccountStatementImport(models.TransientModel):
                         sanitized_account_number,
                     ),
                 ],
-                limit=1,
             )
+            if len(journal) > 1:
+                journal = journal.filtered(
+                    lambda j: (j.currency_id or company.currency_id) == currency
+                )[:1]
             ctx_journal_id = self.env.context.get("journal_id")
             if journal and ctx_journal_id and journal.id != ctx_journal_id:
                 ctx_journal = journal_obj.browse(ctx_journal_id)


fd-oncodna:
Hi everyone, any chance this gets merged anytime soon?

fd-oncodna:
Anyone? @alexis-via @etobella? FYI it is used in production at OncoDNA since March.

fd-oncodna:
@alexis-via @etobella?

etobella:
I would be more in favor of a hook in the domain without this change and you make it in a custom domain.

fd-oncodna:
I understand your point, but I still believe my change makes more sense than the existing code. If there are multiple journals linked to the same bank account, selecting one at random is clearly not an ideal approach. Since this is not something frequent, most of the time it makes no difference but it does resolve the issue in cases where it matters. It’s not an exotic solution either - if there are multiple journals linked to the same bank account, it's most probable that they have different currencies. Honestly, I can’t think of any other situations that would require a specific domain.

pedrobaeza:
@etobella I know this case from Belgium where several accounts having the same number, so I don't think this is bad at all.

etobella (approved):
Well, If there is other people with the idea, I don't have a personal problem with the concept, I wanted to avoid issues if possible.

pedrobaeza (approved):
Please fw-port it to upper versions:

/ocabot merge patch

PR #782, opened 2025-03-09
[16.0][FIX] account_statement_import_online_gocardless: Swap bookingDate and valueDate

pedrobaeza:
Continuation of #689 

This fixes behavior when there are debit card transactions with bookingDate behind the actual valueDate. In current situation, this leads to be discarded later in here:

https://github.com/OCA/bank-statement-import/blob/e91e28117d5ae3fbb5e1fb168b734f6690689433/account_statement_import_online/models/online_bank_statement_provider.py#L345-L351

@Tecnativa


pedrobaeza:
Merging after being opened and with reviews in #689 

/ocabot merge patch

PR #785, opened 2025-03-11
[18.0][FIX] checklog-odoo.cfg: solve the merge issue

xaviedoanhduy:


checklog-odoo.cfg
```diff
@@ -1,7 +1,3 @@
 [checklog-odoo]
 ignore=
-<<<<<<< before updating
     WARNING.*0 failed, 0 error\(s\).*
-=======
-    WARNING.* 0 failed, 0 error\(s\).*
```
pedrobaeza, line 6: This is the one you should preserve.

xaviedoanhduy: done

pedrobaeza (approved):

PR #797, opened 2025-05-07
[16.0][FIX] account_statement_import_online_gocardless: Pass correct provider ID

pedrobaeza:
Using `active_id context` key may lead to incorrect values, as it depends on the wizard starting URL. As we have the provider ID already in a context key, let's use it to always make sure it works.

@Tecnativa TT55399


carlos-lopez-tecnativa (approved):

victoralmau (approved):

pedrobaeza:
Already fixed in 17.0.

PR #798, opened 2025-05-08
[16.0][IMP] account_statement_import_online: Pull wizard with dates and one fetch

pedrobaeza:
This commit improves the provider pull wizard in 2 ways:

- It asks for dates instead of datetimes, as at the end, the minimum pull unit is the whole day, and it was very confusing for the user to select times without knowing that behind the scenes, the whole day will be pulled. Now, the user selects only the day. Quicker and cleaner...
- It includes a check "One fetch", that instructs the pull method to ask to the provider the whole interval in one shot. This way, you don't consume the request rate limits for the providers with them, or only due to speed purposes. It's checked by default, as it's the more natural way of fetching an interval on demand.

@Tecnativa TT55399

### DIFF
--- a/account_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_statement_import_online/models/online_bank_statement_provider.py
@@ -188,13 +188,15 @@ class OnlineBankStatementProvider(models.Model):
                 )[0][1],
             }
 
-    def _pull(self, date_since, date_until):
+    def _pull(self, date_since, date_until, one_fetch=False):
         """Pull data for all providers within requested period."""
         is_scheduled = self.env.context.get("scheduled")
         debug = self.env.context.get("account_statement_online_import_debug")
         debug_data = []
         for provider in self:
             statement_date_since = provider._get_statement_date_since(date_since)
+            if one_fetch:
+                statement_date_since = date_since
             while statement_date_since < date_until:
                 # Note that statement_date_until is exclusive, while date_until is
                 # inclusive. So if we have daily statements date_until might
@@ -202,6 +204,8 @@ class OnlineBankStatementProvider(models.Model):
                 statement_date_until = (
                     statement_date_since + provider._get_statement_date_step()
                 )
+                if one_fetch:
+                    statement_date_until = date_until
                 try:
                     data = provider._obtain_statement_data(
                         statement_date_since, statement_date_until
--- a/account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
+++ b/account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import pprint
+from datetime import datetime
 
 from odoo import fields, models
 
@@ -12,15 +13,13 @@ class OnlineBankStatementPullWizard(models.TransientModel):
     _name = "online.bank.statement.pull.wizard"
     _description = "Online Bank Statement Pull Wizard"
 
-    date_since = fields.Datetime(
-        string="From",
-        required=True,
-        default=fields.Datetime.now,
-    )
-    date_until = fields.Datetime(
-        string="To",
-        required=True,
-        default=fields.Datetime.now,
+    date_since = fields.Date(string="From", required=True, default=fields.Date.today)
+    date_until = fields.Date(string="To", required=True, default=fields.Date.today)
+    one_fetch = fields.Boolean(
+        string="One fetch",
+        help="If checked, retrieve the whole interval in only one request to the "
+        "provider. This is useful if the provider sets request limits.",
+        default=True,
     )
 
     def _get_provider(self):
@@ -35,10 +34,20 @@ class OnlineBankStatementPullWizard(models.TransientModel):
             provider = active_record
         return provider
 
+    def _get_datetimes(self):
+        """Return the datetime values for the entered dates, including whole starting
+        and ending days.
+        """
+        self.ensure_one()
+        since = datetime.combine(self.date_since, datetime.min.time())
+        until = datetime.combine(self.date_until, datetime.max.time())
+        return since, until
+
     def action_pull(self):
         """Pull statements from provider and then show list of statements."""
         provider = self._get_provider()
-        provider._pull(self.date_since, self.date_until)
+        since, until = self._get_datetimes()
+        provider._pull(since, until, one_fetch=self.one_fetch)
         action = self.env.ref("account.action_bank_statement_tree").sudo().read([])[0]
         action["domain"] = [("journal_id", "=", provider.journal_id.id)]
         return action
@@ -46,10 +55,10 @@ class OnlineBankStatementPullWizard(models.TransientModel):
     def action_debug(self):
         """Pull statements in debug and show result."""
         provider = self._get_provider().with_context(
-            active_test=False,
-            account_statement_online_import_debug=True,
+            active_test=False, account_statement_online_import_debug=True
         )
-        data = provider._pull(self.date_since, self.date_until)
+        since, until = self._get_datetimes()
+        data = provider._pull(since, until, one_fetch=self.one_fetch)
         wizard = self.env["online.bank.statement.pull.debug"].create(
             {"data": pprint.pformat(data)}
         )
--- a/account_statement_import_online/wizards/online_bank_statement_pull_wizard.xml
+++ b/account_statement_import_online/wizards/online_bank_statement_pull_wizard.xml
@@ -13,6 +13,7 @@
                 <group name="main">
                     <field name="date_since" />
                     <field name="date_until" />
+                    <field name="one_fetch" />
                 </group>
                 <footer>
                     <button
--- a/account_statement_import_online_gocardless/tests/test_account_statement_import_online_gocardless.py
+++ b/account_statement_import_online_gocardless/tests/test_account_statement_import_online_gocardless.py
@@ -126,6 +126,7 @@ class TestAccountBankAccountStatementImportOnlineGocardless(common.TransactionCa
         vals = {
             "date_since": "2020-10-30",
             "date_until": "2020-11-11",
+            "one_fetch": False,
         }
         wizard = (
             self.env["online.bank.statement.pull.wizard"]
--- a/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
+++ b/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py
@@ -35,7 +35,7 @@ class OnlineBankStatementProvider(models.Model):
             ("ponto", "MyPonto.com"),
         ]
 
-    def _pull(self, date_since, date_until):
+    def _pull(self, date_since, date_until, one_fetch=False):
         """For Ponto the pulling of data will not be grouped by statement.
 
         Instead we will pull data from the last available backwards.
@@ -52,7 +52,7 @@ class OnlineBankStatementProvider(models.Model):
         debug = self.env.context.get("account_statement_online_import_debug")
         debug_data = []
         data = super(OnlineBankStatementProvider, self - ponto_providers)._pull(
-            date_since, date_until
+            date_since, date_until, one_fetch=one_fetch
         )
         if debug:
             debug_data += data


account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -188,20 +188,24 @@ def _compute_update_schedule(self):
...
+            if one_fetch:
+                statement_date_since = date_since
             while statement_date_since < date_until:
                 # Note that statement_date_until is exclusive, while date_until is
                 # inclusive. So if we have daily statements date_until might
                 # be 2020-01-31, while statement_date_until is 2020-02-01.
                 statement_date_until = (
                     statement_date_since + provider._get_statement_date_step()
                 )
+                if one_fetch:
```
alexey-pelykh, line 207: How checking this by default is compatible with providers that limit fetch to X days per fetch?

alexey-pelykh (changes requested):

account_statement_import_online/models/online_bank_statement_provider.py
pedrobaeza: Can you put any example of that limit? All that I know don't have such limit. And anyway, if you put a big interval, you get in return the message error if so, so you can limit your interval in the wizard. Remember that this is only for the interactive pull wizard.

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
```diff
@@ -12,15 +13,13 @@ class OnlineBankStatementPullWizard(models.TransientModel):
...
-        string="To",
-        required=True,
-        default=fields.Datetime.now,
+    date_since = fields.Date(string="From", required=True, default=fields.Date.today)
+    date_until = fields.Date(string="To", required=True, default=fields.Date.today)
+    one_fetch = fields.Boolean(
+        string="One fetch",
+        help="If checked, retrieve the whole interval in only one request to the "
+        "provider. This is useful if the provider sets request limits.",
+        default=True,
```
alexey-pelykh, line 22: This is not universally true for all providers and `_get_statement_date_step` exists for a reason that some providers set limit to how many days may be requested in one go. If some provider does not have that limit but limits request _rate_, alter the behavior with `_get_statement_date_step` being maxed out.

account_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: E.g. PayPal - one year only, Wise.com - one year only. User may as well select 5 years when importing for the first time on new setup. Such providers should set the step to 1 year making the `one_step` setting irrelevant.

Obviously you're aiming to fix experience for some specific providers - what step do they set? Since if they set one day but can handle much more so that one_fetch works - those providers' step should be fixed and not an override added.

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
pedrobaeza: I insist: please provide one example of provider with such limit. And again, this is only a convenient default value for users doing interactively the pull through the wizard. You can uncheck it. If you prefer, we can put an overridable method on the to return True or False depending on the provider, but I think this is over-engineering something so straight-forward.

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
```diff
@@ -12,15 +13,13 @@ class OnlineBankStatementPullWizard(models.TransientModel):
...
-    date_since = fields.Datetime(
-        string="From",
-        required=True,
-        default=fields.Datetime.now,
-    )
-    date_until = fields.Datetime(
-        string="To",
-        required=True,
-        default=fields.Datetime.now,
+    date_since = fields.Date(string="From", required=True, default=fields.Date.today)
```
alexey-pelykh, line 16: I would very much prefer having date_since as date but have datetime_since hidden unless debug is active, quite useful when dealing with consequences of some outdages

alexey-pelykh: So that regular users are not scared off by datetime but admins can do their job.

account_statement_import_online/models/online_bank_statement_provider.py
pedrobaeza: Yeah, and if they do that, an error will be arisen from the Paypal provider, so I don't see any problem. Instead, with the default value that is daily, they will make 365*5 = 1825 requests, taking a lot of time and make them to think that it's hung.

We can put a warning if you prefer saying that the provider may not accept the petition in one block, but what is not sane is to not do it by default.

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
pedrobaeza: Alexey, the datetime doesn't matter, as it's transformed before pulling:

https://github.com/OCA/bank-statement-import/blob/be1727629813227f962b8032523904af84484658/account_statement_import_online/models/online_bank_statement_provider.py#L445

so debug or not, you won't get the exact time pulling.

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
alexey-pelykh: I already provided - https://github.com/OCA/bank-statement-import/pull/798#discussion_r2079099506
I counter-request - what providers don't set a limit for a timespan that can be requested in one go [that have this setting]? what providers will benefit from `one_fetch` - and what is their step?

pedrobaeza: GoCardless, Ponto... Their API doesn't indicate any limit in from/to, and if any hidden, as said, they will return the corresponding error for doing the user aware. On most of the cases, you won't hit that limit. It's not realistic to say that you are going to regularly download more than one year of Paypal things, and if so, as said, you will find the error.

alexey-pelykh: Thanks for the info! Also, putting this aside:
https://github.com/OCA/bank-statement-import/blob/16.0/account_statement_import_online/models/online_bank_statement_provider.py#L453 why not to add `statement_creation_mode == "all"` or False? `one_fetch` seems quite like a hacky override when there's a perfect way via statement_creation_mode to configure provider

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
alexey-pelykh: Indeed, I take it back :)

account_statement_import_online/models/online_bank_statement_provider.py
alexey-pelykh: Let's use just one of the two threads :D why adding `statement_creation_mode` like `all` was not an option?

pedrobaeza: Because you want to do this only once, not for every automatic pull that happens once you do the initial pulling, that is only needed to be updated. Another possible use is any temporary problem that prevented to download an interval, and in such case, you also want only one pull, not one by one. Example: during a week, you have invalid credentials. Once you put the valid ones, having create statements as daily, only last day will be downloaded, so you click on the pull wizard and download the last week. Why would you want to do this by default through 7 requests instead of just one?

alexey-pelykh: Let's pause for a second to investigate the idea behind `statement_creation_mode` having day/week/month:
By configuring `statement_creation_mode` on the provider to day/week/month we control how many statements area created, or in other words - how transactions are grouped.

If I got your argument for `one_fetch` correctly, it is proposed to solve the API rate limiting. As a side effect it overrides the `statement_creation_mode`, creating a single statement from all. I definitely see issue with that.
In your example, with daily setting, and one_fetch checked - one statement covering entire period will be generated.

I fully acknowledge the API rate limiting issue that may arise for some providers when you pull yearly data with day granularity. However I strongly disagree with the proposed approach.

As a counter-proposal, I'd suggest:

- adding `_get_statement_date_step_max()` (None by default) so that providers like PayPal may set it to 365 days.
- refactoring the code to use `_get_statement_date_step_max` (effectively hardcoding `one_fetch` behavior) for fetch
- refactoring code to step through the downloaded data according to `statement_creation_mode`

that would allow to fetch data in minimal amount of requests (one for some providers, couple for providers like PayPal) while respecting the statement grouping granularity of day/week/month

victoralmau (commented):
IMO this change is coherent.

account_statement_import_online/wizards/online_bank_statement_pull_wizard.py
```diff
@@ -35,21 +34,31 @@ def _get_provider(self):
             provider = active_record
         return provider
 
+    def _get_datetimes(self):
+        """Return the datetime values for the entered dates, including whole starting
+        and ending days.
+        """
+        self.ensure_one()
```
alexey-pelykh, line 41: **Timezone concern**: `datetime.combine(self.date_since, datetime.min.time())` produces a naive datetime (no tzinfo). When `one_fetch=True`, this value is used directly as `date_since` in `_pull`, bypassing `_get_statement_date_since()` which normally handles date alignment. For providers operating in non-UTC timezones, this could cause off-by-one-day boundary issues.

Consider using `fields.Datetime.to_datetime()` or applying the provider's timezone to ensure consistency with the normal code path.

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -188,20 +188,24 @@ def _compute_update_schedule(self):
                 )[0][1],
             }
 
-    def _pull(self, date_since, date_until):
+    def _pull(self, date_since, date_until, one_fetch=False):
         """Pull data for all providers within requested period."""
         is_scheduled = self.env.context.get("scheduled")
         debug = self.env.context.get("account_statement_online_import_debug")
```
alexey-pelykh, line 194: **Design concern**: This overrides `_get_statement_date_since()` which providers may use to align the start date to statement boundaries (e.g., start of week/month for weekly/monthly statement creation modes). Bypassing it means the fetched data may not align with the provider's expected statement boundaries.

The `_get_statement_date_since` method exists to allow providers to customize this behavior -- silently skipping it via a boolean flag breaks that contract.

account_statement_import_online/models/online_bank_statement_provider.py
```diff
@@ -188,20 +188,24 @@ def _compute_update_schedule(self):
...
+                statement_date_since = date_since
             while statement_date_since < date_until:
                 # Note that statement_date_until is exclusive, while date_until is
                 # inclusive. So if we have daily statements date_until might
                 # be 2020-01-31, while statement_date_until is 2020-02-01.
                 statement_date_until = (
                     statement_date_since + provider._get_statement_date_step()
                 )
+                if one_fetch:
+                    statement_date_until = date_until
```
alexey-pelykh, line 208: **Statement grouping side effect**: Setting `statement_date_until = date_until` here means a single iteration of the while loop, producing one statement for the entire period. This overrides `statement_creation_mode` (daily/weekly/monthly) -- the user configures daily statements but gets a single multi-day statement when using the wizard with `one_fetch` checked.

Fetch optimization and statement grouping are separate concerns that should not be coupled through a single flag.

alexey-pelykh (commented):
## Review: `account_statement_import_online` Pull Wizard Improvements

Thanks for the contribution, @pedrobaeza. The Date-vs-Datetime UX improvement is valuable. However, the `one_fetch` mechanism raises architectural concerns that merit further discussion before merging.

### What works well

- **Date fields instead of Datetime**: Good UX improvement. Users selecting dates is more intuitive when the minimum pull granularity is a full day.
- **`_get_datetimes()` helper**: Clean conversion method with proper `min.time()` / `max.time()` boundaries.
- **Ponto override propagation**: The `_pull` signature change is correctly forwarded through `super()` in the Ponto provider.

### Design concerns with `one_fetch`

1. **Bypasses `_get_statement_date_since`**: When `one_fetch=True`, the method overrides `statement_date_since = date_since` at line 194, skipping the provider-specific logic in `_get_statement_date_since()` that may adjust the start date (e.g. for providers that need alignment to statement boundaries). This silent override could produce unexpected results for providers that rely on that adjustment.

2. **Bypasses `_get_statement_date_step` grouping**: When `one_fetch=True`, `statement_date_until = date_until` at line 208 means a single call to `_obtain_statement_data` for the entire period. The resulting statement will cover the full date range regardless of `statement_creation_mode` (daily/weekly/monthly). This is a side effect -- the user configures daily statements but gets one big statement when pulling interactively.

3. **Timezone concern in `_get_datetimes()`**: The method produces naive datetimes via `datetime.combine(date, datetime.min.time())`. When `one_fetch=True`, the normal path through `_get_statement_date_since` (which handles timezone-aware date alignment) is skipped. This could cause subtle off-by-one-day issues for providers in non-UTC timezones.

4. **Default `True` is risky**: Defaulting `one_fetch=True` means the wizard will by default bypass the carefully designed step-based iteration for all providers, including future ones that may depend on it. A provider author implementing `_get_statement_date_step` to return 30 days (because their API limits requests to 30-day windows) would find that the wizard silently ignores their constraint.

### Alternative approach suggestion

As discussed in the earlier review thread, a cleaner architecture might be:
- Add `_get_statement_date_step_max()` returning `None` by default (unlimited)
- Providers with API limits (PayPal, Wise) override it to their max window
- The fetch loop uses `_get_statement_date_step_max` for fetching but still creates statements per `statement_creation_mode`

This separates the two concerns: **fetch granularity** (how many API calls) vs **statement grouping** (how transactions are organized).

### Test coverage

The GoCardless test is updated to set `one_fetch=False`, preserving existing behavior. However, there is no test exercising the `one_fetch=True` path to verify it actually works correctly (single fetch, statement boundaries, etc.).

### Verdict

The Date field change is ready to merge on its own. The `one_fetch` mechanism needs design refinement to avoid bypassing provider-specific constraints and statement grouping configuration. Consider splitting this into two PRs: one for the Date UX improvement, and one for the fetch optimization after the architectural approach is agreed upon.

PR #802, opened 2025-05-28
[16.0][FIX] account_statement_import_online: Adapt wizard datetimes to user…

aritzolea:
… timezone

UTC datetimes where being used for statement importing, this could bring wrong statements.


aritzolea:
Errors seem not to be related to the changes in the PR

PR #804, opened 2025-06-17
[17.0][IMP] account_statement_import_sheet_file: Current year if no year

pedrobaeza:
There can be files where the date is only expressed with day and month, but no year. You put in that case a timestamp expression like '%d/%m'. For that, strptime parses the content and return the day and month in the year 1900.

This code checks that condition and replace the year by the current one, as clearly 1900 is not a year to post transactions. There's also logic for putting previous year if the current month is lower than the imported one (i.e., current month is January and you are importing December).

@Tecnativa


carlosdauden:
Maybe you can check if the current month is January and the imported movements are from December to set the previous year, or some other smart algorithm to avoid problems with the change of year.

pedrobaeza:
Yeah, good idea. Added.

carlosdauden (approved):

victoralmau (approved):

PR #805, opened 2025-06-30
[17.0][FIX] account_statement_import_sheet_file: fix none decimal separator

jbaudoux:
Allow to parse "123456" as 1234.56 when using "none" as decimal separator

Before it was returning "1.2.3.4.5.6" and crashing

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
@@ -1,5 +1,6 @@
 # Copyright 2019 ForgeFlow, S.L.
 # Copyright 2020 CorporateHub (https://corporatehub.eu)
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import itertools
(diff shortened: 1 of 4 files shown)


alexey-pelykh:
I'm not sure I'm aboard this the fix - if there's no decimal separator and 123456 is given, it should be parsed as 123456 - why should it work otherwise?

jbaudoux:
> I'm not sure I'm aboard this the fix - if there's no decimal separator and 123456 is given, it should be parsed as 123456 - why should it work otherwise?

It will be parsed as 123456 if you configure there should be a separator and the value doesn't contain any.
The bank provides a file without separator and the currency has 2 decimals. When the amount 123456 is given, it means 1234.56
Before this it was adding a "." between each figure so I guess that option never worked and has never been used

alexey-pelykh:
> It will be parsed as 123456 if you configure there should be a separator and the value doesn't contain any.

I think it calls for a documentation extension that "If no decimal delimiter is configured, ...... Example: ...."
Otherwise I'm good with the PR

jbaudoux:
help added on decimal separator field

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -506,5 +507,14 @@ def _parse_decimal(self, value, mapping):
...
+        float_separator = "."
+        if decimal:
+            value = value.replace(decimal, float_separator)
+        else:
+            journal = self.env["account.journal"].browse(
+                self.env.context.get("journal_id")
+            )
+            currency = journal.currency_id or journal.company_id.currency_id
+            decimal_places = currency.decimal_places if currency else 2
+            value = value[:-decimal_places] + float_separator + value[-decimal_places:]
```
alexey-pelykh, line 519: Nit: if `decimal_places` happens to be 0 (e.g. JPY), `value[:-0]` evaluates to `""` and this would produce `".123456"` instead of `"123456"`. Unlikely scenario for "none" separator, but a guard like `if decimal_places:` around the slicing (with a plain `value` passthrough otherwise) would make this bulletproof.

alexey-pelykh (approved):
Thanks for the fix and for adding the help text. The root cause is correctly identified -- `str.replace("", ".")` inserts a dot between every character, so checking `if decimal:` before the replace is the right approach. The currency-based decimal shift logic and the fallback to 2 decimal places are both sound. One minor edge case: if `decimal_places` is 0 (e.g. JPY), `value[:-0]` would produce an empty string and the result would be incorrect, but that's an unlikely combination with "none" separator and not a blocker.

PR #808, opened 2025-07-13
[17.0][FIX] account_statement_import_online_gocardless: token refresh/renew behavior

pedrobaeza:
Forward-port of #748 

Correction of the endpoint token/refresh request datas, plus the behavior of the _gocardless_get_token method which in case of a non expired refresh token was calling the endpoint token/new instead of token/refresh.

@Tecnativa


pedrobaeza:
Merging as already merged in previous version and for homogenizing all the branches with the proper patch:

/ocabot merge patch

PR #817, opened 2025-07-30
[16.0][FIX] account_statement_import_file: Corrects amounts with currencies when importing wigh foreign_currency

sergiobstoj:
When importing statements with specific foreign currency and amount, Journal Entry created is not entirely correct.
image

When applying this fix, journal entry is created as expected:
image


sergiobstoj:
@alexey-pelykh could you review this please?

PR #819, opened 2025-08-04
[16.0][FIX] account_statement_import_sheet_file: fix header_lines_skip_count behavior and fix tests accordingly

cvinh:
Prior to this commit, there was some confusion between indexes and line numbers. We’ve renamed several variables to improve clarity.
As a result, we updated the existing tests and added new ones to cover additional use cases.


cvinh:
I don't manage to fix pre-commit...

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
```diff
@@ -525,6 +626,58 @@ def test_metadata_separated_debit_credit_xlsx(self):
...
+                "amount_debit_column": "3",
+                "amount_credit_column": "4",
+            }
+        )
+        data = self._data_file(
+            "fixtures/meta_data_separated_credit_debit_no_header.xlsx"
+        )
+        wizard = self.AccountStatementImport.with_context(journal_id=journal.id).create(
+            {
+                "statement_filename": "fixtures/meta_data_separated_credit_debit_no_header.xlsx",
```
pedrobaeza: This line is too long.

cvinh: > This line is too long.

Yes but if I cut it, pre-commit puts it back

pedrobaeza: Put the string in a variable before, and put here the variable.

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -212,16 +209,19 @@ def _parse_rows(self, mapping, currency_code, data, columns):  # noqa: C901
         else:
             numrows = len(str(data_file.strip()).split("\\n"))
 
-        label_line = mapping.header_lines_skip_count
-        footer_line = numrows - mapping.footer_lines_skip_count
+        header_skip = mapping.header_lines_skip_count
+        data_first_index_line = not mapping.no_header and header_skip + 1 or header_skip
```
alexey-pelykh:
```python
data_first_index_line = header_skip if mapping.no_header else header_skip
```
or 
```python
data_first_index_line = header_skip + (0 if mapping.no_header else 1)
```

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -233,7 +233,11 @@ def _parse_rows(self, mapping, currency_code, data, columns):  # noqa: C901
                         cell_value = xldate_as_datetime(cell_value, book.datemode)
                     values.append(cell_value)
             else:
-                if index >= footer_line:
+                if mapping.no_header and index < data_first_index_line:
```
alexey-pelykh, line 236: I'm not sure I follow this part

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -212,16 +209,19 @@ def _parse_rows(self, mapping, currency_code, data, columns):  # noqa: C901
...
+        data_line_count = data_last_line - data_first_line + 1
 
         if isinstance(csv_or_xlsx, tuple):
-            rows = range(label_line, footer_line)
+            rows = range(data_first_index_line, data_last_line)
         else:
             rows = csv_or_xlsx
 
         lines = []
-        for index, row in enumerate(rows, label_line):
```
alexey-pelykh, line 224: Why not to start with `header_skip`?

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
alexey-pelykh: or even
```python
data_first_index_line = mapping.header_lines_skip_count + (0 if mapping.no_header else 1)
```

alexey-pelykh (changes requested):

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
cvinh: I prefer 
`data_first_index_line = header_skip + (0 if mapping.no_header else 1)`
for better understanding

cvinh: > or even
> 
> ```python
> data_first_index_line = mapping.header_lines_skip_count + (0 if mapping.no_header else 1)
> ```

We need the variable header_skip lower so I prefer to keep it

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
cvinh: > Why not to start with `header_skip`?

I did a wrong answer before, sorry for the wrong notification.
The reason why I choose not to start with `header_skip` is that if we have header lines to skip, I had a problem with the index and row and it removed the first lines of data.
So I prefered to start from the beginning and add conditions for different use cases

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
cvinh: > I'm not sure I follow this part

If we start the index at first line of data, we have to ignore all the header lines. Then, in the case without header, as long as we are before the data_first_index_line, we continue

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -212,16 +209,19 @@ def _parse_rows(self, mapping, currency_code, data, columns):  # noqa: C901
         else:
             numrows = len(str(data_file.strip()).split("\\n"))
 
-        label_line = mapping.header_lines_skip_count
-        footer_line = numrows - mapping.footer_lines_skip_count
+        header_skip = mapping.header_lines_skip_count
+        data_first_index_line = header_skip + (0 if mapping.no_header else 1)
```
mondot, line 213: Based on what I see about your changes in the existing tests, I think there might be some confusion about the use of the `mapping.no_header` parameter.

My first thought when reviewing your changes was that the parameter `mapping.no_header` creates a lot of complications and that your code could be much simpler if you ignored it.

From what I can see in your changes to the existing tests, I believe I’m correct in saying that the `no_header` param is only meant to serve as a marker indicating: "will I find integers or strings in the colum params ?".
Therefore, when someone uses the `header_lines_skip_count` and `footer_lines_skip_count` parameters, they should count the header line (if it exists). That makes your code much more readable, and the changes in this PR as well.

The code would look like

```python
        data_first_index_line = mapping.header_lines_skip_count
        data_last_line = numrows - mapping.footer_lines_skip_count
        data_line_count = data_last_line - data_first_index_line
        
        [...]
                if index < data_first_index_line:
                    continue
                if index >= data_first_index_line + data_line_count:
                    continue
```
This way, you'd only need to add your new tests without changing the existing ones.

I hope this is clear. Please feel free to let me know if it’s not, or if I’ve misunderstood something.

mondot (changes requested):
Hello,
I've just ran into this same bug. Thanks for proposing this PR :slightly_smiling_face:

mondot:
Hello @cvinh, I hope you're well :)
Any news on this PR ?

cvinh:
Yes I'll be able to work on it tomorrow

mondot:
> Yes I'll be able to work on it tomorrow

Great news :) 
I'm currently at the OCA days, we can meet there if you come this year :)

cvinh:
> Great news :)
> I'm currently at the OCA days, we can meet there if you come this year :)

I'll arrive at end of the day

cvinh:
As discussed yesterday with @mondot  the approach here is to better understand the 3 parameters no_header , skip_header_line and skip_footer_line
I think we managed to agree and he's going to review again

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
mondot: I propose an alternative in this PR : https://github.com/OCA/bank-statement-import/pull/834/files

mondot:
> As discussed yesterday with @mondot the approach here is to better understand the 3 parameters no_header , skip_header_line and skip_footer_line I think we managed to agree and he's going to review again

Hello, I hope everyone is well.
I propose an alternative there : https://github.com/OCA/bank-statement-import/pull/834
I would love to have a review there :)

Best regards

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
alexey-pelykh, line 236: This filtering logic is difficult to reason about and I suspect has off-by-one issues.

For the `no_header=False` case: `index` comes from `enumerate(rows)` starting at 0, where `rows` is the CSV reader that yields **all** remaining rows (after `parse_header` has consumed some via `next()`). Meanwhile `data_line_count = data_last_line - data_first_line + 1` is computed from absolute row counts. The comparison `index >= data_line_count` assumes that the CSV reader's position perfectly aligns with the computed data count, but `parse_header` consumes `header_skip` rows via `next()` plus one more for the actual header, so the reader's position depends on what `parse_header` did.

For the `no_header=True` case: `parse_header` returns `[]` without consuming any rows, so the CSV reader starts at row 0 and `index` truly is the absolute row number. But then `header_skip + data_line_count` as an upper bound mixes a skip count with a derived count in a way that's hard to verify.

I would recommend either:
1. Computing an explicit `rows_to_skip_at_start` and `rows_to_skip_at_end` count relative to the CSV reader's current position, or
2. Adding inline comments documenting what `index` represents in each case and what invariants hold

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -163,14 +163,16 @@ class AccountStatementImportSheetMapping(models.Model):
     )
     footer_lines_skip_count = fields.Integer(
         string="Footer lines skip count",
-        help="Set the Footer lines number."
-        "Used in some csv/xlsx file that integrate meta data in"
-        "last lines.",
+        help="Define the number of footer lines to be skipped in CSV/XLSX"
+        "files containing metadata at the bottom.",
         default="0",
     )
```
alexey-pelykh, line 169: Missing space between string literals. Python concatenates adjacent strings without adding a space, so this renders as:
`"...skipped in CSV/XLSXfiles containing metadata at the bottom."`

Should be:
```python
help="Define the number of footer lines to be skipped in CSV/XLSX "
"files containing metadata at the bottom.",
```

Same issue applies to the `header_lines_skip_count` help text below (multiple missing spaces).

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -163,14 +163,16 @@ class AccountStatementImportSheetMapping(models.Model):
...
     header_lines_skip_count = fields.Integer(
         string="Header lines skip count",
-        help="Set the Header lines number.",
+        help="Define the number of header lines to be skipped in CSV/XLSX"
+        "files containing metadata at the top."
+        "The header line containing the titles of columns should not"
+        "be counted",
         default="0",
```
alexey-pelykh, line 176: Multiple missing spaces in this help text. It will render as:
`"...in CSV/XLSXfiles containing metadata at the top.The header line containing the titles of columns should notbe counted"`

Suggested fix:
```python
help="Define the number of header lines to be skipped in CSV/XLSX "
"files containing metadata at the top. "
"The header line containing the titles of columns should not "
"be counted",
```

account_statement_import_sheet_file/data/map_data.xml
```diff
@@ -8,7 +8,7 @@
     <record id="sample_statement_map" model="account.statement.import.sheet.mapping">
         <field name="name">Sample Statement</field>
         <field name="footer_lines_skip_count">0</field>
-        <field name="header_lines_skip_count">1</field>
+        <field name="header_lines_skip_count">0</field>
```
alexey-pelykh, line 11: Changing the default from `1` to `0` makes sense given the new semantics, but this is effectively a **breaking change for existing installations**. Any existing `sample_statement_map` records (or user copies of them) will still have `header_lines_skip_count=1`, which under the new semantics means "skip one line of metadata above the column headers" rather than the old meaning of "there is one header row."

Consider adding a `post_init_hook` to adjust existing records: `header_lines_skip_count = max(0, header_lines_skip_count - 1)` for all `account.statement.import.sheet.mapping` records.

alexey-pelykh (changes requested):
Thank you for tackling this confusion between indexes and line numbers -- the intent is good and I appreciate the added test fixtures. However, after careful analysis I believe the new CSV row-filtering logic in `_parse_rows` has correctness issues that need to be resolved before this can be merged.

I also notice that PR #834 proposes an alternative approach to the same bug. It would be worth coordinating to avoid duplicate/conflicting work.

---

### Critical: CSV row-filtering logic is hard to reason about and likely has off-by-one issues

The new filtering block (lines ~236-241 in the PR) uses three separate `continue` conditions with `index` coming from `enumerate(rows)` starting at 0, while `data_line_count`, `data_first_index_line`, and `header_skip` are computed from absolute line numbers. For XLSX, `rows = range(data_first_index_line, data_last_line)` naturally constrains the range, but for CSV the reader yields **all** rows and the filtering must be done manually.

The conditions:
```python
if mapping.no_header and index < data_first_index_line:
    continue
if not mapping.no_header and index >= data_line_count:
    continue
if mapping.no_header and index >= header_skip + data_line_count:
    continue
```

For the `no_header=False` case, only the second condition fires: `index >= data_line_count`. But `index` is a 0-based row counter over the CSV reader (which includes ALL rows -- header, data, and footer), while `data_line_count` is `data_last_line - data_first_line + 1` (a count of data rows). This means the code skips rows once the 0-based index exceeds the data count, which works only coincidentally when the header skip is 0. With larger header skips, the CSV reader will have already consumed header rows via `parse_header`'s `next()` calls, making the index offset unpredictable.

For the `no_header=True` case, both the first and third conditions fire, but `header_skip + data_line_count` is not obviously the right upper bound when `index` is 0-based from `enumerate`.

I would recommend:
1. Adding clear comments documenting what `index` represents in each branch
2. Considering a unified approach (e.g., tracking absolute line number instead of relying on enumerate offset)
3. Verifying with edge cases: what happens with `header_lines_skip_count=0, no_header=True, footer_lines_skip_count=0`?

### Moderate: Missing spaces in multi-line help strings

Python string literal concatenation does not insert spaces. The help texts will render as:
- `"...skipped in CSV/XLSXfiles containing metadata at the bottom."`
- `"...skipped in CSV/XLSXfiles containing metadata at the top.The header line..."`
- `"...should notbe counted"`

Each continuation line needs a leading space.

### Moderate: Semantic change to `header_lines_skip_count` affects existing data

The meaning of `header_lines_skip_count` changed: previously it counted header lines **including** the column header row (so `1` meant "there is one header row with column names"). Now it counts lines to skip **above** the column header row (so the column header row is always implicitly handled). This is reflected in the sample data XML changing from `1` to `0` and all tests reducing their skip counts by 1.

This is a valid design improvement, but it is a **breaking change** for existing installations. Users who had `header_lines_skip_count=1` (meaning "standard single header row") will now skip one extra line above the header. Consider:
- Adding a `pre_init_hook` or `post_init_hook` to migrate existing mapping records (`header_lines_skip_count = max(0, header_lines_skip_count - 1)`)
- Or at minimum documenting this behavioral change in the PR description

### Minor: Removed negative test cases reduce coverage

The PR removes two `assertRaises(UserError)` blocks from `test_offsets` and `test_skip_empty_lines`. These were testing that the default mapping (without offsets/skip-empty configured) would raise an error on malformed input. Even if the new semantics make those specific error conditions no longer apply, removing them without replacement means we lose coverage for error paths. If the semantics changed such that these files now parse without error, that should be explicitly tested and documented.

### Suggestion: Coordinate with PR #834

Reviewer @mondot has opened PR #834 as an alternative fix for the same underlying issue. It would be valuable to align on a single approach. The core question seems to be whether `no_header` + `header_lines_skip_count` should allow skipping metadata lines when there is no column header row -- both PRs agree on this, but differ in implementation.

---

PR #820, opened 2025-08-11
[17.0][FIX] account_statement_import_online_gocardless: Date type selectable

pedrobaeza:
On #689, the date selection was flipped arguing that the transactions are discarded for a date interval out the other date. I have checked that when requesting transactions over a period, GoCardless returns the transactions based on both dates, so it's not a problem that the transaction is discarded in one round, because in the other it will appear, so we can still use "bookingDate".

The problem can be if the booking date is previous to the value date, and the transaction arrives to the bank later. You'll lost it depending on several factors: the download interval, if the transaction is near the interval change.

Let's be conservative, keeping the current value, but allowing to select the other date value.

@Tecnativa TT57550

### DIFF
--- a/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
+++ b/account_statement_import_online_gocardless/models/online_bank_statement_provider.py
@@ -35,6 +35,11 @@ class OnlineBankStatementProvider(models.Model):
         store=True,
         readonly=False,
     )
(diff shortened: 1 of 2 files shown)


Rad0van (approved):
LGTM.

victoralmau (approved):

PR #828, opened 2025-09-09
[IMP] account_statement_import_camt: add a journal option to ignore transaction details

glitchov:
When importing bank statements, all payment return transactions are imported individually.
To improve the reconciliation performance, it's better to have an option to group all payment return in a one transaction.

Migration of #263

### DIFF
--- a/account_statement_import_camt/__manifest__.py
+++ b/account_statement_import_camt/__manifest__.py
@@ -8,5 +8,5 @@
     "website": "https://github.com/OCA/bank-statement-import",
     "category": "Banking addons",
     "depends": ["account_statement_import_file"],
-    "data": ["views/account_bank_statement_import.xml"],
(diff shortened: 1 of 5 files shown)


baimont:
> When importing bank statements, all payment return transactions are imported individually. To improve the reconciliation performance, it's better to have an option to group all payment return in a one transaction.
> 
> Migration of #263

This does not really group in one transaction, it just sets some fields (partner_id, ...) to False when importing camt statements, right? Your boolean is therefore well renamed.

Could you then check in your unit test that these informations are ignored on transactions? And not just test the number of statement lines.

baimont:
@glitchov @gva-acsone @sbejaoui @qgroulard

gva-acsone (approved):
LGTM functional test

PR #830, opened 2025-09-22
[16.0][FIX] online_bank_statement_provider_paypal: Avoid first transaction not found error in some cases

victoralmau:
Avoid first transaction not found error in some cases

We want to intentionally force 23:59:59 because in some very specific cases and without apparent explanation, a transaction that has a specific date and time defined is not obtained at that time, therefore, we want to ensure that it will be obtained in this request to avoid false errors.

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT58112


pedrobaeza (approved):

pilarvargas-tecnativa (approved):

victoralmau:
This is ready to merge.

PR #831, opened 2025-09-22
[16.0][IMP] account_statement_import_online: Avoid unnecessary log messages

victoralmau:
Avoid unnecessary log messages

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa

### DIFF
--- a/account_statement_import_online/models/account_journal.py
+++ b/account_statement_import_online/models/account_journal.py
@@ -45,9 +45,6 @@ class AccountJournal(models.Model):
                 journal.online_bank_statement_provider_id
                 and service == journal.online_bank_statement_provider_id.service
             ):
-                _logger.info(
-                    "Journal %s already linked to service %s", journal.name, service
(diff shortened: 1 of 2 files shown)


account_statement_import_online/models/account_journal.py
```diff
@@ -45,9 +46,12 @@ def _update_providers(self):
                 journal.online_bank_statement_provider_id
                 and service == journal.online_bank_statement_provider_id.service
             ):
-                _logger.info(
-                    "Journal %s already linked to service %s", journal.name, service
-                )
+                if not config["test_enable"] or (
+                    config["test_enable"] and service != "dummy"
+                ):
+                    _logger.info(
```
pedrobaeza: I would vote for directly remove the logger, as it's not bringing useful information at all.

pilarvargas-tecnativa (approved):

PR #832, opened 2025-09-23
[IMP] account_statement_import_online: Running schedule should not trigger redundant `write` on journal and provider

dzungtran89:
Running schedule should not trigger redundant `write` on journal and provider

### DIFF
--- a/account_statement_import_online/models/online_bank_statement_provider.py
+++ b/account_statement_import_online/models/online_bank_statement_provider.py
@@ -122,7 +122,10 @@ class OnlineBankStatementProvider(models.Model):
     def write(self, vals):
         """Set provider_id on journal after creation."""
         result = super().write(vals)
-        self._update_journals()
+
+        if "journal_id" in vals or "service" in vals:
+            self._update_journals()
+
         return result
 
     def _update_journals(self):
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
dot notation doesn't trigger writes until a flush is done, so I don't understand your change.

dzungtran89:
> dot notation doesn't trigger writes until a flush is done, so I don't understand your change.

The call `self.next_run` inside the loop called `write` many times on [this](https://github.com/OCA/bank-statement-import/blob/0561ee3901321bd594a76822d66f43c136dce8a3/account_statement_import_online/models/online_bank_statement_provider.py#L124) and [this](https://github.com/OCA/bank-statement-import/blob/0561ee3901321bd594a76822d66f43c136dce8a3/account_statement_import_online/models/online_bank_statement_provider.py#L134). Some breakpoints give us the hints

```
  /home/dzungt/code/oca/bank-statement-import/16.0/account_statement_import_online/models/online_bank_statement_provider.py(525)_adjust_schedule()
-> breakpoint()
  /home/dzungt/code/odoo/odoo/16.0/odoo/fields.py(1337)__set__()
-> records.write({self.name: write_value})
  /home/dzungt/code/oca/bank-statement-import/16.0/account_statement_import_online/models/online_bank_statement_provider.py(125)write()
-> breakpoint()
> /home/dzungt/code/oca/bank-statement-import/16.0/account_statement_import_online/models/online_bank_statement_provider.py(136)_update_journals()
-> {
(Pdb) 
```

The change is to avoid that.

StefanRijnhart (commented):
Thank you, this is actually the root cause of the excessive logging circumvented in https://github.com/OCA/bank-statement-import/pull/835.

Additionally, would you consider not updating the journals every time that the next run is updated?

image

StefanRijnhart:
Given that this is for 16.0 I will still submit my own 18.0 version in https://github.com/OCA/bank-statement-import/pull/853

pedrobaeza:
@dzungtran89 can you make the other change?

dzungtran89:
> @dzungtran89 can you make the other change?

Hello @pedrobaeza , I pushed the update.

@StefanRijnhart , thanks for the suggestion.

pedrobaeza (approved):
/ocabot merge patch

PR #836, opened 2025-09-29
[18.0][MIG] account_statement_import_caisse_epargne: Migration to 18.0

P-H-Phuc:
This module was initially part of the [Foodcoop set of addons](https://github.com/AwesomeFoodCoops/odoo-production/tree/12.0). We're now migrating it to 18.0 and taking this occasion to contribute it to OCA.

- Renamed `account_bank_statement_import_caisse_epargne` to `account_statement_import_caisse_epargne`
- Since Odoo 16.0, `account_bank_statement_import` module was renamed to `account_statement_import_file` (See [pull#604](https://github.com/OCA/bank-statement-import/pull/604))
- Add tests for statement import


account_statement_import_caisse_epargne/wizard/account_statement_import.py
```diff
@@ -0,0 +1,245 @@
...
+        (
+            file_version,
+            bank_group_code,
+            openning_date,
+            closing_date,
+            bank_account_number,
+            opening_balance,
+            closing_balance,
+            currency,
+        ) = result
```
alexey-pelykh, line 167: `data_file` here is a list (from `.splitlines()` above), but `super()._parse_file()` expects `bytes`. This will crash for any non-Caisse d'Épargne file. Store the original bytes before decoding.

alexey-pelykh (commented):
Thanks for the migration! Found a couple of things:

In `_parse_file`, `data_file` gets decoded and split into lines early on, but when the file isn't recognized as Caisse d'Épargne format, `super()._parse_file(data_file)` is called with the already-split list instead of the original bytes. This would crash on any non-matching file. Consider keeping the original bytes in a separate variable.

Also, there are two `.pot` files — `account_bank_statement_import_caisse_epargne.pot` (old module name) and `account_statement_import_caisse_epargne.pot`. The old one should be removed, and the remaining one should reference the current module name internally.

account_statement_import_caisse_epargne/wizard/account_statement_import.py
Kimkhoi3010: Thanks @alexey-pelykh, I have updated the PR.

PR #838, opened 2025-10-03
[IMP] account_statement_import_camt - import multi bank accounts statements

baimont:
Goal is to be able to import CAMT statements

- from Import Statement menu: should import everything and assign the right journal to each bank statement imported (was not the case, would keep the last one)
- from a journal via Import (OCA) button: should import all the statements filtered for that journal's account number and currency

This is based on an old pr:
https://github.com/OCA/bank-statement-import/pull/213

Issue over here https://github.com/odoo/odoo/issues/33162


sbejaoui (approved):

account_statement_import_camt/models/account_statement_import.py
```diff
@@ -14,22 +14,18 @@ class AccountStatementImport(models.TransientModel):
...
-                    account_number = None
-                    transactions = []
+                    result = []
                     for member in data.namelist():
-                        currency, account_number, new = self._parse_file(
-                            data.open(member).read()
-                        )
-                        transactions.extend(new)
-                return currency, account_number, transactions
+                        result = self._parse_file(data.open(member).read())
```
MaxyMoos, line 26: This line overwrites the `result` list declared above, should be changed to `parsed = self._parse_file(...)` then `result.extend(parsed)` ?

alexey-pelykh, line 26: Bug here: `result` is reassigned to the return value of `_parse_file`, then `result.extend(result)` doubles it (extends the list with itself).

Should be:
```python
result = []
for member in data.namelist():
    parsed = self._parse_file(data.open(member).read())
    result.extend(parsed)
```

Compare with #865 which handles this correctly.

alexey-pelykh (changes requested):
Thanks for working on multi-bank account support -- this is a real pain point when banks export all accounts into a single CAMT file.

The journal-based filtering logic in the parser looks solid, but there's a bug in the zip handling in `_parse_file` that would double entries for zip imports. See inline comment.

Also heads up: #865 modifies the same files, so these two PRs will conflict.

PR #839, opened 2025-10-03
[14.0][IMP] account_statement_import_txt_xlsx: add space in the thousands s…

matthieusaison:
…eparator selection

### DIFF
--- a/account_statement_import_txt_xlsx/models/account_statement_import_sheet_mapping.py
+++ b/account_statement_import_txt_xlsx/models/account_statement_import_sheet_mapping.py
@@ -15,7 +15,12 @@ class AccountStatementImportSheetMapping(models.Model):
     )
     float_thousands_sep = fields.Selection(
         string="Thousands Separator",
-        selection=[("dot", "dot (.)"), ("comma", "comma (,)"), ("none", "none")],
+        selection=[
+            ("dot", "dot (.)"),
+            ("comma", "comma (,)"),
+            ("space", "space ( )"),
+            ("none", "none"),
+        ],
         default="dot",
     )
     float_decimal_sep = fields.Selection(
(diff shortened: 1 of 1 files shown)


alexey-pelykh (commented):
## Review: [14.0][IMP] account_statement_import_txt_xlsx: add space in the thousands separator selection

**Note: This PR targets the `14.0` branch, which has reached end-of-life.** OCA no longer actively maintains 14.0 modules, so this PR will not be reviewed in detail. The contributor should consider porting this change to a maintained branch (16.0 or 18.0) where it would have a realistic path to merge.

### Brief observations (for reference if ported forward)

The change itself is straightforward and correct in intent -- adding `"space"` as a thousands separator option is a valid functional improvement, since space-separated thousands (e.g., `1 234 567,89`) are common in many European and international number formats (ISO 31-0, SI standard).

A few points to address if porting to a supported branch:

1. **Missing version bump in `__manifest__.py`**: Adding a new selection value to a field is a functional change that should be accompanied by a version bump (e.g., `14.0.3.2.1` -> `14.0.3.3.0` or equivalent for the target branch).

2. **Missing migration script**: Existing databases upgrading the module will need the new selection value registered. Depending on the Odoo version, a `pre-migrate` or `post-migrate` script may be needed to handle the selection field update cleanly.

3. **No test coverage**: The module has a test suite (`tests/test_account_statement_import_txt_xlsx.py`), but this PR does not add a test case exercising the `"space"` thousands separator. A test importing a CSV/TXT with space-separated thousands would strengthen the contribution.

4. **i18n/pot file not regenerated**: The new selection label `"space ( )"` should appear in updated `.pot`/`.po` files for translators.

5. **Order in the separators dict**: The `"space"` entry is placed after `"none"` in the `_get_float_separators` method, which breaks the visual correspondence with the field's `selection` list (where `"space"` comes before `"none"`). Minor, but keeping consistent ordering aids readability.

Overall, this is a useful improvement. I would encourage porting it to 16.0+ and addressing the above points.

---

PR #843, opened 2025-10-09
[14.0][FIX] account_statement_import_online_paypal: error when no transactions returned

matthieusaison:
If you call transaction for an empty transaction days, payload returns no fields 'transaction_details' and raise an error.

### DIFF
--- a/account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
+++ b/account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
@@ -416,18 +416,22 @@ class OnlineBankStatementProviderPayPal(models.Model):
                 data = self.with_context(
                     invalid_data_workaround=invalid_data_workaround,
                 )._paypal_retrieve(url, token)
-                interval_transactions = map(
-                    lambda transaction: self._paypal_preparse_transaction(transaction),
-                    data["transaction_details"],
-                )
-                transactions += list(
-                    filter(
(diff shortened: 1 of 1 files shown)


DeeEmm:
Ahh just stumbled across the same issue and came here to post a fix. Slightly different solution but essentially the same.

```
                data = self.with_context(
                    invalid_data_workaround=invalid_data_workaround,
                )._paypal_retrieve(url, token)
                
                
                transaction_details = data.get("transaction_details", []) 
                
                #    If the key is missing, it will use an empty list instead.
                transaction_details = data.get("transaction_details", [])
                
                interval_transactions = map(
                    lambda transaction: self._paypal_preparse_transaction(transaction),
                    transaction_details,
                )
               
                transactions += list(
                    filter(
                        lambda transaction: interval_start
                        <= self._paypal_get_transaction_date(transaction)
                        < interval_end,
                        interval_transactions,
                    )
                )
```

kwoychesko:
Great job and thank you @matthieusaison -- this same fix worked for me in v18. It would be great to get this PR approved, then the fix can be applied in other versions (which I'm happy to help with).

alexey-pelykh (approved):
Good fix — guarding against empty `transaction_details` prevents crashes when PayPal returns no transactions for a time interval. Clean and targeted. LGTM

tarteo (approved):
Yes, this issue is in 18.0 too

PR #846, opened 2025-10-16
[18.0][FIX] account_statement_import_online_stripe: store valid json in raw_data

Dirbaio:
Storing `tx` directly means `raw_data` would end up containing the stringified python dict, which
is not actually JSON and is harder to parse.

### DIFF
--- a/account_statement_import_online_stripe/models/online_bank_statement_provider_stripe.py
+++ b/account_statement_import_online_stripe/models/online_bank_statement_provider_stripe.py
@@ -77,7 +77,7 @@ class OnlineBankStatementProviderStripe(models.Model):
                     "amount": float(tx["amount"]) / (10**currency.decimal_places),
                     "date": datetime.fromtimestamp(tx["created"]),
                     "unique_import_id": tx["id"],
-                    "raw_data": tx,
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
~~`raw_data` is a `Text` field, not `jsonb`, so doing the dump is only going to encode the information, but not to be a real JSON, so it doesn't serve for Odoo purposes.~~

EDIT: I have read it the contrary.

pedrobaeza (approved):

Dirbaio:
I agree storing JSON in a text field is not ideal, but it's better than storing a stringified dict.  A stringified dict can be parsed from Python with `ast.literal_eval` or similar even though it's a bit hacky, but it's impossible to parse in other programming languages. JSON is more universal.

Also, other modules are already storing json in text: https://github.com/OCA/bank-statement-import/blob/65682ce7fa7eb4a057d25b4ac28ed2dcfb5d9398/account_statement_import_online_ponto/models/online_bank_statement_provider_ponto.py#L197

pedrobaeza:
Fast-tracking this:

/ocabot merge patch

PR #847, opened 2025-10-16
[17.0][FIX] account_statement_import_online_stripe: store valid json in raw_data.

Dirbaio:
Storing `tx` directly means `raw_data` would end up containing the stringified python dict, which
is not actually JSON and is harder to parse.

backport of #846 to 17.0


pedrobaeza (approved):
/ocabot merge patch

PR #849, opened 2025-10-24
[18.0][MIG] account statement import paypal: Migration to 18.0

javierjcf:


account_statement_import_paypal/models/account_statement_import.py
```diff
@@ -0,0 +1,39 @@
...
+        comodel_name="account.statement.import.paypal.mapping",
+    )
+
+    def _parse_file(self, data_file):
+        self.ensure_one()
+        if self.paypal_mapping_id:
+            try:
+                Parser = self.env["account.statement.import.paypal.parser"]
+                logging.warning(
+                    Parser.parse(
```
alexey-pelykh, line 28: This `logging.warning(Parser.parse(...))` calls `Parser.parse()` a second time just to log the result. Remove this — the actual return is on the next line.

alexey-pelykh (commented):
Thanks for working on this migration! There's a debug leftover in `account_statement_import.py` — `logging.warning(Parser.parse(...))` on line 27 calls the parser a second time for nothing. This also means every import runs the full parse twice. Please remove that call.

Also, the test class still uses `setUp` instead of `setUpClass`, which is the OCA convention for 18.0. CI is currently failing on both pre-commit and tests, so those need to be addressed too.

PR #850, opened 2025-10-28
[14.0][BACKPORT] account_statement_import_online_ponto (from 16.0)

metaminux:
In 14.0, account_statement_import_online_ponto have bugs.

Most important bugs are :

- Ponto API is providing transactions from the last created backwards (so comparing to "Value Date" or "Execution Date" to end retrieval can lead to transaction loss)
- Also, taking "Value Date" or "Execution Date" as statement "date" field is leading to filtering out transactions not imported yet

Those bugs have been fixed in 16.0, mainly in the migration commit from 15.0 to 16.0

I suggest to backport changes from 16.0 to get same behaviour between versions...


metaminux:
I had to take some changes to account_statement_import_online (from #548)

Most of them are tests rewrite though...

I had to adapt account_statement_import_online_ponto_ing too (not ported to 16.0)

Can anyone review this ? @alexey-pelykh ? @NL66278 ? @astirpe ? @vaab ? @victoralmau ?

CC @sebastienbeau

NL66278 (approved):
👍 LGTM

alexey-pelykh (commented):
## OCA PR Review: `[14.0][BACKPORT] account_statement_import_online_ponto (from 16.0)`

### EOL Branch Notice

**This PR targets the `14.0` branch, which has reached End of Life.** OCA maintenance for 14.0 has ended, so this PR is unlikely to be merged. If the fixes are needed, consider targeting a supported branch (16.0+) or maintaining the backport in a private fork.

That said, below is a technical review of the changes for informational purposes.

---

### Summary

This is a large backport (11 files, +487/-232) that brings Ponto provider improvements from 16.0 to 14.0. The core changes address real bugs:
- Ponto API returns transactions from latest backwards, so date-based comparisons could lose transactions
- Using value/execution date as statement date could filter out un-imported transactions
- Adds `ponto_last_identifier` for scheduled run continuity

### Correctness

1. **`_ponto_retrieve_data` stop-condition logic**: The dual-mode (scheduled vs wizard) retrieval logic is well-structured. Scheduled runs stop on `ponto_last_identifier` match or `date_stop`, while wizard runs filter by date range. This correctly addresses the original bug where date-based stopping could miss transactions.

2. **`_update_providers` search-or-create pattern**: The new `_update_providers()` method searches for an existing provider before creating a new one. This is a good improvement over the old behavior that unconditionally unlinked and recreated.

3. **`_update_vals` clears provider on source change**: When `bank_statements_source` is changed away from `"online"`, both the selection field and the Many2one are cleared. This is correct.

4. **Removed `ondelete="restrict"`**: The old code prevented deleting a provider linked to a journal. Removing this changes data integrity behavior -- providers can now be deleted without constraint. This is intentional per the 16.0 design but should be noted.

5. **Deleted `test_provider_unlink_restricted` and `test_source_change_cleanup`**: These tests are correctly removed since they tested behavior that no longer exists (restrict constraint and old cleanup logic). However, no replacement tests cover the new `_update_providers` search-or-create logic or `_update_vals` clearing behavior, which is a gap.

### Odoo Best Practices

6. **`@api.model_create_multi`**: Good -- the `create` method is updated to use `@api.model_create_multi` with `vals_list`, which is the correct pattern for Odoo 14.0+.

7. **Class rename `OnlineBankStatementProviderPonto` -> `OnlineBankStatementProvider`**: In the ponto module, the class is renamed to match the model name convention. This is fine since `_inherit` is what matters, not the Python class name. However, note that `account_statement_import_online_ponto_ing` still uses the old class name `OnlineBankStatementProviderPonto` -- this inconsistency is harmless but worth noting.

8. **`_selection_service` method on journal**: The new `_selection_service` delegates to the provider model's `_selection_service`. The old method was `_selection_online_bank_statement_provider`. This rename may break other OCA modules that depend on the old method name.

9. **Removed `values_online_bank_statement_provider` and `widget="dynamic_dropdown"`**: The view no longer uses the dynamic dropdown widget. This simplifies the UI but loses the ability to filter "dummy" provider to only debug mode users.

### Security

10. **Added `groups="account.group_account_user"` on pull button**: Good security improvement -- the "Pull Online Bank Statement" button is now restricted to account users.

11. **Timeout added to `requests.post`/`requests.get`**: Good -- all HTTP calls now have explicit timeouts (`timeout=60` or `timeout=(60, 300)`), preventing indefinite hangs.

### Performance

12. **`_pull` override instead of `_obtain_statement_data`**: The new approach overrides `_pull` at a higher level, filtering ponto providers and delegating non-ponto ones to `super()._pull()`. This is a more efficient dispatch pattern.

### i18n

13. **`_()` used on logger messages with `.format()`**: Several places use patterns like `_("...").format(...)`. This is an anti-pattern in Odoo -- `_()` should receive the full untranslated string, and `%` or `format()` should not be called on the result. The translated string might have different placeholder positions. The correct pattern in Odoo 14.0 is either `_("message %s") % value` or using the lazy `%` operator. Examples:
    - `ponto_interface.py`: `_("Ponto : wrong configuration, account {account} not found in {data}").format(...)` 
    - `ponto_interface.py`: `_("Server returned status code {response_code}: {response_text}").format(...)`

14. **Logger messages with `_()` wrapping**: Debug/info logger messages should generally not be wrapped in `_()` for translation. Log messages are for developers/admins and translating them makes log analysis harder. This is pre-existing in the codebase though, so not introduced by this PR.

### Test Coverage

15. **Expanded test suite**: Good -- tests now cover execution_date vs value_date selection, multi-page transaction retrieval, scheduled pulls, and scheduled pulls from a stored identifier. This is much more thorough than the original 3 tests.

16. **CoA loading in setUp**: The test now loads a Chart of Accounts if none exists. This improves test portability across different Odoo installations.

17. **Missing tests for base module changes**: The `account_statement_import_online` base module changes (new `_update_providers`, `_update_vals`, `action_online_bank_statements_pull_wizard`, `_selection_service`) lack dedicated test coverage. The deleted `test_account_journal.py` is not replaced.

### Manifest / Documentation

18. **No `__manifest__.py` version bump**: For a backport of this magnitude, the module version should be bumped to signal the behavioral change to users who upgrade.

19. **No migration script**: The addition of `ponto_last_identifier` field and removal of `ondelete="restrict"` may need a pre/post-migration script for existing installations. Existing provider records will have `ponto_last_identifier = False`, which means the first scheduled run will pull all transactions back to `date_stop` -- this could result in duplicates if `unique_import_id` handling is not robust.

### Minor

20. **Set to list change**: `{"description", "counterpartName", "counterpartReference"}` changed to `["description", "counterpartName", "counterpartReference"]`. Good -- sets have non-deterministic iteration order, which could cause inconsistent `ref` construction across Python versions.

21. **`# pylint: disable=missing-return`**: The `_pull` method has a pylint disable for missing return. This is acceptable since the parent class method also has no meaningful return value.

22. **Typos**: "wether" appears in several places (should be "whether"). Pre-existing in some cases but new instances added.

---

### Verdict

The changes are technically sound and address real bugs in the Ponto integration. The backport from 16.0 is well-executed. However, **this targets an EOL branch (14.0)** and is unlikely to be merged by OCA maintainers.

If this were targeting a supported branch, the main items to address would be:
- i18n anti-pattern with `_().format()` (item 13)
- Missing version bump in `__manifest__.py` (item 18)
- Consider migration script for `ponto_last_identifier` (item 19)
- Add tests for base module changes (item 17)

metaminux:
Hi @alexey-pelykh,

Is it possible to consider merging this PR for homogeneity between versions ?

PR #835, opened 2025-09-25
[18.0][IMP] account_statement_import_online: Avoid unnecessary log messages

victoralmau:
FWP from 17.0: https://github.com/OCA/bank-statement-import/pull/833

Avoid unnecessary log messages

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa

PR #853, opened 2025-11-06
[18.0][FIX] account_statement_import_online: fix excessive writes

StefanRijnhart:
* Don't update journals unless a relevant field is updated on the online configuration.
* Don't update next_run on the online configuration during every iteration of the interval. This could be many iterations if the configuration is not actually valid and the interval is very short. In the past, this also led to excessive logging which was adressed in https://github.com/OCA/bank-statement-import/pull/835.

A similar fix for 16 can be found in https://github.com/OCA/bank-statement-import/pull/832


pedrobaeza (approved):
/ocabot merge patch

PR #856, opened 2025-11-13
[18.0] [MIG] account_statement_import_online_wise

Kiplangatdan:


account_statement_import_online_wise/models/online_bank_statement_provider_wise.py
```diff
@@ -69,18 +66,18 @@ def values_transferwise_profile(self):
...
     def _get_available_services(self):
         return super()._get_available_services() + [
-            ("transferwise", "Wise.com (TransferWise.com)"),
+            ("wise", "Wise.com"),
         ]
 
     def _obtain_statement_data(self, date_since, date_until):
         self.ensure_one()
-        if self.service != "transferwise":
+        if self.service != "wise":
```
thomaspaulb: Is it really necessary to change this? Nobody will see the underlying key anyway. If we just update the description, it saves us from having to write a migration script.

account_statement_import_online_wise/data/neutralize.sql
```diff
@@ -0,0 +1,5 @@
+-- disable wise
+UPDATE online_bank_statement_provider
+SET certificate_private_key = 'dummy',
+    certificate_public_key = 'dummy'
+WHERE service='wise';
```
thomaspaulb: (if we keep it transferwise, will also need to be kept here, and in the tests)

thomaspaulb (approved):
Code review. Functional review upcoming after go-live with this

kobros-tech (approved):
It is fine by implementation

kobros-tech:
@Kiplangatdan 

I need to know the reason for deleting this test file?

account_statement_import_online_wise/tests/test_account_bank_statement_import_online_transferwise.py

We usually not update tests or change or delete but under obvious reasons like changes in odoo base code or similar.

alexey-pelykh (approved):
Thanks for picking up the Wise migration. As a co-author of the original module — the SCA handling with `X-2FA-Approval` + private key signing is preserved correctly, and the `_obtain_statement_data` pagination logic looks right.

LGTM.

Issue #857, opened 2025-11-14
Paypal Online Payment Statement Import raises error when PayPal returns datetime.datetime object instead of ISO string


sefirosweb:
## Module
`payment_paypal_online_payment_statement_import`  
(OCA module for PayPal transaction import — Odoo 16)

---

## Describe the bug
When importing PayPal transactions, the module attempts to store the original PayPal payload into the `raw_data` field of the bank statement line.

However, if PayPal returns the field `transaction_updated_date` as a **Python `datetime.datetime` object** instead of a string, Odoo fails to serialize this value when saving it into the JSON field `raw_data`, causing an exception.

The problem occurs **specifically when the module tries to write the raw PayPal payload into `raw_data`**, because Python datetime objects are **not JSON serializable**, and the module does not convert them to a proper ISO string.

As a result, the transaction cannot be imported.

---

### **Affected versions**
- Odoo **16** (OCA)
- Module: `payment_paypal_online_payment_statement_import`
- Environment: **Docker**
- Python: **3.9.2**

### **Steps**
1. Configure PayPal sync using the module.  
2. Receive a PayPal payload where the field `transaction_updated_date` is returned as a Python `datetime.datetime` object.  
3. The module attempts to save the payload into `raw_data`.  
4. Odoo raises a serialization error and the import stops.

### **Problematic example payload (ofuscated)**  
⚠️ Sensitive data masked, but structure and the problematic `datetime` object are preserved.

```python
{
    'paypal_account_id': 'XXXXXXXXXXXXXX',
    'transaction_id': '11111111111111111',
    'paypal_reference_id': 'YYYYYYYYYYYYYYYYY',
    'paypal_reference_id_type': 'TXN',
    'transaction_event_code': 'T0006',
    'transaction_initiation_date': '2025-10-30T22:28:36Z',
    'transaction_updated_date': datetime.datetime(2025, 10, 30, 22, 28, 36),  # <-- Problem (not serializable)
    'transaction_amount': {
        'currency_code': 'EUR',
        'value': '216.98'
    },
    'fee_amount': {
        'currency_code': 'EUR',
        'value': '-6.64'
    },
    'sales_tax_amount': {
        'currency_code': 'EUR',
        'value': '34.71'
    },
    'shipping_amount': {
        'currency_code': 'EUR',
        'value': '16.99'
    },
    'transaction_status': 'S',
[...]

PR #858, opened 2025-11-14
[FIX] account_statement_import_online_paypal: Fix error importing Pay…

sefirosweb:
[FIX] account_statement_import_online_paypal: Fix error importing PayPal transactions with null fee_amount (#857)

This commit solves issue #857 (https://github.com/OCA/bank-statement-import/issues/857), which caused an error when importing PayPal transactions with a null fee_amount.

Additionally, it fixes the way raw_data is stored in the database to ensure it can be correctly retrieved and processed later.

### DIFF
--- a/account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
+++ b/account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
@@ -438,7 +438,9 @@ class OnlineBankStatementProviderPayPal(models.Model):
     def _paypal_get_transaction_date(self, transaction):
         # NOTE: CSV reports from PayPal use this date, search as well
         return (
-            dateutil.parser.parse(transaction["transaction_info"]["transaction_updated_date"])
+            dateutil.parser.parse(
+                transaction["transaction_info"]["transaction_updated_date"]
+            )
             .astimezone(pytz.utc)
             .replace(tzinfo=None)
         )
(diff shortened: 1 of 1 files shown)


sefirosweb:
This PR removes the function `_paypal_preparse_transaction` because it has 
become obsolete. The method is no longer used anywhere in the PayPal provider’s 
functional workflow, and its original purpose has been replaced by newer logic 
inside `_paypal_get_transaction_date`.

The only remaining reference to this method is in a very specific unit test, 
which calls `_paypal_preparse_transaction` directly. Since the function no 
longer plays any role in the module’s behavior, this test now fails simply 
because it is asserting behavior that no longer exists in the codebase.

In other words: the failure does not come from a regression, but from a test 
that targets functionality that is no longer part of the module.

I am opening this PR so the maintainers can review whether this test still has 
any reason to exist, or if it should be updated/removed to align with the 
current implementation of the PayPal provider.

Please confirm whether the test should be kept, adapted, or removed given the 
current state of the module.

best,

alexey-pelykh (commented):
Good refactoring — consolidating the date parsing into `_paypal_get_transaction_date` and dropping `_paypal_preparse_transaction` makes the flow cleaner. One thing to note: `raw_data` changes from a dict to `json.dumps(transaction)` (string), which is a behavior change for anything reading `raw_data` downstream. Tests are currently failing — once CI is green this will be easier to evaluate fully.

PR #868, opened 2025-12-16
[18.0][MIG] account_statement_import_sheet_file: Migration to 18.0

pedrobaeza:
Apart from the usual migration procedure, it includes in a separate commit a whole refactor of the test suite for making it usable, as it was victim of the copy/paste syndrome when expanding more test use cases. As this is not touching the features of the module, it's OK to include it here as part of the migration PR.

@Tecnativa TT55498


jbaudoux (approved):

account_statement_import_sheet_file/models/account_statement_import.py
```diff
@@ -0,0 +1,51 @@
...
+            try:
+                Parser = self.env["account.statement.import.sheet.parser"]
+                return Parser.parse(
+                    data_file, self.sheet_mapping_id, self.statement_filename
+                )
+            except BaseException as exc:
+                if self.env.context.get("account_statement_import_sheet_file_test"):
+                    raise
+                _logger.warning("Sheet parser error", exc_info=True)
+                raise UserError(_("Bad file/mapping: ") + str(exc)) from exc
```
carlos-lopez-tecnativa: Non-blocking comment, but if you can update the translation method in all occurrences, that would be great.

```suggestion
                raise UserError(self.env._("Bad file/mapping: ") + str(exc)) from exc
```

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -0,0 +1,497 @@
...
+        lines = list(sorted(lines, key=lambda line: line["timestamp"]))
+        first_line = lines[0]
+        last_line = lines[-1]
+        data = {
+            "date": first_line["timestamp"].date(),
+            "name": _("%(code)s: %(filename)s")
+            % {
+                "code": journal.code,
+                "filename": path.basename(filename),
+            },
```
carlos-lopez-tecnativa: Instead of passing the arguments as format string, pass them as parameters.
```suggestion
            "name": self.env._("%(code)s: %(filename)s", code=journal.code, filename=path.basename(filename))
```

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -0,0 +1,497 @@
...
+                f"{transaction_id}-{int(timestamp.timestamp())}"
+            )
+
+        transaction["payment_ref"] = description or _("N/A")
+        if reference:
+            transaction["ref"] = reference
+
+        note = ""
+        if bank_name:
+            note += _("Bank: %s; ") % (bank_name,)
```
carlos-lopez-tecnativa:
```suggestion
            note += self.env._("Bank: %s; ", bank_name)
```

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
```diff
@@ -0,0 +1,492 @@
...
+        wizard = self._get_import_wizard("fixtures/sample_statement_en.csv")
+        wizard.import_file_button()
+        statement = self.AccountBankStatement.search(self.statement_domain)
+        self.assertEqual(len(statement), 1)
+        self.assertEqual(len(statement.line_ids), 2)
+
+    def test_import_empty_csv_file(self):
+        wizard = self._get_import_wizard("fixtures/empty_statement_en.csv")
+        with self.assertRaises(UserError):
+            wizard.import_file_button()
```
carlos-lopez-tecnativa, line 90: It could be useful to use `assertRaisesRegex` instead of `assertRaises` to specify the expected message and be more precise.
```suggestion
        with self.assertRaisesRegex(UserError, expected_msj):
            wizard.import_file_button()
```

carlos-lopez-tecnativa (approved):
LGTM. Just a few minor non-blocking comments, but if you could address them, that would be great. Thanks for the improvements done in the tests.

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
pedrobaeza: Although it can prevent to catch a non-desired error, the scope of this call is very limited, and I don't think it's very useful to restrict to a specific error message. On contrary, putting that message and if later the message is changed, refactored to use a general one, etc, the test won't be resilient and you may not know exactly where the problem comes (you get that the raise is not triggered, but it is, but with another message).

Thus, I prefer in general to not do this.

pedrobaeza:
The rest of the suggestions have been included, so merging finally the migration of this module!

/ocabot merge nobump

PR #871, opened 2025-12-19
[18.0][ADD] statement_import_sheet_html_file

Daemo00:
Allow to import XLS[X] files that actually are HTML.


account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -1,6 +1,7 @@
-# Copyright 2019 ForgeFlow, S.L.
+# Pbclevtug 2019 SbetrSybj, F.Y.
```
alexey-pelykh: This line got garbled -- `Pbclevtug` is ROT13 of `Copyright`. Likely a pre-commit hook artifact. Needs to be restored to `# Copyright 2019 ForgeFlow, S.L.`

alexey-pelykh (commented):
Interesting edge case -- some bank exports do produce HTML tables disguised as .xlsx files. The concept is solid.

However, CI is fully failing and there's a garbled copyright header on line 1 of the parser file (`Pbclevtug` instead of `Copyright` -- looks like ROT13 got applied). Also `is_HTML` should follow Python naming as `is_html`. Worth fixing those before this can move forward.

account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
Daemo00: I don't know how that happened 😆 thanks for catching it!

Daemo00 (commented):
> Also `is_HTML` should follow Python naming as `is_html`.

Done this too, it was all uppercase because it is an acronym, but apparently acronyms should be lowercase too in variable names

pedrobaeza (changes requested):
For me, this seems a very specific case where the HTML content is then a pure table, but I don't think this is a generic case to include more code in this already bloated module. Or am I missing something?

Daemo00:
> For me, this seems a very specific case where the HTML content is then a pure table, but I don't think this is a generic case to include more code in this already bloated module. Or am I missing something?

Thanks for having a look!
I see this module is bloated, but I also want to share this improvement with others that might need it.
What alternative do I have? Should I add a hook in this module and import the new file in a new `statement_import_sheet_html_file`? That seems uglier to me.

This module is made for importing XLSX files, and my XLSX files have this particular structure (they are actually HTML) so I improved this module to be able to import them.
I don't know how specific this case is but I can assure you I'm not the only customer of my bank 😆

pedrobaeza:
Yes, extension module should be the solution, and in 19 we should have a base module with the configuration, and extensions for each format (CSV, Excel...).

Daemo00:
> Yes, extension module should be the solution, and in 19 we should have a base module with the configuration, and extensions for each format (CSV, Excel...).

Reimplemented as an extension, please let me know what you think

christian-ramos-tecnativa (commented):
IMO now that we want the module to be extensible it could be the moment to remove the try and repeat with each parser. Maybe adding a tuple like `(parser, [files extensions])`  so we can choose the parser based in the file extension. WDYT

Daemo00:
> IMO now that we want the module to be extensible it could be the moment to remove the try and repeat with each parser. Maybe adding a tuple like `(parser, [files extensions])` so we can choose the parser based in the file extension. WDYT

Hi, thanks for having a look!
Here I was only trying to enable the parsing of XLSX files that are actually HTML files, so the XLSX file extension could be parsed by both the `html_parser` and the `xlrd_parser` and we'd be back to a `try` loop: so I don't see much improvement in associating a parser to a file extension.

I'd also like to keep this PR as focused as possible on the described improvement, without adding too much refactoring.

alexey-pelykh (commented):
Thanks for the updates, @Daemo00 — re-reviewed against the current diff.

The three points from my earlier pass are all resolved:

- CI is green now (pre-commit, tests on OCB and Odoo, codecov all passing).
- The garbled `# Pbclevtug 2019 ...` line is restored to `# Copyright 2019 ForgeFlow, S.L.`.
- `is_HTML` → `is_html`.

I like where you landed: exposing `_get_data_parsers()` as an extension hook in the base module and shipping the HTML handling as a separate `account_statement_import_sheet_html_file` module. That keeps the HTML-specific logic out of the base parser and matches the extension-module approach discussed in the thread.

One small thing in `_parse_data_html`:

```python
wr.writerow([col.text for col in rows[0].xpath("//th")])
wr.writerows([[col.text for col in row.xpath(".//td")] for row in rows[1:]])
```

`rows[0].xpath("//th")` is an absolute path, so it matches every `<th>` in the whole parsed document rather than just the header row — note the next line correctly uses the row-relative `.//td`. For a single-table export the result is identical, but it would pull in stray `<th>` from anywhere if the input ever contains more than one table. Worth aligning to `rows[0].xpath(".//th")` for consistency.

Non-blocking from my side. I'll leave the merge call to @pedrobaeza, since his change request is what drove the restructure.

Daemo00:
> One small thing in `_parse_data_html`:
> 
> ```python
> wr.writerow([col.text for col in rows[0].xpath("//th")])
> wr.writerows([[col.text for col in row.xpath(".//td")] for row in rows[1:]])
> ```
> 
> `rows[0].xpath("//th")` is an absolute path, so it matches every `<th>` in the whole parsed document rather than just the header row — note the next line correctly uses the row-relative `.//td`. For a single-table export the result is identical, but it would pull in stray `<th>` from anywhere if the input ever contains more than one table. Worth aligning to `rows[0].xpath(".//th")` for consistency.
> 
> Non-blocking from my side. I'll leave the merge call to @pedrobaeza, since his change request is what drove the restructure.
> 

Thanks for the review!
I added `.//th` instead of `.//td`, please update your review

Daemo00:
Conflicts resolved

Daemo00:
@christian-ramos-tecnativa thanks for the review!
Please check my reply https://github.com/OCA/bank-statement-import/pull/871#issuecomment-4194704364 and let me know what you think

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
```diff
@@ -4,112 +4,45 @@
...
-        with open(path.join(path.dirname(__file__), filename), mode) as file:
-            data = file.read()
-            if encoding:
-                data = data.encode(encoding)
-            return b64encode(data)
 
+class TestAccountStatementImportSheetFile(Common):
     def test_import_csv_file(self):
-        wizard = self._get_import_wizard("fixtures/sample_statement_en.csv")
+        wizard = self._get_import_wizard("sample_statement_en.csv")
```
pedrobaeza: Please reduce the diff not modifying this (in all the places)

pedrobaeza (commented):
You can check how it has been done in 19, which we separate the parsers, to make it similar, although I won't force you to do it due all the work you have already put. Just check my comment inline.

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
Daemo00: You are right this is just noise in a commit that is doing something else, but I think it's good to remove this repetition so I have moved it to a dedicated commit.

pedrobaeza: I think on contrary, as you are forcing a fixed path that not always will be the proper one, and following Python zen, better explicit than implicit, so I would ask to leave it as it was.

Daemo00: In this case I wanted to assign the actual filename to a parameter called `filename` instead of the path `"fixtures/some_file.csv"`, and avoid repetition, I think that would agree with Python zen too.
I understand that it means forcing the developer to use the `fixtures` folder, but that's not much different than using `tests` for test cases, and the name of the folder is generic enough so that would do no harm.

Anyway, I think the functionality I've been wanting to contribute for a year and a half (since https://github.com/OCA/bank-statement-import/pull/763) is more important than this quibble so I have restored the previous implementation.

pedrobaeza (approved):
/ocabot merge patch

Daemo00:
> You can check how it has been done in 19, which we separate the parsers, to make it similar, although I won't force you to do it due all the work you have already put.

Thanks for improving the module in `19.0`!
Is it https://github.com/OCA/bank-statement-import/pull/958? Are you planning to backport that to `18.0`?

pedrobaeza:
Yes, that one, but no backport is possible, as it affects stability, splitting a module into several of them.

PR #873, opened 2025-12-27
[18.0][IMP] account_statement_sheet_file: replace xlrd for openpyxl

andyengit:
Added openpyxl as a dependency for XLSX file handling and updated the parser to utilize openpyxl instead of xlrd.

https://github.com/pandas-dev/pandas/issues/28547


account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -149,17 +154,15 @@ def _get_column_names(self):
...
-                ),
+            workbook = openpyxl.load_workbook(
+                filename=BytesIO(data_file),
+                data_only=True,
             )
             csv_or_xlsx = (
                 workbook,
-                workbook.sheet_by_index(0),
+                workbook.worksheets[0],
             )
```
alexey-pelykh, line 164: `(BadZipFile, Exception)` is redundant -- `Exception` is the parent class and already covers `BadZipFile`. Just `except Exception:` works.

alexey-pelykh (commented):
Good direction -- xlrd hasn't seen active maintenance in a while and openpyxl is the standard choice now. The `_process_row_values()` extraction is a nice cleanup too.

Small thing: `except (BadZipFile, Exception)` is redundant since `Exception` already catches `BadZipFile`. Can be simplified to just `except Exception`.

andyengit:
@alexey-pelykh i make the changes

lk-eska:
This PR become more relevant as xlrd 2.0 doesn't support xlsx....

andyengit:
@pedrobaeza can u check this pr?

pedrobaeza:
Please check instead #958

andyengit:
@pedrobaeza  In this case, is it advisable to wait until the solution is available for version 19 and then port it to 18.0?

I've been using this program since last year.

pedrobaeza:
Sorry, I didn't see that this is for 18. Then no refactor is expected here, but the problem is that you can't change dependencies for this. Your problem is that you are using Python 3.12, while the expected Python version everywhere for 18 is 3.11. You should stick to that version for not having problems IMO. If not, you should do an extra module for using the XLSX with openpyxl.

lk-eska:
why don't we just accept this for 3.12 users

andyengit:
@pedrobaeza  I'll keep that in mind. I'm new to working with OCA. Do you recommend taking the extraction request currently being made for version 19.0 and migrating it to 18.0? I see you're creating a module specifically for xls and xlsx.

Or would there be no problem with how I handle this solution in 18.0?

pedrobaeza:
Because it will break all the existing installations (not deployed with PyPi), changing the dependencies of an stable module.

alexey-pelykh (commented):
Thanks @andyengit — re-reviewed against the current diff.

My earlier nit is resolved: the exception handling is now just `except Exception:`, no more redundant `(BadZipFile, Exception)`. 👍

The underlying need is real — with xlrd 2.0 no longer reading `.xlsx` at all, `openpyxl` is the right engine going forward. But as a maintainer of `account_statement_import_sheet_file`, I share @pedrobaeza's concern about the approach: swapping the dependency (`xlrd` → `openpyxl`) inside this *stable* 18.0 module changes its runtime deps and breaks existing installations that don't resolve deps from PyPI. On a stable branch we can't do that in place.

The non-breaking path for 18.0 is a separate module that layers openpyxl-based parsing on top of this one — the same pattern #871 uses to add HTML support — rather than replacing the parser here. (The 19.0 refactor already merged in #958.)

Two practical blockers on the current state as well:

- `codecov/patch` and `codecov/project` are red — the new `_process_row_values()` and the openpyxl branch in `_parse_rows` aren't covered (no tests were added in the PR).
- The branch is conflicting with `18.0` and needs a rebase.

The openpyxl parsing itself looks reasonable — it's really the packaging for a stable branch that needs to change. Happy to review a standalone-module version.

PR #876, opened 2025-12-30
[FIX] account_statement_import_sheet_file: support for reading XLSX files in addition to XLS and CSV formats

rov-adhoc:


alexey-pelykh (commented):
Thanks for tackling XLSX support with openpyxl. One thought -- there's also #873 which takes a cleaner approach by fully replacing xlrd with openpyxl rather than maintaining a dual-path with `hasattr()` checks throughout. The dual approach adds maintenance burden since every future change to sheet parsing needs both xlrd and openpyxl code paths. Might be worth coordinating with #873 on the direction.

PR #879, opened 2026-01-02
[FIX] account_statement_import_sheet_file: str value for sub separator

cav-adhoc:
The `_parse_decimal` method was passing an integer value to `re.sub()` in the separator case, causing "expected string or bytes-like object, got 'int'" error. Now the value is converted to string before `re.sub()` call. The method still returns a decimal value as expected, so behavior remains unchanged.

## Update

Following @pedrobaeza's and @alexey-pelykh's review, the numeric case is now handled directly in the early `isinstance` branch instead of stringifying and re-parsing it through `re.sub()`.

- `int` values are now returned through the numeric branch (`isinstance(value, (float, int))`), as `float`, so they never reach the separator parsing.
- Reverted the `str(value)` cast in `re.sub()`: only actual strings reach that path now.

This avoids a lossy round-trip: an integer parsed as a string would be corrupted when no decimal separator is configured (e.g. `1234` -> `12.34`).

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
@@ -481,8 +481,8 @@ class AccountStatementImportSheetParser(models.TransientModel):
     def _parse_decimal(self, value, mapping):
         if isinstance(value, Decimal):
             return float(value)
-        elif isinstance(value, float):
-            return value
+        elif isinstance(value, float | int):
+            return float(value)
         thousands, decimal = mapping._get_float_separators()
         # Remove all characters except digits, thousands separator,
         # decimal separator, and signs
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
@@ -488,7 +488,9 @@ class AccountStatementImportSheetParser(models.TransientModel):
         # decimal separator, and signs
         value = (
             re.sub(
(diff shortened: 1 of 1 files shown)


alexey-pelykh (approved):
Thanks for the fix! Clean and focused — handles the case where XLSX cells return numeric types. LGTM

pedrobaeza (changes requested):
I think it should be included in the previous `elif`:

```
        elif isinstance(value, float | int):
            return value

alexey-pelykh (commented):
Re-reviewing this — I should retract my earlier approval. @pedrobaeza is right: rather than `str(value)` and then regex-stripping a numeric cell in `_parse_decimal`, the `float`/`int` case is better returned directly through the existing `isinstance(value, float | int)` branch above, avoiding a lossy round-trip through string parsing (locale thousands/decimal handling can misfire on a stringified float). My earlier "LGTM" missed that. As a maintainer of this module, agreeing with the requested change — @cav-adhoc, could you move the numeric handling up into that `elif`?

cav-adhoc:
> Re-reviewing this — I should retract my earlier approval. @pedrobaeza is right: rather than `str(value)` and then regex-stripping a numeric cell in `_parse_decimal`, the `float`/`int` case is better returned directly through the existing `isinstance(value, float | int)` branch above, avoiding a lossy round-trip through string parsing (locale thousands/decimal handling can misfire on a stringified float). My earlier "LGTM" missed that. As a maintainer of this module, agreeing with the requested change — @cav-adhoc, could you move the numeric handling up into that `elif`?
> 

@alexey-pelykh Done!

PR #880, opened 2026-01-02
[18.0][IMP] import xml file individually from zip

davidwul:
with this PR:
1. you can import a zip file containing camt files for different bank accounts. 
2. each xml files are attached to its proper bankstatement.
3. empty camt statements (with no transaction) are ignore and don't trigger any error

I was not able to implement this in the parse_file method which can be an issue if other parser allow to import zip files. 
I you have any idea to avoid that?
But for now, I don't think that's the case.


alexey-pelykh (commented):
Thanks for the contribution! Processing individual XML files from a ZIP is a useful feature.

A few things:

- f-strings in `_logger.exception()` — OCA/pylint-odoo convention is `%s` formatting for lazy evaluation
- Bare `except Exception` is quite broad — might want to narrow it down
- Tests are currently failing (both OCB and Odoo), so hard to evaluate the full behavior

Happy to re-review once CI is green.

davidwul:
Thanks for the review, CI is green now @alexey-pelykh

alexey-pelykh (approved):
All prior feedback addressed — logging now uses lazy %s formatting, exception handling is properly scoped to (ValueError, UserError) and BadZipFile, and CI is green. LGTM.

ecino (approved):

ecino:
Can we merge this one?

Issue #885, opened 2026-01-10
Add DtTm to parse date in CAMT.53 parser


javiermatos:
I am using CAMT importer for loading Wise extract data. The problem is Wise extract entries use dates in a different node so it raises an error when processing its extract files, which are correct according to CAMT.53 definition. Here is a little example of the XML data it contains:

```xml
<BookgDt>
    <DtTm>2025-12-01T13:47:43.673741+01:00</DtTm>
</BookgDt>
```

## Module
account_statement_import_camt

## Describe the bug
Date for entry cannot be filled because it is checking on Dt attribute but Wise CAMT.53 files are using DtTm (for date and time) instead.

## To Reproduce
Just import a CAMT.53 file replacing BookgDt > Dt with BookgDt > DtTm.

**Expected behavior**
Just import the file.

**Additional context**
Not relevant.

PR #886, opened 2026-01-10
[18.0][FIX] Issue #885 Fixes CAMT import adding DtTm to parse not just date but also date and time for entries

javiermatos:


javiermatos:
I have implemented the fix for 18.0 as this is the version I am working on. Anyway, this should work with previous versions too. It is a trivial change. Is there anything else I should do to get this merged?

javiermatos:
Already merged #895.

PR #888, opened 2026-01-15
[18.0][FIX] account_statement_import_base: sanitized acc number

ChristophAbenthungCibex:
Cherry-Pick of v16 Commit
Original PR: #729 

The field acc_number in the database is not sanitized by default.

This leads to account_bank_statement_line with an empty partner_id because the check done in _statement_line_import_update_hook() uses the sanitized_acc_number instead of the acc_number.

For a right comparison the sanitized version of the acc_number should be put in the speeddict.


pedrobaeza (approved):
Let's merge it being a direct fw-port.

/ocabot merge patch

PR #889, opened 2026-01-15
[17.0][FIX] account_statement_import_online_paypal: handle missing key

tendil:
- Fixes a crash when pulling PayPal online bank statements: PayPal can return a response without `transaction_details`, which currently breaks the pull with a server error (`KeyError: 'transaction_details'`);
- Improve error handling and return clear UserError messages;
- Ensure PayPal datetime formatting is schema-safe (no microseconds);
- Writе tests to cover the new logic and keep coverage green.

### DIFF
--- a/account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
+++ b/account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
@@ -184,6 +184,18 @@ class OnlineBankStatementProviderPayPal(models.Model):
             ("paypal", "PayPal.com"),
         ]
 
+    @api.model
+    def _paypal_format_datetime(self, dt_naive_utc):
+        """Format naive UTC datetime for PayPal query params.
+
+        PayPal is picky about schema; microseconds often break requests.
+        Use RFC3339-like format with seconds and Z (UTC).
+        """
+        if not dt_naive_utc:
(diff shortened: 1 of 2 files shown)


Easyway-IT-User (approved):
Fehler behoben.

ivs-cetmix (approved):
Code review LGTM

ivs-cetmix:
Hey @pedrobaeza looks like there was a bug in the original module, here is the fix.

pedrobaeza:
This is changing more than what the message says. I see changes in the date handling, so that changes should be reflected as well on the commit/PR message, and balance if they should be in separate commits as well.

ivs-cetmix:
@pedrobaeza thank you! So should we split this this into several commits or just update the message?

pedrobaeza:
My thumb rule is the following: if the lines of code of both patches are mixed and they aren't different code blocks, or if each patch splits across several files which are not clearly identified, then 2 commits. If not, with one commit explaining both patches is enough.

ivs-cetmix:
@pedrobaeza done, could you please check now?

pedrobaeza (approved):
Good job! The tests are related to the other commits and can accompany them, each part of the test with the fixed functionality, but that's nitpicking.

/ocabot merge patch

PR #890, opened 2026-01-16
[17.0][IMP] account_statement_import_online_paypal: map cart item_code

tendil:
Improvement import so Sales Order reference is preserved for matching.

PayPal transactions can include the Odoo-paid Sales Order name in `cart_info.item_details[].item_code`.
PR extracts those `item_code` values and writes them into the bank statement line `narration`, improving downstream payment matching and reducing manual reconciliation.

Additionally, the PR stores the original PayPal transaction JSON payload on each imported bank statement line in the technical field `transaction_details` (including fee lines), so we can later use the full provider data for troubleshooting and matching logic without re-fetching from PayPal.

Additionally, for PayPal transfer/withdrawal transactions the PR prepends PayPal’s bank reference to payment_ref to improve identification and matching.


account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
```diff
@@ -342,10 +355,36 @@ def _paypal_transaction_to_lines(self, data):
...
+        :param data: PayPal transaction payload (one ``transaction_details`` item).
+        :return: List of unique, non-empty item codes in first-seen order.
+        """
+        item_details = ((data or {}).get("cart_info") or {}).get("item_details") or []
+        seen: set[str] = set()
+        result: list[str] = []
+
+        for item in item_details:
+            code = str((item or {}).get("item_code") or "").strip()
+            if code and code not in seen:
```
etobella, line 387: Wouldn't be easier to use the set directly and then transform the set to a list for return

something like 

```python
for item in item_details:
    code = ....
    if code:
        seen.add(code)
return list(seen)
``

etobella (approved):
TEchnically looks ok, just a non blocking comment.

account_statement_import_online_paypal/models/online_bank_statement_provider_paypal.py
tendil: Thanks for your comment!

i specifically preserve the first-seen order, so list(set) is not suitable: set does not guarantee order -> non-deterministic narration
Current version - dedupe + preserve order

infonetconnections (approved):

ivs-cetmix:
Hi @etobella @OCA/banking-maintainers would appreciate you having a look at this before merged. We left 4 separate commits on purpose to make the history tracking more convenient. It would be also easier to cherry-pick them in other versions.

alexey-pelykh (approved):
Thanks for this! As maintainer of this module, really glad to see cart item code mapping land — that's been a gap for Odoo-originated PayPal payments.

The `_paypal_get_cart_item_codes` helper is clean with good deduplication logic. The `transaction_details` storage and bank reference fallback for withdrawal notes are both practical additions. Tests are thorough.

Code review LGTM.

PR #892, opened 2026-01-21
[16.0][IMP] account_statement_import_file: attachment now visible in form view of bank statement

alexis-via:
BEFORE:

before_st

AFTER : 

after_st

### DIFF
--- a/account_statement_import_file/wizard/account_statement_import.py
+++ b/account_statement_import_file/wizard/account_statement_import.py
@@ -4,7 +4,7 @@
 import base64
 import logging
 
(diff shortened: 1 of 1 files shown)


alexey-pelykh (approved):
Thanks for the contribution. Moving the attachment creation to `import_single_file` and linking via `attachment_ids` on the statement is the right fix — previously the attachment was orphaned from the form view.

Also good to see the `[0, False, line]` → `Command.create(line)` modernization.

LGTM.

etobella (approved):
LGTM

PR #893, opened 2026-01-21
[FIX] account_statement_import_sheet_file: Handle integer values in column concatenation

cav-adhoc:
When mapping multiple columns for concatenation (e.g., "Ref number,Concept"), the _get_values_from_column method would only join values if all were strings. If one column contained an integer (like reference numbers), it would return only the first value instead of concatenating both columns as expected.

This fix adds specific handling for cases where any content is an integer, converting all values to strings before joining, while preserving the original fallback behavior for other mixed data types.


alexey-pelykh (commented):
Same fix as the 19.0 version (#894), logic looks correct. Note that codecov shows the new code paths aren't covered by the existing test suite on 18.0 -- worth checking if a test case exercising integer column values could be added to confirm the fix long-term.

PR #894, opened 2026-01-22
[FIX] account_statement_import_sheet_file: Handle integer values in column concatenation

cav-adhoc:


alexey-pelykh (approved):
Thanks for this. The integer handling in column concatenation is a real issue with XLSX files -- cell values can come back as `int` rather than `str`. Fix looks correct, CI green. LGTM

cav-adhoc:
included in https://github.com/OCA/bank-statement-import/pull/957

PR #897, opened 2026-02-05
[16.0][IMP] account_statement_import_sheet_file: auto transaction ID from row hash

siddolo:
## Description

This PR adds an option to automatically generate the transaction ID during sheet import.
When enabled, the ID is computed by hashing the row values, preventing duplicates when no explicit ID column is available.

## Changes

1) Added Automatic operation ID option in the mapping; Hide transaction_id_column when the option is enabled.
2) Generate the ID via row-value hash.
3) Updated strings and translations.

### DIFF
--- a/account_statement_import_sheet_file/views/account_statement_import_sheet_mapping.xml
+++ b/account_statement_import_sheet_file/views/account_statement_import_sheet_mapping.xml
@@ -86,6 +86,7 @@
                         </group>
                         <group>
                             <field name="timestamp_column" />
+                            <field name="transaction_id_auto" />
                             <field name="currency_column" />
                             <field name="amount_type" />
                             <field
@@ -121,7 +122,10 @@
                                     'invisible': [('amount_type', '!=', 'absolute_value')],
                                     }"
                             />
-                            <field name="transaction_id_column" />
+                            <field
+                                name="transaction_id_column"
+                                attrs="{'invisible': [('transaction_id_auto', '=', True)]}"
+                            />
                             <field name="description_column" />
                             <field name="notes_column" />
                             <field name="reference_column" />
--- a/account_statement_import_sheet_file/i18n/account_statement_import_sheet_file.pot
+++ b/account_statement_import_sheet_file/i18n/account_statement_import_sheet_file.pot
@@ -495,6 +495,11 @@ msgid "Unified Chinese (gb18030)"
 msgstr ""
 
 #. module: account_statement_import_sheet_file
+#: model:ir.model.fields,field_description:account_statement_import_sheet_file.field_account_statement_import_sheet_mapping__transaction_id_auto
+msgid "Automatic operation ID"
+msgstr ""
+
+#. module: account_statement_import_sheet_file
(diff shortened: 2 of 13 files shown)


siddolo:
Could you please change the title to:
`[16.0][IMP] account_statement_import_sheet_file: auto transaction ID from row hash`

Thanks

account_statement_import_sheet_file/views/account_statement_import_sheet_mapping.xml
```diff
@@ -121,7 +122,10 @@
                                     'invisible': [('amount_type', '!=', 'absolute_value')],
                                     }"
                             />
-                            <field name="transaction_id_column" />
+                            <field
+                                name="transaction_id_column"
+                                attrs="{'invisible': [('transaction_id_auto', '=', True)]}"
```
alexey-pelykh, line 127: On 18.0+, use `invisible="transaction_id_auto"` directly instead of the `attrs` dict. The `attrs` syntax is deprecated.

alexey-pelykh (commented):
Thanks! The auto-ID hashing approach is clean — SHA-256 with proper value normalization handles the edge cases well.

Two things:
- Missing tests for the `transaction_id_auto` feature. At minimum: verify the hash is deterministic for the same row, and changes when row data changes.
- The view uses old-style `attrs` dict for the `invisible` condition — on 18.0, this should be `invisible="transaction_id_auto"` directly on the field element.

mt-software-de:
@siddolo i did something similar for v18 but implemented in the base.
I think this would fix your need as well. I can create a backport. When we have a common ground. 
https://github.com/OCA/bank-statement-import/pull/909

account_statement_import_sheet_file/views/account_statement_import_sheet_mapping.xml
siddolo: Thanks. This PR targets 16.0, where the `attrs` dict is still the supported syntax. The direct `invisible="transaction_id_auto"` expression applies to newer Odoo series, so I kept the 16.0 form here.

siddolo:
> @siddolo i did something similar for v18 but implemented in the base. I think this would fix your need as well. I can create a backport. When we have a common ground. #909

I think it's important to be able to configure this feature in the view.
In addition, I think it's more robust to hash all the fields.

siddolo:
Thanks for the review.

I rebased the PR on the latest 16.0 branch and added tests covering:
- deterministic hashes for identical row values;
- different hashes when row data changes;
- automatic transaction IDs through the real sheet parser.

Regarding the view syntax: this PR targets Odoo 16.0, where `attrs` is still the correct syntax, so I kept it unchanged.

Pre-commit passes. The current Runboat build is still pending; the preceding rebased build passed. I also checked the authenticated CI logs: both Odoo and OCB stop before reaching this module's tests while loading the unrelated `account_statement_import_online_ofx` module, with `ImportError: cannot import name 'ofxhome' from 'ofxtools'`. PR #986 fails with the same shared dependency/base issue.

@alexey-pelykh, could you please review again and remove the `stale` label if the automation does not clear it?

PR #898, opened 2026-02-06
[FIX] account_statement_import_sheet_file: Bugs with amount_type selection

feg-adhoc:
To replicate this error

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
@@ -256,11 +256,14 @@ class AccountStatementImportSheetParser(models.TransientModel):
                         mapping,
                     )
 
-            amount = _decimal("amount_column", values)
-            if not amount:
+            if mapping.amount_type == "distinct_credit_debit":
                 amount = abs(_decimal("amount_debit_column", values) or 0)
-            if not amount:
-                amount = -abs(_decimal("amount_credit_column", values) or 0)
+                if not amount:
+                    amount = -abs(_decimal("amount_credit_column", values) or 0)
+            elif mapping.amount_type == "simple_value":
+                amount = _decimal("amount_column", values)
+            elif mapping.amount_type == "absolute_value":
+                amount = abs(_decimal("debit_credit_column", values) or 0)
 
             balance = (
                 self._get_values_from_column(values, columns, "balance_column")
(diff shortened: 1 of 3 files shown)


account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
```diff
@@ -256,11 +256,14 @@ def _decimal(column_name, values):
...
+            if mapping.amount_type == "distinct_credit_debit":
                 amount = abs(_decimal("amount_debit_column", values) or 0)
-            if not amount:
-                amount = -abs(_decimal("amount_credit_column", values) or 0)
+                if not amount:
+                    amount = -abs(_decimal("amount_credit_column", values) or 0)
+            elif mapping.amount_type == "simple_value":
+                amount = _decimal("amount_column", values)
+            elif mapping.amount_type == "absolute_value":
+                amount = abs(_decimal("debit_credit_column", values) or 0)
```
alexey-pelykh, line 266: `debit_credit_column` is the D/C indicator ("D"/"C"), not a numeric column -- `_parse_decimal` on it will return 0. Should be `amount = abs(_decimal("amount_column", values) or 0)` -- the `if debit_credit is not None:` block below already handles the sign.

account_statement_import_sheet_file/views/account_statement_import_sheet_mapping.xml
```diff
@@ -77,8 +73,8 @@
                             <field name="amount_type" />
                             <field
                                 name="amount_column"
-                                invisible="amount_type == 'distinct_credit_debit'"
-                                required="amount_type != 'distinct_credit_debit'"
+                                invisible="amount_type != 'simple_value'"
```
alexey-pelykh, line 76: This hides `amount_column` for `absolute_value`, but that mode needs it for the numeric amount. The original `invisible="amount_type == 'distinct_credit_debit'"` was correct here.

alexey-pelykh (changes requested):
Thanks for tackling this -- branching the parser by `amount_type` instead of cascading is the right fix. The old fallback from `amount_column` to debit/credit was wrong when amount happened to be 0. The `simple_value` and `distinct_credit_debit` paths look correct now.

The `absolute_value` path has a problem though: it reads `debit_credit_column` as the numeric amount, but that column holds the D/C indicator (e.g. "D"/"C"). Parsing that through `_parse_decimal` gives 0. For `absolute_value`, the amount should still come from `amount_column` -- the sign is already handled by the existing `if debit_credit is not None:` block further down.

Same issue on the view side -- `amount_column` should stay visible for `absolute_value` (the old `invisible="amount_type == 'distinct_credit_debit'"` was correct). And `debit_value`/`credit_value` were removed from the form but the parser still references them (`mapping.debit_value`/`mapping.credit_value`) to determine sign -- they should stay visible when `amount_type == 'absolute_value'`.

Happy to re-review once updated.

PR #900, opened 2026-02-14
[18.0][IMP] account_statement_import_sheet_file: import pdf

jbaudoux:
Pre-process pdf to convert to csv and parse

This preprocessor allows to import some credit card pdf statements. I'm using it for Mastercard from BNP bank.
It works because pdftotext returns a result close to a csv file and requires little transformation.

I tried with a VISA pdf statement from Belfius bank and pdftotext returns nothing.

I provided the pre-processor I'm using as an example in the README.

It's maybe a bit mis-leading to let users think they can easily import any pdf statement. Nevertheless, for my use case, it's very convenient. Maybe it can help others.


pedrobaeza:
Hi, this can be already covered by https://github.com/OCA/edi/tree/18.0/base_import_pdf_by_template

pedrobaeza:
I'm seeing that I already told you about it. I don't think this should be included in the same module. It's very bloated. Maybe an extension module if you don't like the one I pointed.

jbaudoux:
> I'm seeing that I already told you about it. I don't think this should be included in the same module. It's very bloated. Maybe an extension module if you don't like the one I pointed.

Yes I agree, it's quite hackish and really depends on how the pdf is built. Even if the pre-processor is simple, it doesn't provide an easy solution out of the box. But still it's very convenient once it works. So I though I'll still share the approach.

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -1,9 +1,19 @@
 # Copyright 2019 ForgeFlow, S.L.
 # Copyright 2020 CorporateHub (https://corporatehub.eu)
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import re
+from io import BytesIO
+
+from PyPDF2 import PdfReader
```
alexey-pelykh, line 9: PyPDF2 has been deprecated — consider switching to `pypdf` (same API). `from pypdf import PdfReader` and `pypdf>=3` in dependencies.

alexey-pelykh (commented):
Thanks for this — the preprocessor concept is solid and the test wizard is a nice touch.

A few notes:

- `PyPDF2` was deprecated and renamed to `pypdf` (same API, just `from pypdf import PdfReader`). Since this is new code, worth switching to the actively maintained package.
- No tests for the preprocessor logic — at least a basic test exercising `_preprocess()` and `_eval_code()` would help with coverage and regression safety.
- The eval context only exposes `re.match`. Users working with PDF text extraction will likely need `re.search`, `re.findall`, `re.sub` too — consider exposing the full `re` module or at least a broader set.

pedrobaeza (changes requested):
I think this should be a separate module over the base one, as this one gets very bloated, and there are risks that on migration, all the features are not correctly migrated by one that doesn't use that format.

PR #902, opened 2026-03-02
[16.0][IMP] account_statement_import_online_paypal: PayPal improvements backport from PR #890

rlizana:
Backport of PayPal improvements from PR #890 (17.0 branch) to 16.0.

This PR includes 4 commits that enhance the PayPal statement importer:
- Format PayPal datetimes safely (remove microseconds)
- Store PayPal transaction details as JSON for audit trail
- Extract and map cart item codes to statement line narration
- Merge bank reference into payment_ref for better identification

I had to modify the functionality because the `transaction_details` field exists in version 17.0 (https://github.com/OCA/OCB/blob/ee41922101a7d2f37bb511aff88f77d0a4af2a62/addons/account/models/account_bank_statement_line.py#L147) but not in version 16.0


alexey-pelykh (changes requested):
Nice backport, Roberto -- the move to `urlencode(params)` and the defensive `.get()` calls throughout are solid improvements. The `total_pages` fallback to prevent infinite loops and the new error-body check in `_paypal_retrieve` are exactly the kind of robustness this module needs against PayPal API quirks.

Two things need attention though:

**`transaction_details` missing from line dicts**

The tests assert a `transaction_details` key on line dicts returned by `_paypal_transaction_to_lines`, but the code never adds it. In `_paypal_preparse_transaction` you store `_odoo_transaction_details` on the raw transaction dict, but nothing maps that into the line dict. On 17.0 this mapping likely happens in a base-module hook that doesn't exist in 16.0 — so the tests should fail as-is. Either add the `transaction_details` key to the line dict explicitly in `_paypal_transaction_to_lines`, or remove the assertion from the tests if there's genuinely no field to store it in 16.0.

**Python 3.8 compatibility**

`list[str]` and `set[str]` lowercase generic annotations in `_paypal_get_cart_item_codes` require Python 3.9+. Odoo 16.0 still officially supports 3.8. Either add `from __future__ import annotations` at the top or switch to `typing.List`/`typing.Set`.

The rest looks good — the `_paypal_format_datetime` helper is clean, cart item code extraction will be useful for SO matching, and the bank reference merge makes withdrawal lines much more readable.

rlizana:
Thanks for the review, @alexey-pelykh. I've addressed both points:

- Added from `__future__` import annotations for Python 3.8 compatibility
- Added `transaction_details` to the line dicts in `_paypal_transaction_to_lines`, extracted from `_odoo_transaction_details` set during preparsing

rlizana:
Failing check is not related to the changes in this PR, the error comes from account_statement_import_online_ofx:

`ImportError: cannot import name 'ofxhome' from 'ofxtools'`

It's a dependency issue in the account_statement_import_online_ofx module, not in account_statement_import_online_paypal

almumu:
@rlizana thank you for your work

An observation, not a request: this PR currently carries two kinds of change with quite different urgency and risk. The three fix commits (78ec31cd, dd192947, a2e3867a) are self-contained and address a crash that breaks the pull outright; 
the improvement commits add functionality and, as you noted yourself, needed adaptation for 16.0 because the `transaction_details` field only exists from 17.0. Two later commits are follow-ups to that part.

Splitting them into two PRs would let the fix half land on its own timeline while the improvement half keeps its review. 

As a data point: we applied those three commits on top of a clean 16.0 and they cherry-pick without conflicts, with the three added tests passing and no regressions in the module's existing suite.

almumu:
@rlizana  About failing tests, it seems the failure happens while **loading** `account_statement_import_online_ofx`, before any test runs — the whole registry fails to initialise, so no test is executed at all:

  ```
  CRITICAL odoo.modules.module: Couldn't load module account_statement_import_online_ofx
  ImportError: cannot import name 'ofxhome' from 'ofxtools'
    .../account_statement_import_online_ofx/models/online_bank_statement_provider_ofx.py:11
      from ofxtools import ofxhome, utils
  ```

`requirements.txt` declares `ofxtools` unpinned. The 1.x series, released in June 2026 after four years on 0.9.5, no longer exports `ofxhome` at package root. Without a pin, each run resolves whatever version is available at that moment.


It seems specific to 16.0: `account_statement_import_online_ofx` does not exist on 14.0, 15.0, 17.0 or 18.0, and this is the only branch requiring `ofxtools`  — the others only need `ofxparse`.

almumu:
@rlizana I opened #991 to bound the dependency problem, and its four checks are green.

PR #904, opened 2026-03-05
[FIX] account_statement_import_online_wise: HTTP Error 403 Forbidden

Kiplangatdan:
We noticed that on some servers, the connection to Wise meets with a 403 while on others it succeeds. This line of code seems to fix it.

### DIFF
--- a/account_statement_import_online_wise/models/online_bank_statement_provider_wise.py
+++ b/account_statement_import_online_wise/models/online_bank_statement_provider_wise.py
@@ -18,7 +18,7 @@ from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import padding, rsa
 from dateutil.relativedelta import relativedelta
 
-from odoo import _, api, fields, models
+from odoo import _, api, fields, models, service
 from odoo.exceptions import UserError
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


thomaspaulb (approved):

account_statement_import_online_wise/models/online_bank_statement_provider_wise.py
```diff
@@ -306,6 +306,7 @@ def _wise_urlopen(self, url, api_key, ott=None, signature=None):
             raise UserError(_("No API key specified!"))
         request = urllib.request.Request(url)
         request.add_header("Authorization", f"Bearer {api_key}")
+        request.add_header("Accept", "*/*")
```
thomaspaulb: I repeated my tests in shell and now, other headers, specifically the "Content-Type" one seem to be the key difference between 403 and success.

Maybe we add all:

```
request.add_header("User-Agent", "MyApp/1.0")
request.add_header("Content-Type", "application/json")
request.add_header("Connection": "keep-alive")
request.add_header("Accept-Encoding", "gzip, deflate")
```

thomaspaulb (changes requested):

thomaspaulb:
I noticed actually that the 403's are not because of these headers, or maybe not only.

Wise has a system called [SCA](https://docs.wise.com/guides/developer/auth-and-security/sca-and-2fa) that randomly gives back 403's when it requests for extra authentication. When and if this happens, is determined by a combination of factors.

If it happens, a header is returned that can be used to automate presenting the user with extra prompts to pass the 2FA test, after which things will be working again.

thomaspaulb:
Wise came back saying that SCA has been totally disabled for EU/EEA countries, due to PSD2 regulations. It's now not possible to use the API from those countries, only from US, Canada, New Zealand, Singapore and some other countries that do not have those regulations.

PR #908, opened 2026-03-11
fix: populate narration, transaction_type and ref from OFX import

StuartJMackintosh:
The OFX importer concatenates NAME, CHECKNUM and MEMO into a single payment_ref field ("payee checknum : memo"), discarding the OFX TRNTYPE and preventing reconciliation models from matching on the Note or Transaction Type conditions independently.

Distribute OFX fields to their corresponding Odoo fields:

- payment_ref: payee name only (NAME), with fallback to MEMO then TRNTYPE if NAME is empty (avoids reconciliation widget balance miscalculation per odoo/odoo#33104)
- narration: MEMO and CHECKNUM, enabling reconciliation model Note matching
- transaction_type: TRNTYPE (debit, credit, directdebit etc.), enabling reconciliation model Transaction Type matching
- ref: CHECKNUM when present, for searchable payment references

Uses markupsafe.Markup for narration to prevent Odoo wrapping plain text in HTML paragraph tags.


StuartJMackintosh:
Pre-commit failures are in setuptools-odoo-make-default and setuptools-odoo-get-requirements hooks due to pkg_resources removal in Python 3.11+. This is a CI environment issue affecting all PRs - not related to this change.

PR #909, opened 2026-03-13
[18.0][IMP] account_statement_import_base: Ensure unique_import_id is set

mt-software-de:
Right now the unique_import_id is not set by default. 
This PR allows to configure a journal to generate a unique id if not set.

### DIFF
--- a/account_statement_import_base/models/account_journal.py
+++ b/account_statement_import_base/models/account_journal.py
@@ -1,14 +1,22 @@
 # Copyright 2022 Akretion France (http://www.akretion.com/)
+# Copyright 2026 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # @author: Alexis de Lattre <alexis.delattre@akretion.com>
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
-from odoo import api, models
+from odoo import api, fields, models
 
 from odoo.addons.base.models.res_bank import sanitize_account_number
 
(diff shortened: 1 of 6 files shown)


account_statement_import_base/models/account_journal.py
```diff
@@ -1,15 +1,23 @@
...
 
 class AccountJournal(models.Model):
     _inherit = "account.journal"
 
+    ensure_unique_import_id = fields.Boolean(
+        help="Ensures that a unique_import_id is set"
+        "on a bank statement lines when importing it via bank statement import"
+        "If no unique identifier was provided, the value is created from the "
+        "account number, date, payment reference and amount."
+    )
```
mt-software-de, line 19: This should be maybe renamed to.
`auto_generate_unique_identifier`

account_statement_import_base/models/account_journal.py
```diff
@@ -53,11 +61,27 @@ def _statement_line_import_update_hook(self, st_line_vals, speeddict):
                     speeddict["account_number"][st_line_vals["account_number"]]
                 )
 
+    @api.model
+    def _get_unique_identifier_keys(self):
+        return ["account_number", "date", "payment_ref", "amount"]
```
mt-software-de, line 66: Instead of using just a few values.
We could also create a hash out of all values?

mt-software-de:
@jbaudoux could you give this a review?

PR #913, opened 2026-03-21
[18.0][FIX] account_statement_import_online_plaid: pending-to-settled dedup

dnplkndll:
## Description

When a Plaid transaction settles it receives a new `transaction_id`; the
original pending ID moves to `pending_transaction_id`. The previous code used
`transaction_id` directly as `unique_import_id`, so the OCA duplicate guard
could not detect that a settled transaction superseded its pending counterpart
— both were imported as separate statement lines.

## Changes

Add `_plaid_replace_settled_pending_lines()`, which runs before lines are handed
to the base module. For each settled transaction whose `pending_transaction_id`
matches an existing unreconciled statement line, the pending line is removed so
the settled version takes its place. Reconciled lines are never touched.

Tests cover the replace path, the reconciled-line carve-out, and add a
transaction factory plus pull/read helpers.

---
AI-assisted (Claude Code); every change reviewed, tested, and owned by the author.


Daniel15:
I think a better solution is to not sync the pending transactions at all. That's the approach I'm taking in #955, and that matches what the Stripe and GoCardless integrations do. Pending transactions should never be reconciled because it's unknown whether they'll actually settle and if the value will remain the same. In the USA for example, tips can increase the transaction value. 

There's also some unrelated changes in this PR (Mercury integration).

PR #915, opened 2026-03-24
[18.0][ADD] account_statement_import_online_wealthreader

davidwealthreader:
New module: online bank statement provider for the [Wealthreader](https://www.wealthreader.com) bank aggregation API.

Wealthreader aggregates banks from 14 countries (Spain, Portugal, Italy and Latin America among others) through a single API. This module adds it as a service to `account_statement_import_online`:

- Pulls account transactions from `POST /entities/` for the journal's period and maps them to bank statement lines (amount, operation/value date, payment reference, counterparty from transfer details, composed narration).
- Bank account selection by stored UUID, journal IBAN match, or single-account fallback.
- Credential tokenization: after the first successful login the API returns a token that is persisted and used for subsequent pulls, so bank credentials are not re-sent; token rotation is handled.
- Configurable date field (operation vs value date) and statement balance from the account's current balance.
- API errors are surfaced as `UserError` with the provider's error code and message.

52 unit tests included (`requests.post` mocked). The flow has also been exercised end-to-end against the live API on a real Odoo 18 instance using the API's public mock users (statement creation, deduplication on re-pull, token-only re-authentication, and error paths).

Wealthreader (the company) develops and will maintain this module.


pcastelovigo (approved):

davidwealthreader:
Thanks for the review, @pcastelovigo!

CI is green (Odoo and OCB tests, pre-commit, codecov) and the branch is up to date. This PR only adds a new module, so there is no impact on existing ones.

It has one approval already, so a second review would be enough to move it forward. Any feedback is welcome.

PR #926, opened 2026-04-20
[18][account_statement_import_file_reconcile_oca][FIX] Make active_ids and active_model consistent with active_id

florian-dacosta:
With the current code, in the bank statement reconciliation view, the active_id is the one of the journal, while active_ids is the id of the import wizard and active_model is the also from the wizard (`account.statement.import`)
I don't think there is a bug because of it, but it is totally inconsistent and needs to be fixed.


rvalyi (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #933, opened 2026-04-25
[18.0][FIX] account_statement_import_sheet_file: preserve file order for equal timestamps

Devstellar:
This fixes imports where multiple transactions share the same date/timestamp.

Current behavior sorts only by timestamp, which may reorder lines incorrectly when:

- the statement file contains only dates (no time column)
- multiple rows share the same timestamp
- files are exported newest-first

This may produce incorrect:

- starting balance
- running balance
- transaction ordering

This patch preserves original file order as a tie breaker while respecting the selected statement order.

Tested with Wise CSV exports.

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
@@ -98,6 +98,18 @@ class AccountStatementImportSheetMapping(models.Model):
     balance_column = fields.Char(
         help="Balance after transaction in journal's currency",
     )
+    statement_order = fields.Selection(
+        selection=[
+            ("asc", "Oldest first"),
+            ("desc", "Newest first"),
+        ],
+        default="asc",
+        required=True,
+        help=(
+            "Order of the statement file rows. This is used to preserve the "
+            "original file order when several transactions have the same timestamp."
+        ),
+    )
     original_currency_column = fields.Char(
         help=(
             "In case statement provides original currency for transactions "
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please reduce the commit to one thing, and do everything in one commit.

alexey-pelykh (approved):
Nice catch on the equal-timestamp ordering — this is a real problem and the fix is correct.

I verified the `sequence` direction against core `account.bank.statement.line` (18.0): `_order = "internal_index desc"`, and `internal_index` is built from `MAXINT - sequence` (the field itself carries the comment that the sequence "is working reversed because the default order is reversed"). So a higher `sequence` produces a smaller index component and therefore sorts as the *older* line — exactly what your inline comment states. `sequence = total_lines - index` over the ascending-sorted lines gives the oldest row the highest sequence, so it lines up with core precisely.

`_sort_lines` is sound as well: `(timestamp, _file_sequence)` for asc and `(timestamp, -_file_sequence)` for desc preserves the file's relative order within a same-timestamp group in the chosen direction. One non-blocking nit: on the `asc` path Python's `sorted` is already stable, so the `_file_sequence` tiebreaker is technically redundant there — but it keeps the asc/desc handling symmetric and explicit, so I'd leave it.

Code looks good to me; the only open item is the commit-squash already requested on the thread.

Devstellar:
Just checking in on this PR, as it has been open for a while and was approved in July.

Is there anything else needed from my side before it can be merged? If any changes, tests, or adjustments are still required, I will take care of them.

Devstellar:
@alexey-pelykh Just to clarify: the requested commit squash was already done on May 4, and the PR now contains a single commit (`580c9d9`).

PR #934, opened 2026-04-27
[16.0] [IMP] account_statement_import_sheet_file: add a hook function

luisDIXMIT:
We add a hook function that helps filter the rows. In other modules or in your private modules, you can override this function to adapt it to the characteristics of your file.

Example:
If we have a statement import file with denied transactions, we can filter them out to skip them.

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
@@ -252,3 +252,6 @@ class AccountStatementImportSheetMapping(models.Model):
 
     def _get_column_delimiter_character(self):
         return self._decode_column_delimiter_character(self.delimiter)
+
+    def _skip_row(self, values, columns):
+        return False
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_parser.py
@@ -236,6 +236,8 @@ class AccountStatementImportSheetParser(models.TransientModel):
                 if index >= footer_line:
                     continue
                 values = list(row)
+            if mapping._skip_row(values, columns):
+                continue
             if mapping.skip_empty_lines and not any(values):
                 continue
 


account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -252,3 +252,6 @@ def _decode_column_delimiter_character(self, delimiter):
 
     def _get_column_delimiter_character(self):
         return self._decode_column_delimiter_character(self.delimiter)
+
+    def skip_row(self, values, columns):
```
etobella: make it private, `_skip_row

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
```diff
@@ -780,3 +780,55 @@ def test_skip_empty_lines(self):
         self.assertEqual(statement.balance_start, 0.0)
         self.assertEqual(statement.balance_end_real, 2291.5)
         self.assertEqual(statement.balance_end, 2291.5)
+
+    def test_skip_row_method(self):
```
etobella: This specific test is not needed

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
```diff
@@ -780,3 +780,55 @@ def test_skip_empty_lines(self):
...
+            "amount_column": "100.00",
+        }
+        columns_mapping = {
+            "timestamp_column": "0",
+            "description_column": "1",
+            "amount_column": "2",
+        }
+        self.assertFalse(mapping.skip_row(values, columns_mapping))
+
+    def test_skip_row(self):
```
etobella: Why do you need this test? it is probably covered in other places... might have more sense to have a mock to ensure that it is imported partially

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
luisDIXMIT: Done, thanks!

account_statement_import_sheet_file/tests/test_account_statement_import_sheet_file.py
luisDIXMIT: What does it mean imported partialy? I removed both test.

etobella (approved):

LoisRForgeFlow (approved):
👍

etobella:
/ocabot merge patch

Can you port it to 17, 18 and 19?

pedrobaeza:
On 19, we are refactoring the module in #958 which is worth IMO.

luisDIXMIT:
> On 19, we are refactoring the module in #958 which is worth IMO.

Hi @pedrobaeza, do you mean you want to cherry-pick this into that PR, or should I create a PR?

pedrobaeza:
No, it was just a call for action to review it, hehe.

PR #943, opened 2026-05-06
UPD camt parser to include charges in transaction details

ecino:
Updated the `parse_entry` method to include both transaction details (`TxDtls`) and charge records where charges are included (`Chrgs/Rcrd[ChrgInclInd='true']`) in the `details_nodes` list. This ensures that all relevant transaction and charge information is processed.


ecino:
This one is also a very small safe-to-merge improvement in my opinion. Could someone take the time to review?

PR #944, opened 2026-05-06
[18.0] ADD multicurrency support in CAMT parser

ecino:
This enables the CAMT parser to parse bank statements in foreign currencies.


davidwul (approved):
Thanks this is a very useful feature

ecino:
We sometimes deal with transactions with several currencies within the same bank statement. This change handles better those cases. I think it would be a nice addition. Could someone review?

PR #949, opened 2026-05-12
[19.0][I18N] account_statement_import_qif: Add Chinese (zh_CN) translation

Moneeeeee:
Adds Chinese Simplified (zh_CN) translation for account_statement_import_qif (7 translatable strings).

Tested locally on Odoo 19.0 Community with --i18n-overwrite; translation file loads successfully and module remainsfunctional.
Follow-up: I will add a translation for account_statement_import_camt in this same PR once style is confirmed.


pedrobaeza:
Please use Weblate for this. 

https://www.odoo-community.org/resources/translate

PR #907, opened 2026-03-11
[19.0][MIG] account_statement_import_sheet_file: Migration to 19

RabbitJon-S73:
### Refactored Functions

* **`_parse_lines`**: Centralized data extraction by processing `label_line` and `footer_line` directly into the data flow.
* **`_parse_rows`**: Improved modularity by delegating line-specific logic to the new `_parse_one_line` helper method.
* **`_parse_one_line`**: Streamlined logic to reduce complexity and significantly enhance code readability.

PR #870, opened 2025-12-18
[MIG] account_statement_import_sheet_file: Migration to 19.0

feg-adhoc:

PR #957, opened 2026-05-22
[FIX] account_statement_import_sheet_file: Apply fixes on top of OCA 19.0 migration

cav-adhoc:
---
  Summary
  
  This PR applies ADHOC-specific fixes and improvements on top of the OCA 19.0 migration (merged via #907), replacing the
  previous ADHOC PR #894.

  Changes included

  Bug fixes from ADHOC PRs #870 / #894 not included in OCA #907:

  - openpyxl support for .xlsx files: OCA's migration only kept xlrd (old .xls format). This adds a fallback chain xlrd → 
  openpyxl → CSV so modern Excel files continue to work. Affects parse_header, _parse_lines, and _parse_rows.
  - Integer values in column concatenation (core fix of PR #894): _get_values_from_column now handles integer cell values when
   concatenating multiple columns, preventing a crash when xlrd/openpyxl returns numeric types.
  - Case-insensitive column name matching: _get_column_indexes now tries an exact match first, then falls back to
  case-insensitive before raising ValueError.
  - str() cast in _parse_decimal: Handles non-string cell values that can arrive from xlrd/openpyxl, preventing re.sub from
  crashing.

  Model fixes:

  - Remove duplicate amount_column field definition introduced in OCA #907 (defined twice with conflicting help text).
  - Fix typo in amount_type help text: "use igned amount" → "use signed amount".
  - Add _clear_amount_columns onchange: automatically clears irrelevant amount columns when amount_type changes, preventing
  invalid configurations.

  Tests:

  - No changes — kept OCA's original test structure from the 19.0 migration as-is.

  Replaces

  This PR replaces #894 (which was based on a pre-#907 migration). The previous PR should be closed.

  🤖 Generated with Claude Code

### DIFF
--- a/account_statement_import_sheet_file/i18n/es.po
+++ b/account_statement_import_sheet_file/i18n/es.po
@@ -246,8 +246,8 @@ msgstr "Recuento de líneas omitidas a pie de página"
 
 #. module: account_statement_import_sheet_file
 #: model:ir.model.fields,field_description:account_statement_import_sheet_file.field_account_statement_import_sheet_mapping__header_lines_skip_count
-msgid "Header lines skip count"
-msgstr "Recuento de líneas de cabecera omitidas"
+msgid "Header row number"
+msgstr "Número de fila del encabezado"
 
 #. module: account_statement_import_sheet_file
 #: model:ir.model.fields,field_description:account_statement_import_sheet_file.field_account_statement_import_sheet_mapping__id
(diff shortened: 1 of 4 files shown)


Andrii9090-tecnativa:
We are currently working on a refactoring of this module so that we can extend the this module, which by default will only work with CSV files. This gives us greater flexibility and the option to use modules to add the ability to import XLS and XLSX and other files.

pedrobaeza:
Replaced by #969

PR #958, opened 2026-05-26
[19.0][REF] account_statement_import_sheet_file: Refactoring

Andrii9090-tecnativa:
This refactoring separates parsing responsibilities by moving XLS and XLSX support out of the base `account_statement_import_sheet_file` module. The base module now focuses on TXT/CSV import, while format-specific parsers and tests are implemented in `account_statement_import_sheet_file_xls` and `account_statement_import_sheet_file_xlsx`. This improves modularity, reduces coupling, and makes maintenance and future format extensions easier.

@victoralmau @pedrobaeza  ping 
@Tecnativa
TT62696

PR #961, opened 2026-06-03
account_statement_import_sheet_file: add xlsx support via openpyxl

rov-adhoc:
xlrd 2.x dropped support for .xlsx files. When an xlsx was uploaded, xlrd raised XLRDError, the code fell through to the CSV path, and failed with UnicodeDecodeError because xlsx is binary.

- Try openpyxl as a fallback between xlrd failure and CSV parsing
- Add _OpenpyxlWorkbookWrapper and _OpenpyxlSheetWrapper to expose the same interface as xlrd objects (row_values, cell_type, cell_value, nrows, row_len), keeping _parse_rows unchanged
- Handle openpyxl native types: datetime objects pass through as-is, int/float values are returned directly without xldate conversion
- Fix _parse_decimal to accept int values returned by openpyxl for whole-number cells (e.g. amounts like 100)

### DIFF
--- a/account_statement_import_sheet_file/wizard/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/wizard/account_statement_import_sheet_parser.py
@@ -10,7 +10,7 @@ import re
 from collections.abc import Iterable
 from datetime import datetime
 from decimal import Decimal
-from io import StringIO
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please check #958 with a refactoring and the support for XLSX.

PR #963, opened 2026-06-11
[17.0][IMP] account_statement_import_sheet_file: add a hook function to filter certain rows from a sheet

luisDIXMIT:
Forward port of #934

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
@@ -254,3 +254,6 @@ class AccountStatementImportSheetMapping(models.Model):
 
     def _get_column_delimiter_character(self):
         return self._decode_column_delimiter_character(self.delimiter)
+
+    def _skip_row(self, values, columns):
+        return False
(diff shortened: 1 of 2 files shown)


BhaveshHeliconia (approved):
LGTM!

alexey-pelykh (approved):
Thanks for this — `_skip_row(values, columns)` is a clean extension point and I'm glad to have it in the module.

It follows the module's existing protected-hook convention (it sits right alongside `_get_column_delimiter_character` / `_decode_column_delimiter_character` on the mapping model), and defaulting to `False` makes it fully backwards-compatible: every existing mapping keeps its current behaviour, and a downstream module can override `_skip_row` to drop rows it doesn't want (subtotals, "pending" markers, footnotes, etc.).

Placement in `_parse_rows` is right, too — it runs after `values = list(row)` and before the `skip_empty_lines` short-circuit, with both `values` and `columns` in scope, so an override can filter on any raw row.

For whoever drives the merge: the red `codecov/patch` here is expected and not a real gap. The only uncovered line is the `continue`, which is unreachable until a subclass overrides `_skip_row` to return `True`, so the base module can't exercise it on its own. Fine to merge past.

PR #967, opened 2026-06-18
[19.0][MIG] account_statement_import_online_ponto: Migration to 19.0

miquelpascual:
Migration to v19

cc https://github.com/APSL 48427

@shirashi3771 @lbarry-apsl @miquelalzanillas @peluko00 @BernatObrador @cubells @palomagrc93 @ppyczko @isugac @javierobcn @LlucSG please review.


BernatObrador (approved):
LGTM

peluko00 (approved):
LGTM!

lbarry-apsl (approved):
Code and Runboat reviewed. LGTM

miquelpascual:
Hi @pedrobaeza, sorry for pinging you directly. Would it be possible to merge this?

palomagrc93 (approved):
LGTM!

account_statement_import_online_ponto/models/ponto_interface.py
```diff
@@ -88,9 +88,12 @@ def _set_access_account(self, access_data, account_number):
                 return
         # If we get here, we did not find Ponto account for bank account.
         raise UserError(
-            _(
-                "Ponto : wrong configuration, account {account} not found in {data}"
-            ).format(account=account_number, data=data)
+            self.env._(
+                "Ponto : wrong configuration, "
+                "account %(account)s not found in %(data)s",
```
pedrobaeza: Please don't modify translations, or it will force every language to retranslate. You can switch to the expected format for no pre-commit error without changing the translation itself.

account_statement_import_online_ponto/models/ponto_interface.py
```diff
@@ -133,7 +136,9 @@ def _get_request(self, access_data, url, params):
         """Interact with Ponto to get next page of data."""
         headers = self._get_request_headers(access_data)
         _logger.debug(
-            _("GET request to %(url)s with headers %(headers)s and params %(params)s"),
+            self.env._(
+                "GET request to %(url)s with headers %(headers)s and params %(params)s"
```
pedrobaeza, line 140: The same. I don't continue pointing each of these cases, but please correct all without changing the translation itself.

pedrobaeza (changes requested):
/ocabot migration account_statement_import_online_ponto

account_statement_import_online_ponto/models/ponto_interface.py
miquelpascual: Could you please give me some more information? Do you mean not modifying the string, or not using self.env._? If I leave the string as it was before, the pre-commit checks do not pass. I adapted it so that it follows the required format.

I look forward to your response. Thank you very much!

pedrobaeza: You have to leave the string as it was. The problem is not the string. You have to pass the parameters to the formatting inside the call to `_` or `self.env._`.

miquelpascual: I left it as it was but now the pre-commit is failing. Any idea on how could i fix that? Many thanks!

account_statement_import_online_ponto/models/ponto_interface.py
```diff
@@ -0,0 +1,168 @@
...
+                ponto_account.get("attributes", {}).get("reference", "")
+            )
+            if ponto_iban == account_number:
+                access_data["ponto_account"] = ponto_account.get("id")
+                return
+        # If we get here, we did not find Ponto account for bank account.
+        raise UserError(
+            self.env._(
+                "Ponto : wrong configuration, account {account} not found in {data}"
+            ).format(account=account_number, data=data)
```
pedrobaeza: I was talking about doing this:

```suggestion
            self.env._(
                "Ponto : wrong configuration, account {account} not found in {data}", account=account_number, data=data,
            )
```

But it seems the translation method doesn't handle `{}` formatters. Just `%()s`, so then yes, the change is needed.

But can you please then change all the occurrences in .po/.pot?

miquelpascual: Great! 
How should I modify the translations in this case? Should I do it directly in the PR or through Weblate?

Thanks in advice!

pedrobaeza: You can do it here as part of the migration, as it's something derived from it.

miquelpascual: Done! Many thanks for your help!

account_statement_import_online_ponto/i18n/es.po
```diff
@@ -82,31 +87,27 @@ msgstr "Proveedor de Extractos Bancarios en Línea"
...
 
 #. module: account_statement_import_online_ponto
 #. odoo-python
 #: code:addons/account_statement_import_online_ponto/models/ponto_interface.py:0
-#, python-format
-msgid "Ponto : wrong configuration, account {account} not found in {data}"
+msgid "Ponto : wrong configuration, account %(account)s not found in %(data)s"
 msgstr ""
-"Ponto : configuración errónea, cuenta {account} no encontrada en {data}"
+"Ponto : configuración errónea, cuenta %{account}s no encontrada en %{data}s"
```
pedrobaeza: `{` > `(`

pedrobaeza (commented):
In several places.

account_statement_import_online_ponto/i18n/es.po
miquelpascual: Fixed!

pedrobaeza (approved):
/ocabot merge nobump

PR #969, opened 2026-06-23
[19.0][FIX] account_statement_import_sheet_file: Remaining fixes on top of #958 refactoring

cav-adhoc:
This PR applies the remaining fixes from the closed PR #957 that were **not covered** by the refactoring PR #958 (which separated XLS/XLSX into standalone modules).

### 1. `account_statement_import_sheet_mapping.py`

**Remove duplicate `amount_column` field**

The field was defined twice — the first definition was dead code that never took effect because Python class resolution uses the last definition. Removed to avoid confusion and noise.

**Fix typo: "igned" → "signed" in `amount_type` help**

The help text read "use igned amount" which is clearly a typo for "use signed amount".

**Add `_clear_amount_columns` onchange**

When the user changes `amount_type`, the previously configured column fields (`amount_column`, `debit_credit_column`, `amount_debit_column`, `amount_credit_column`) become semantically invalid. This onchange clears them automatically, preventing stale/invalid configurations from being submitted.

**Improve `header_lines_skip_count` string/help**

Renamed string from "Header lines skip count" to "Header row number" with help "Row number where the column headers are located (first row is 0)." The old name implied a skip count when it's actually a 0-based row index — the field value IS the row number, not how many rows to skip.

### 2. `account_statement_import_sheet_parser.py`

**Case-insensitive column name matching in `_get_column_indexes`**

Different banks export files with varying capitalization (e.g., "Date" vs "date" vs "DATE"). The old code only did exact match via `header.index()`, which would raise `ValueError` on case mismatch. Now tries exact match first, then falls back to case-insensitive matching before raising. This method is inherited by both the XLS and XLSX parsers, so all formats benefit.

**`str()` cast in `_parse_decimal`**

The XLS parser (`_get_xls_row_values`) does not stringify cell values, so raw `int`/`float` values from xlrd can reach `_parse_decimal` via `_get_values_from_column`. Without this cast, `re.sub(r"[^\d...]+", "", value)` crashes with a `TypeError`. The XLSX parser already stringifies values, so this is mainly a safety net for XLS and any future parser.

### 3. `i18n/es.po`

Updated translations for the new `header_lines_skip_count` string and help text.

### DIFF
--- a/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
+++ b/account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
@@ -77,9 +77,6 @@ class AccountStatementImportSheetMapping(models.Model):
             "transaction from"
         ),
     )
-    amount_column = fields.Char(
-        help="Amount of transaction in journal's currency",
-    )
     amount_debit_column = fields.Char(
         string="Debit amount column",
         help="Debit amount of transaction in journal's currency",
@@ -122,7 +119,7 @@ class AccountStatementImportSheetMapping(models.Model):
         required=True,
         default="simple_value",
         help=(
-            "Simple value: use igned amount in amount column\n"
+            "Simple value: use signed amount in amount column\n"
             "Absolute Value: use a same column for debit and credit\n"
             "(absolute value + indicate sign)\n"
             "Distinct Credit/debit Column: use a distinct column for debit and credit"
@@ -173,8 +170,8 @@ class AccountStatementImportSheetMapping(models.Model):
         default="0",
     )
     header_lines_skip_count = fields.Integer(
-        string="Header lines skip count",
-        help="Set the Header lines number.",
+        string="Header row number",
+        help="Row number where the column headers are located (first row is 0).",
         default="0",
     )
     skip_empty_lines = fields.Boolean(
@@ -231,6 +228,13 @@ class AccountStatementImportSheetMapping(models.Model):
             self.float_thousands_sep = "comma"
         elif "comma" == self.float_thousands_sep == self.float_decimal_sep:
             self.float_thousands_sep = "dot"
+
+    @api.onchange("amount_type")
+    def _clear_amount_columns(self):
+        self.amount_column = False
+        self.debit_credit_column = False
+        self.amount_debit_column = False
+        self.amount_credit_column = False
 
     @api.constrains("offset_column")
     def _check_columns(self):
--- a/account_statement_import_sheet_file/wizard/account_statement_import_sheet_parser.py
+++ b/account_statement_import_sheet_file/wizard/account_statement_import_sheet_parser.py
@@ -413,6 +413,38 @@ class AccountStatementImportSheetParser(models.TransientModel):
         return [transaction]
 
     @api.model
+    def _check_decimal_mark(self, value, thousands, decimal):
+        """Refuse a value whose decimal mark the mapping does not account for.
+
+        A thousands separator always groups three digits, so a "." or a "," that
+        is followed by fewer than three digits at the end of the value can only
+        be a decimal mark. When it is not the configured one it gets stripped as
+        noise, which multiplies the amount by ten or a hundred -- silently, and
+        with no way to tell afterwards, since the factor varies per value.
+
+        Skipped when no decimal separator is configured: that mode reads the
+        last digits as the decimals, so it has a contract of its own.
+        """
+        if not decimal:
+            return
+        cleaned = re.sub(r"[^\d\-+.,]+", "", value)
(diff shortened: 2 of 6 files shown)


pedrobaeza:
Hi, Camilla, thanks for the head-up.

Please check pre-commit, and write here the changes done and the motivation.

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -77,9 +77,6 @@ class AccountStatementImportSheetMapping(models.Model):
             "transaction from"
         ),
     )
-    amount_column = fields.Char(
```
pedrobaeza, line 80: Why removing this field?

cav-adhoc: This definition is overridden by the later one, so it has no effect here [(line 131)](https://github.com/OCA/bank-statement-import/pull/969/changes#diff-836eb5b89e54bc4efa1341519457a08eba2465553a69b529dc9bf14488e81ecbL131)

pedrobaeza: OK, understood, thanks.

alexey-pelykh (commented):
Thanks for cleaning these up. One small, non-blocking robustness note on the new case-insensitive fallback in `_get_column_indexes`:

```python
if h.lower() == column_name_or_index.lower()
```

`column_name_or_index` is safe (it comes from the Char mapping field, so it's always a string by the time it reaches here), but `h` is a raw header cell. If a header cell arrives as a non-string — which can happen with spreadsheet sources where a header is numeric- or date-typed — then `h.lower()` raises `AttributeError` instead of the clean `ValueError` the exact-match path used to raise. This PR already guards for exactly this shape in `_parse_decimal` (`if not isinstance(value, str): value = str(value)`), so `str(h).lower()` here would keep the two consistent.

cav-adhoc:
> Thanks for cleaning these up. One small, non-blocking robustness note on the new case-insensitive fallback in `_get_column_indexes`:
> 
> ```python
> if h.lower() == column_name_or_index.lower()
> ```
> 
> `column_name_or_index` is safe (it comes from the Char mapping field, so it's always a string by the time it reaches here), but `h` is a raw header cell. If a header cell arrives as a non-string — which can happen with spreadsheet sources where a header is numeric- or date-typed — then `h.lower()` raises `AttributeError` instead of the clean `ValueError` the exact-match path used to raise. This PR already guards for exactly this shape in `_parse_decimal` (`if not isinstance(value, str): value = str(value)`), so `str(h).lower()` here would keep the two consistent.
> 

@alexey-pelykh Done!

cav-adhoc:
@pedrobaeza @alexey-pelykh Hi! Just wanted to check if there’s anything else needed from my side, or if we could get this PR merged when you have a chance. Thanks!

account_statement_import_sheet_file/models/account_statement_import_sheet_mapping.py
```diff
@@ -232,6 +229,13 @@ def onchange_decimal_separator(self):
         elif "comma" == self.float_thousands_sep == self.float_decimal_sep:
             self.float_thousands_sep = "dot"
 
+    @api.onchange("amount_type")
+    def _clear_amount_columns(self):
```
pedrobaeza: Please use the convention name:


```suggestion
    def _onchange_amount_type(self):
```

account_statement_import_sheet_file/wizard/account_statement_import_sheet_parser.py
```diff
@@ -144,7 +144,22 @@ def _get_column_indexes(self, header, column_name, mapping):
                     column_indexes.append(column_index)
             else:
                 if column_name_or_index:
-                    column_indexes.append(header.index(column_name_or_index))
+                    try:
+                        column_indexes.append(header.index(column_name_or_index))
+                    except ValueError:
```
pedrobaeza: I don't think we should add this fallback, as it seems random. You may have the same everything in upper, or with other upper combination, so for me, the match should continue exact.

account_statement_import_sheet_file/wizard/account_statement_import_sheet_parser.py
```diff
@@ -397,13 +412,48 @@ def _update_currency(column_name, value):
 
         return [transaction]
 
+    @api.model
+    def _check_decimal_mark(self, value, thousands, decimal):
```
pedrobaeza: I don't understand this use case to be included here.

pedrobaeza (approved):

PR #974, opened 2026-07-08
[18.0][ADD] account_statement_import_camt: Fill payment_ref field from bank statement using simplified AddtlNtryInf tag

javiermatos:
This PR extends que CAMT.053 parser to extract payment_ref data from CAMT.053 valid files that use a simplified AddtlNtryInf tag for storing information about the transaction. This applies to Wise bank statements in format CAMT.053.

### DIFF
--- a/account_statement_import_camt/models/account_statement_import_camt_parser.py
+++ b/account_statement_import_camt/models/account_statement_import_camt_parser.py
@@ -337,6 +337,13 @@ class AccountStatementImportCamtParser(models.AbstractModel):
 
         details_nodes = node.xpath("./ns:NtryDtls/ns:TxDtls", namespaces={"ns": ns})
         if len(details_nodes) == 0:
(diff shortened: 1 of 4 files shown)


clementmbr:
Hi @javiermatos I tried to use your PR but I realised it didn't catch the good transaction description for the wire transfers in Wise.com so I made a new one which is more complete than yours : #1005 

Is it ok for you?

clementmbr (changes requested):
Hi @javiermatos I tried to use your PR but I realised it didn't catch the good transaction description for the wire transfers in Wise.com so I made a new one which is more complete than yours : https://github.com/OCA/bank-statement-import/pull/1005

Is it ok for you?

Devstellar:
Superseded by #1006, which consolidates the no-`TxDtls` case from this PR together with the single-`TxDtls` case while preserving the existing behavior for entries containing multiple `TxDtls`.

PR #976, opened 2026-07-20
[18.0][FIX] account_statement_import_camt: TxDtls amounts not equals to Ntry amount

jguenat:
I found an issue where the imported bank statement lines where not correct.

Is some cases, e.g payment fees (Chrgs...), the sum of TxDtls/amt is not equals to the Ntry/Amt.

exemple 
Ntry/Amt 1526
Ntry/NtryDtls/TxDtls/Amt 1500
Ntry/NtryDtls/TxDtls/Chrgs/Rcrd/Amt 26

The issue with the current implementation is that 1526 will be lost and replaced by 1500. The statement line is then wrong by 26.

One solution would have been to look for Amt in Chrgs but I was not able to find a standard for this.

The chosen approach is to check if the sum of all  TxDtls is equals to the Ntry/Amt. If not then add the difference to the bigger TxDtls. The advantage of this is to cover other cases that might happen (other fees....) across different banks CAMT implementation.

Also add the Chrgs info to the narration.

### DIFF
--- a/account_statement_import_camt/models/account_statement_import_camt_parser.py
+++ b/account_statement_import_camt/models/account_statement_import_camt_parser.py
@@ -166,6 +166,13 @@ class AccountStatementImportCamtParser(models.AbstractModel):
             transaction["narration"],
             f"{_('Cheque Number')} (Refs/ChqNb)",
         )
+        self.add_value_from_node(
+            ns,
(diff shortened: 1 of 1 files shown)


jguenat:
@ecino Did this happened to you?

PR #977, opened 2026-07-21
[18.0][FIX] account_statement_import_online_paypal: handle missing key

victoralmau:
FWP from 17.0: https://github.com/OCA/bank-statement-import/pull/889

- Fixes a crash when pulling PayPal online bank statements: PayPal can return a response without `transaction_details`, which currently breaks the pull with a server error (`KeyError: 'transaction_details'`);
- Improve error handling and return clear UserError messages;
- Ensure PayPal datetime formatting is schema-safe (no microseconds);
- Writе tests to cover the new logic and keep coverage green.

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa


pedrobaeza (approved):
The same as 17, so fast-tracking it:

/ocabot merge patch

PR #984, opened 2026-08-05
[19.0][MIG] account_statement_import_online_gocardless: Migration to version 19.0

adasatorres-tecnativa:
Standard migration.
@Tecnativa TT58583


adasatorres-tecnativa:
ping @pedrobaeza

account_statement_import_online_gocardless/static/src/js/select_bank_widget.esm.js
```diff
@@ -36,6 +36,9 @@ export class GocardlessDialog extends Component {
     }
     get_institutions(country, searchString) {
         var institutions = this.props.context.institutions;
+        console.log(this.props.context);
```
pedrobaeza: Please remove this.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
Code review LGTM.
I get this error, but I imagine it is because I don't have a valid account and `institutions` is None. However, if you can avoid the error, that would be great.
Image

PR #985, opened 2026-08-11
[T3354] FIX account_statement_import_camt54: match journal on NtryRef…

danpa32:
# T3354 — Fix failing quality test: Income processing: import CAMT files

## Context / Problem Statement

The import of CAMT files chooses the destination journal automatically from
the account number found in the XML. It always reads it from
`Acct/Id/IBAN` — but for **camt.054** files, the account should instead be
read from `Ntry/NtryRef`, per the v14 branch:
https://github.com/CompassionCH/bank-statement-import/blob/11f86e402efc2c40ef59a3447b53b05f636da56f/account_statement_import_camt54/models/parser.py#L111

## Background: why Acct/IBAN and Ntry/NtryRef differ

Swiss banks issue a **QR-IBAN** — a special IBAN reserved for scanning
QR-bills — in addition to a company's regular IBAN, both pointing at the
same physical account. A camt.054 QR-bill batch reports the QR-IBAN via
`Ntry/NtryRef`, while the notification's own `Acct/IBAN` still shows the
regular IBAN. Compassion has two separate Odoo journals for the same
physical account ("Raiffeisen Current Account CHF" vs "QRR Raiffeisen"),
matched by whichever value ends up as the file's `account_number`.

## Root cause

`account_statement_import_camt54/models/parser.py`'s `CamtParser` (v18) had
no `parse_statement()` override at all — it inherited the base
`account_statement_import_camt` parser's version unchanged, which always
sets `account_number` from `Acct/Id/IBAN` regardless of camt version. The
v14 branch already has this override; it was simply never ported to v18.

## What changed

- `models/parser.py`: added `parse_statement()` — for camt.054 files only,
  prefer `Ntry[1]/NtryRef` over `Acct/Id/IBAN` when setting
  `account_number` (falls back to IBAN if `NtryRef` is absent). Ported
  directly from the v14 branch.

## How to test manually

1. Get a camt.054 file where the notification's `Acct/IBAN` is the regular
   account and `Ntry/NtryRef` is the QR-IBAN (e.g. the file attached to the
   ticket).
2. Accounting → import the file via the statement-import wizard, letting it
   auto-detect the journal (don't pre-select one).
3. Confirm the resulting bank statement lands in the QRR-dedicated journal
   ("QRR Raiffeisen"), not the regular account's journal ("Raiffeisen
   Current Account CHF").

### DIFF
--- a/account_statement_import_camt54/models/parser.py
+++ b/account_statement_import_camt54/models/parser.py
@@ -1,6 +1,8 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import re
+
 from odoo import models
 from odoo.exceptions import UserError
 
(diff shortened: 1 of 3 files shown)


danpa32:
Wrong click when I wanted to create this PR for my fork...sorry

PR #991, opened 2026-08-20
[FIX] account_statement_import_online_ofx: pin ofxtools below 1.0

almumu:
The module imports `ofxhome` from `ofxtools`:

    from ofxtools import ofxhome, utils

That name is gone. It is importable up to ofxtools 0.9.5 and was removed in 1.0.0, so a fresh install picks up the current release (1.1.1) and the module fails to load with:

    ImportError: cannot import name 'ofxhome' from 'ofxtools'

The manifest did not bound the version, and because the failure happens at import time it takes the whole test run of the repository down, not only this module.

Bounding the dependency restores it. Migrating to the standalone `ofxhome` distribution would be the proper fix, but that one exposes a different API (`OFXHome`, `Institution`, `InstitutionList`) instead of the `lookup()` this module calls, so it belongs in its own change.


aritzolea (approved):

WesleyOliveira98 (approved):

WesleyOliveira98:
ping @OCA/banking-maintainers

pedrobaeza (approved):
Let's apply for now this patch

/ocabot merge patch

PR #1000, opened 2026-08-29
[19.0][FIX] account_statement_import_qif: parse Australian dd/mm dates

### LINKED ISSUE #982: QIF Statement Imports in Australia Parsing as US Date Format.
## Module
account_statement_import_qif

## To Reproduce
**Affected versions**: 19.0.1.0.0

1. Using the QIF importer, import range of transactions in QIF format using Australian date format of dd/mm/yyyy, for example the following demo transactions;
!Type:Bank
D31/07/2026
T60.00
Transaction 1
LDEP
^
D08/07/2026
T360.00
Transaction 2
LDEP
^
D01/07/2026
T0.00
Transaction 3
LJDR
^

You can see for the first import, as there is no 31st Month, the importer handles this one correctly as date = 31/07/2026, but for the other 2 transactions, these resolve back to front so their dates are 07/08/2026 and 07/01/2026 respecitvely.

**Expected behavior**
A clear and concise description of what you expected to happen. When the accouinting package is localized to Australia and all currencies listed are AUS only, australian date formats should be respected for bank statement import using the QIF format.

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

ryanduguid:
Fixes #982.

Uses unambiguous dates in a QIF file to select day-first or month-first parsing. If the file supplies no usable preference, the importer uses the company country, then the user language with a company-language fallback. Australian `08/07/2026` therefore imports as 8 July, while the existing US fixture keeps its month-first interpretation.

Explicit year-first dates retain year/month/day order even in a day-first file or locale. Regression coverage includes an ISO-only Australian file and a mixed ISO/day-first file.

Verification, 10 September 2026:

- Six local cases passed against the actual parser method, including ISO, slash-separated and compact year-first dates, Australian dates and US dates.
- The repository's pre-commit checks passed on both changed files. Pylint reports existing unknown-option warnings in the shared configuration.
- Odoo TransactionCase coverage was added. Odoo and OCB integration tests run in upstream CI; neither Docker nor WSL is installed on the local Windows host.

No live bank data was used. This update used AI assistance; the verification above uses fabricated inputs.

### DIFF
--- a/account_statement_import_qif/qif_dates.py
+++ b/account_statement_import_qif/qif_dates.py
@@ -1,0 +1,74 @@
+# Copyright 2026 Ryan Duguid <ryan@duguid.com.au>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import re
+
+# Numeric dates in these countries are written day-first (dd/mm/yyyy).
+# Used only when the QIF file itself does not prove a format.
+QIF_DAYFIRST_COUNTRIES = frozenset(
+    {
+        "AU",
+        "NZ",
+        "GB",
+        "IE",
+        "ZA",
+        "IN",
+        "DE",
+        "FR",
+        "ES",
+        "IT",
+        "NL",
+        "BE",
+        "PT",
+        "AT",
+        "CH",
+        "SE",
+        "NO",
+        "DK",
+        "FI",
+        "PL",
+        "BR",
+    }
+)
+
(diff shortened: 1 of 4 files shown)


ryanduguid:
The red Odoo/OCB checks are the midnight-UTC flake in `account_statement_import_online` (`2 != 1` on `test_wizard` / `test_wizard_on_journal` / `test_pull_scheduled`), not the QIF date change. QIF tests passed. 19.0 port of #918 is #1001.

PR #999, opened 2026-08-29
[19.0][FIX] account_statement_import_qif: Parse Australian DMY dates

ryanduguid:
Fixes #982

## Summary

- detect day/month order from unambiguous QIF dates (a day or month number above 12)
- fall back to the user language date format, then to day-first company countries (`AU`, `NZ`, `GB`, `IE`, `ZA`, `IN`)
- keep US month-first files working, including the existing `8/15/13` fixture

`dateutil.parser.parse()` defaults to month-first. In the report, `D31/07/2026` parsed as 31 July (month 31 is invalid) while `D08/07/2026` and `D01/07/2026` flipped to 7 August and 7 January.

## Verification

- `pre-commit run --files account_statement_import_qif/wizards/account_statement_import_qif.py account_statement_import_qif/tests/test_import_bank_statement.py account_statement_import_qif/__manifest__.py account_statement_import_qif/readme/CONTRIBUTORS.md account_statement_import_qif/README.rst` (passes)
- GitHub CI and Runboat remain the integration checks for `oca_run_tests`

## Ownership
I reviewed, understand, and take responsibility for every line in this contribution.

PR #1001, opened 2026-08-29
[19.0][FIX] account_statement_import_online: fix flaky midnight-UTC test

ryanduguid:
Forward-port of #918 to 19.0.

## Summary

- pin `cls.now` to noon UTC so `now - 1 hour` stays on one calendar day

## Problem

`test_wizard` and `test_wizard_on_journal` fail when CI runs between 00:00 and 00:59 UTC:

```
AssertionError: 2 != 1
Names and dates for statements found: BANK/2026-08-28 - 2026-08-28, BANK/2026-08-29 - 2026-08-29
```

Daily statement mode opens a period per calendar day. A one-hour window that crosses midnight becomes two statements. Seen on #999 and #1000 (00:25–00:29 UTC on 2026-08-29). 19.0 itself is green when tests run later in the day.

#918 already pins `cls.now` on 18.0. That is enough here too. `test_pull_scheduled` keeps a noon-aligned `next_run`, so the 1-hour window `_adjust_schedule()` builds from wall-clock `datetime.now()` never spans two calendar days.

An earlier attempt to `mock.patch` `datetime.now` failed on CPython (`TypeError: cannot set 'now' attribute of immutable type 'datetime.datetime'`). That mock is not needed and is not used.

## Verification

- Isolated check that `mock.patch` of `datetime.now` raises the CI TypeError
- Hour-by-hour simulation of `_adjust_schedule` + daily statement split with noon-aligned `next_run` (0 failing hours)
- GitHub CI and Runboat remain the integration checks for `oca_run_tests`

## Ownership
I reviewed, understand, and take responsibility for every line in this contribution.

### DIFF
--- a/account_statement_import_online/tests/test_account_bank_statement_import_online.py
+++ b/account_statement_import_online/tests/test_account_bank_statement_import_online.py
@@ -103,14 +103,10 @@ class TestAccountBankAccountStatementImportOnline(common.TransactionCase):
     def test_pull_scheduled(self):
         self.provider.next_run = self.now - relativedelta(days=15)
         self._getExpectedStatements(0)
-        # _adjust_schedule() reads datetime.now(), not cls.now. Hold it at
-        # the pinned noon so a 00:00-00:59 UTC run does not cross midnight.
-        with mock.patch(
-            "odoo.addons.account_statement_import_online.models"
-            ".online_bank_statement_provider.datetime.now",
-            return_value=self.now,
-        ):
-            self.provider.with_context(step={"hours": 8})._scheduled_pull()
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Isn't possible to use `freeze_time` instead?

ryanduguid:
The requested freeze_time approach has already landed through #1003. I checked the current 19.0 branch: test_pull_scheduled, test_wizard and test_wizard_on_journal each use Odoo's freeze_time decorator and read the frozen clock inside the test. Closing this older noon-alignment PR as superseded so it does not reintroduce the previous approach.

PR #1003, opened 2026-08-29
[19.0][FIX] account_statement_import_online: freeze the flaky test clock

ryanduguid:
## Problem

Three tests derive one-hour ranges from the live UTC clock. During the first hour after midnight, those ranges span two calendar dates, so the daily dummy provider creates two statements instead of one. The recorded 2026-08-29 Odoo and OCB runs failed `test_pull_scheduled`, `test_wizard` and `test_wizard_on_journal` with `2 != 1`.

## Fix

Apply Odoo 19's `freeze_time` at noon UTC to each of the three affected test methods. Each method derives its window from `fields.Datetime.now()` while its decorator is active, so test inputs and production clock reads share one deterministic instant without freezing the whole `TransactionCase` class or leaking state into other tests.

The original assertions remain unchanged. There are no production-code changes. The effective upstream diff is the contributor copyright line, the `freeze_time` import, three method decorators and three frozen clock reads.

## Verification

- `python -m compileall -q account_statement_import_online` — passed
- OCA pre-commit on the touched test file — passed, including Ruff, Ruff format and mandatory Pylint
- `git diff --check` — passed
- branch contains one commit above `19.0`

Odoo, OCB, dependency detection, Runboat and both Codecov gates are the hosted integration checks for the new head.

## Ownership
I own and take responsibility for this contribution.

### DIFF
--- a/account_statement_import_online/tests/test_account_bank_statement_import_online.py
+++ b/account_statement_import_online/tests/test_account_bank_statement_import_online.py
@@ -1,6 +1,7 @@
 # Copyright 2019-2020 Brainbean Apps (https://brainbeanapps.com)
 # Copyright 2019-2020 Dataplug (https://dataplug.io)
 # Copyright 2022-2023 Therp BV (https://therp.nl)
+# Copyright 2026 Ryan Duguid <ryan@duguid.com.au>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 import logging
 from datetime import date, datetime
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
Please don't put `freeze_time` at class level, as that has side effects in other tests not releasing correctly the frozen date. Instead, put the decorator on each method that should be frozen.

pedrobaeza (commented):
Thanks. Please reduce the number of commits to one.

pedrobaeza:
Still pending the number of commits to be only 1.

ryanduguid:
Squashed to one commit (86673f2); all checks are green. Please re-review when convenient.

pedrobaeza (approved):
We can fast-track it as straight-forward only touching tests:

/ocabot merge patch

PR #1004, opened 2026-08-31
[18.0][FIX] account_statement_import_online_gocardless: match non European accounts

JordiBForgeFlow:
The account matching done when finishing a requisition relied exclusively on the `iban` key of the `/accounts/{id}` endpoint, which raised a KeyError (or simply never matched) for non European accounts, as GoCardless returns a null IBAN and BBAN for them.

The account number of those accounts is available in the `/accounts/{id}/details` endpoint, so it's now used as a fallback. For USD accounts, the `bban` key holds the account number and `additionalAccountData/secondaryIdentification` the routing number, so the concatenation of both is considered as well, as some journals register the bank account that way.

The details endpoint is only requested when the main one doesn't match, as GoCardless applies a strict daily rate limit on it.

All the candidate numbers are now compared sanitized on both sides, using the same sanitizing that Odoo applies to `sanitized_acc_number`.


JordiBForgeFlow:
@binhexsystems @jelenapoblet Might be of your interest

jelenapoblet:
Hi @JordiBForgeFlow , thanks for thinking of us, but I don't think we use this module for our clients.

PR #1005, opened 2026-09-03
[18.0][FIX] account_statement_import_camt: fill payment_ref with  AddtlNtryInf

clementmbr:
Some formats (like Wise.com) put the transaction description in a Ntry-level `AddtlNtryInf` node, outside any `TxDtls`.

Use it as a fallback `payment_ref` if _parse_transaction_details_ (which runs later on the TxDtls node) does not find a more precise one.

Replace #974 :
In #974 the Ntry-level `AddtlNtryInf` node is used to fill `payment_ref` only if there is not TxDtls node (`len(details_nodes) == 0`), but in the Wise.com statements the transfer statements do have a TxDtls node even if the transaction description is still in the Ntry-level `AddtlNtryInf` node. (cf the improved account_statement_import_camt/tests/samples/test-camt053-no-txdtls)

### DIFF
--- a/account_statement_import_camt/models/account_statement_import_camt_parser.py
+++ b/account_statement_import_camt/models/account_statement_import_camt_parser.py
@@ -302,6 +302,17 @@ class AccountStatementImportCamtParser(models.AbstractModel):
             transaction["narration"],
             f"{_('Additional Entry Information')} (AddtlNtryInf)",
         )
+        # Some formats (like Wise.com) put the transaction description in a
+        # Ntry-level `AddtlNtryInf` node, outside any `TxDtls`. Use it as a
+        # fallback `payment_ref` if parse_transaction_details (which runs
+        # later on the TxDtls node) does not find a more precise one.
+        addtl_infos = node.xpath("./ns:AddtlNtryInf", namespaces={"ns": ns})
+        if addtl_infos and transaction.get("payment_ref") == "/":
+            addtl_info = "\n".join(
+                x.text.strip() for x in addtl_infos if x.text and x.text.strip()
+            )
+            if addtl_info:
+                transaction["payment_ref"] = addtl_info
         self.add_value_from_node(
             ns,
             node,
--- a/account_statement_import_camt/tests/samples/golden-camt053-no-txdtls.pydata
+++ b/account_statement_import_camt/tests/samples/golden-camt053-no-txdtls.pydata
@@ -1,0 +1,50 @@
+('EUR',
+ 'BE00000000000000',
+ [{'balance_end_real': 100.0,
+   'balance_start': 0.0,
+   'date': '2026-06-28T21:40:58.810767+02:00',
+   'name': '52995589-87806609-1783466312332',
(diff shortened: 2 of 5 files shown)


clementmbr:
Is it ok for you @javiermatos @ayushin ?

javiermatos:
Hi @clementmbr . Sure it is ok. Thank you.

clementmbr:
@javiermatos great! can review/approve the PR then? :) Thanks!

Devstellar:
Hi @clementmbr, I ran into the same Wise CAMT.053 case and tested this approach against the existing parser tests.

There is one edge case worth considering before changing the existing `golden-camt053-txdtls.pydata`.

`Ntry/AddtlNtryInf` is entry-level information, while one `Ntry` may contain multiple `TxDtls`. If the entry-level `AddtlNtryInf` is assigned before iterating over `TxDtls`, the same description is propagated to every resulting transaction.

This is exactly what happens with the existing `test-camt053-txdtls` fixture: one entry contains multiple `TxDtls` and one entry-level:

`CRÉDIT GROUPÉ BVR TRAITEMENT DU 22.03.2017 ...`

With the unconditional fallback, that same text becomes the `payment_ref` of both transactions. This is why the existing golden file has to be changed in this PR.

I tested an alternative which preserves the existing multiple-`TxDtls` behavior while still fixing Wise:

- **0 `TxDtls`**: use `Ntry/AddtlNtryInf` as fallback
- **1 `TxDtls`**: use `Ntry/AddtlNtryInf` as fallback
- **More than 1 `TxDtls`**: do not propagate an entry-level description to each child transaction

I also added a regression test reproducing the Wise case where a single `TxDtls` exists only to provide structured `RltdPties`, while the payment description remains in `Ntry/AddtlNtryInf`.

Validation:

- The new regression test fails before the fix (`payment_ref == "/"`)
- It passes with the fix
- All existing `TestParser` tests pass unchanged, including the multiple-`TxDtls` fixture

The implementation is essentially:

```python
details_nodes = node.xpath(
    "./ns:NtryDtls/ns:TxDtls",
    namespaces={"ns": ns},
)

if len(details_nodes) <= 1:
    addtl_info = node.xpath("./ns:AddtlNtryInf", namespaces={"ns": ns})
    if (
        addtl_info
        and addtl_info[0].text
        and transaction.get("payment_ref") == "/"
    ):
        transaction["payment_ref"] = addtl_info[0].text.strip()
```

I have the regression test and patch available here:

- Branch: [Devstellar:18.0-pr974-addtlntryinf-single-txdtls](https://github.com/Devstellar/bank-statement-import/tree/18.0-pr974-addtlntryinf-single-txdtls)
- Commit: [a6fbdb8](https://github.com/Devstellar/bank-statement-import/commit/a6fbdb8)

I think this avoids changing the semantics of existing grouped CAMT entries while covering both Wise cases.

Devstellar:
Superseded by #1006, which consolidates both cases while avoiding propagation of entry-level `Ntry/AddtlNtryInf` to multiple child `TxDtls`.

PR #1006, opened 2026-09-04
[18.0][FIX] account_statement_import_camt: handle AddtlNtryInf fallback

Devstellar:
This PR supersedes #974 and #1005, consolidating both CAMT.053 `AddtlNtryInf` cases into a single implementation.

Some CAMT.053 producers, such as Wise, place the transaction description in the entry-level `Ntry/AddtlNtryInf` node instead of inside `TxDtls`.

There are two relevant cases:

- An entry has **no `TxDtls`**.
- An entry has **exactly one `TxDtls`**, for example to provide structured `RltdPties`, while the transaction description remains in `Ntry/AddtlNtryInf`.

In both cases, `Ntry/AddtlNtryInf` can safely be used as a fallback for `payment_ref`.

However, an `Ntry` may also contain **multiple `TxDtls`**. In that case, `AddtlNtryInf` describes the entry as a whole and should not be propagated as the same `payment_ref` to every child transaction.

The resulting behavior is therefore:

- **0 `TxDtls`**: use `Ntry/AddtlNtryInf` as fallback.
- **1 `TxDtls`**: use `Ntry/AddtlNtryInf` as fallback.
- **More than 1 `TxDtls`**: keep the existing per-transaction behavior and do not propagate the entry-level description.

This keeps the existing grouped CAMT behavior unchanged while supporting both Wise formats encountered in practice.

Regression tests are included for:

- `Ntry/AddtlNtryInf` without `TxDtls`.
- `Ntry/AddtlNtryInf` with a single `TxDtls` containing structured related-party information.

Validation:

- The new single-`TxDtls` regression test fails before the fix with `payment_ref == "/"`.
- It passes with the fix.
- All existing `TestParser` tests pass unchanged, including the multiple-`TxDtls` fixture.

Test result:

```text
0 failed, 0 error(s) of 6 tests
```


clementmbr (changes requested):
Hi @Devstellar thanks for your fix and new PR. :pray: 
Indeed, we can fallback to Ntry-level `AddtlNtryInf` only if there is only one `TxDtls` !

Can you just had a comment in the code about why we do this?

(also you can not forget to add the "[18.0]" in the PR's title ;) )

account_statement_import_camt/models/account_statement_import_camt_parser.py
```diff
@@ -336,6 +336,16 @@ def parse_entry(self, ns, node):
         )
 
         details_nodes = node.xpath("./ns:NtryDtls/ns:TxDtls", namespaces={"ns": ns})
+
```
clementmbr, line 339:
```suggestion
    # Some formats (like Wise.com) put the transaction description in a
    # Ntry-level `AddtlNtryInf`, outside any `TxDtls`. Use it as a
    # fallback `payment_ref` when there is a single `TxDtls` node (or none)
    # carrying no description.
```

account_statement_import_camt/models/account_statement_import_camt_parser.py
clementmbr: delete "if"

account_statement_import_camt/tests/test_import_bank_statement.py
```diff
@@ -95,6 +95,15 @@ def test_parse_txdtls(self):
...
         self._do_parse_test("test-camt053-no-ntry", "golden-camt053-no-ntry.pydata")
 
+    def test_parse_no_txdtls(self):
+        self._do_parse_test("test-camt053-no-txdtls", "golden-camt053-no-txdtls.pydata")
+
+    def test_parse_addtlntryinf_with_txdtls(self):
+        self._do_parse_test(
+            "test-camt053-addtlntryinf-with-txdtls",
+            "golden-camt053-addtlntryinf-with-txdtls.pydata",
+        )
```
clementmbr: It can be good to give more explicit test names, like : 

test_parse_txdtls -> test_parse_2_txdtls

test_parse_no_txdtls -> test_parse_addtlntryinf_with_no_txdtls

test_parse_addtlntryinf_with_txdtls -> test_parse_addtlntryinf_with_1_txdtls

(and rename the test files in the same way)

clementmbr (changes requested):

account_statement_import_camt/models/account_statement_import_camt_parser.py
Devstellar: Done, thanks. I added a comment explaining why the entry-level `AddtlNtryInf` fallback is only used when there is at most one `TxDtls`.

I also updated the PR title to include `[18.0]`.

account_statement_import_camt/tests/test_import_bank_statement.py
Devstellar: Done, thanks. I renamed the parser tests and the corresponding sample/golden files to make the 0, 1 and 2 `TxDtls` cases explicit.

clementmbr (approved):
Perfect! Thanks @Devstellar

PR #1010, opened 2026-09-21
[15.0][FIX] account_statement_import_online_gocardless: fix token renewal and surface API errors

carlosdauden:
@Tecnativa TT64556
ping @sergio-teruel @juancarlosonate-tecnativa


juancarlosonate-tecnativa:
This is the same fix as #748 shouldn't it be a backport, cherry-picking and keeping authorship, instead of a rewrite from scratch?

carlosdauden:
This PR isn’t based on that other one; as well as fixing that issue, it resolves another one and adds tests.
Incidentally, it would have been nice if the author had backported this version...

